PR #4, opened 2017-03-09
Fixed signed/unsigned comparison warning.

cesfahani:
Compiling with gcc -Wall caused a warning indicating a comparison between signed and unsigned integer expressions. This is because size_t is unsigned on most platforms, and the socket_write function's size argument was defaulted to -1 (and was then being compared to -1 to check for a defaulted value).
Instead, now we simply overload the socket_write method to have a version with and without the size argument.


yhirose:
@cesfahani, thanks for the fix!

PR #13, opened 2017-08-29
Allow user to set socket flags for server creation

curtpm:
This will allow someone to specify NULL/nullptr for the host field and set AI_PASSIVE to allow listening to incoming requests on any connection (ie. localhost / ip4 / ip6)

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -135,7 +135,7 @@ public:
     void set_error_handler(Handler handler);
     void set_logger(Logger logger);
 
-    bool listen(const char* host, int port);
+    bool listen(const char* host, int port, int socket_flags = 0);
     void stop();
 
 protected:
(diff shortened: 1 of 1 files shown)


yhirose:
@curtpm, thanks for the fine contribution. I added 2 test cases for the change. 
In the 2nd test, I created the client socket with 'locashost' for the server made with AI_PASSIVE, and the test passed successfully. I think it's correct, right?

curtpm:
looks good to me! 

FYI most server API's don't even support setting a host name on the "listen" function because 99% of the time you wan to listen to all connections on the system for the specified port. 
ex: golang http *ListenAndServe* https://golang.org/pkg/net/http/#pkg-examples

Also thanks for accepting the request so fast!

yhirose:
Thanks for the helpful information! 👍

PR #20, opened 2017-11-22
Example fix - should be no const qualifier

vdudouyt:


underscorediscovery:
it should be on both, numbers and hi, btw :)

PR #22, opened 2017-11-22
shield windows defines, in case they are already defined

underscorediscovery:
If these are already defined (like via project settings or command line builds), trying to redefine these is an error. 

This prevents that error and allows the code to be used there as well.


yhirose:
@underscorediscovery, thanks for the fix!

PR #23, opened 2017-11-22
add progress callback for http clients

underscorediscovery:
This is useful for displaying progress bars :) Here is how I use it.

progress

I tried to make the code idiomatic and avoided breaking any existing call sites.

```cpp
httplib::Client client(url, port);

// prints: 0 / 000 bytes => 50% complete
std::shared_ptr<httplib::Response> res = 
    cli.get("/", [](int64_t len, int64_t total) {
        printf("%lld / %lld bytes => %d%% complete\n", 
            len, total,
            (int)((len/total)*100));
    }
);
```

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -66,6 +66,7 @@ namespace httplib
 typedef std::map<std::string, std::string>      Map;
 typedef std::multimap<std::string, std::string> MultiMap;
 typedef std::smatch                             Match;
+typedef std::function<void (int64_t current, int64_t total)> Progress;
 
 struct Request {
     std::string method;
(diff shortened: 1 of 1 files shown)


underscorediscovery:
here is how I use it
progress

yhirose:
@underscorediscovery, thanks for the contribution. This is truly a fantastic feature!

underscorediscovery:
no problem @yhirose , I have some more contributions but I am still working on them. 
For the server, I have made a `listen` with non-blocking flag, and a `poll` function.  It is also minimal changes but makes it easy to use without a thread. I am still testing on more platforms before PR, and I have some other things (like websocket) I can contribute also. Thanks for the lib btw, it's a nice simple basis to use :)

hasanenalbana:
Hi,

What is the best way to get result body from within this lambda function?

PR #25, opened 2017-11-26
Implemented multipart-formdata support + added GCC version check

vdudouyt:


yhirose:
@vdudouyt, thanks for the PR, and I learned a lot from your code.

Here is my actual commit for the issue: https://github.com/yhirose/cpp-httplib/commit/bb8a1df7a383121338c62fd9ef8a31e4bbc287ca

This implementation gives us the filename. It also gives text offset and length in the body content, so that we could reduce memory usage. We can get the file content with the offset, length and body content. Please see the [document]( https://github.com/yhirose/cpp-httplib#multipartform-data-post-data).

Hope you can benefit from it.

Thanks again for your contribution!

PR #41, opened 2018-03-06
Added capture and setting of REMOTE_ADDR in request.

kevinbcarpenter:
Not sure if its the best way to fix, but was the one that worked for me.  Had considered capturing the IP at the accept in Server::listen, but figured I would have to add more scaffolding to pass the value to the request.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -157,6 +157,7 @@ public:
     virtual int read(char* ptr, size_t size) = 0;
     virtual int write(const char* ptr, size_t size1) = 0;
     virtual int write(const char* ptr) = 0;
+    virtual std::string get_remote_addr() = 0;
 
     template <typename ...Args>
     void write_format(const char* fmt, const Args& ...args);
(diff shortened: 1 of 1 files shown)


kevinbcarpenter:
So from my web experience, I know that in PHP you can reference the server variable, such as, `$_SERVER["REMOTE_ADDR"]` to get the remote IP address.  This is similar to `X-Forwarded-For`, however as I understand, that particular header is forwarded in the request itself (such as tracking through a proxy) and can be spoofed.  The `REMOTE_ADDR` in this implementation gives us the direct IP from the network, which helps us to eliminate spoofing at least from the request.

Looking at how Apache and NGINX implemented setting this value helped me understand why it's passed up from the network layer. That said, you are correct that it isn't a _standard_ header and so perhaps it would be better implemented as a function?  I think I biased this implementation due to my PHP experience.

yhirose:
@kevinbcarpenter, your PR looks very good!

I made some formatting changes, fixed a build error with `OPENSSL_SUPPORT` and added a unit test. Thanks again for your fantastic contribution!

Issue #48, opened 2018-04-15
Race in startup/shutdown test code


sgraham:
I think this only really affects the test code, but I thought perhaps you would want to know.

In investigating a possible patch for #46, I split `listen()` into helper functions locally. After I did that I was experiencing an occasional crash in the server tests. I made a repro by adding a short sleep() in the body of listen() on master like this: https://github.com/yhirose/cpp-httplib/compare/master...sgraham:start-stop-crash?expand=1

When I run this (on a Mac), ASAN reports a crash in the thread: 
https://gist.github.com/sgraham/461f286144ef8c7e531dad75ec2960d1

I think this is because the `svr_.is_running()` in the test SetUp() passes while `listen()` is in the sleep, then the server is immediately stopped, *then* the thread continues on to `detail::select_read()`, but `svr_sock_` is now -1, so `select_read()` crashes in `FD_SET(-1, &fds)`.

(Sorry for the long explanation without a fix, I wasn't immediately sure where the correct place to synchronize would be.)

yhirose:
@sgraham, thanks for the bug report. I fixed it at b6df220 with your nice test sample. Please let me know if you still have the problem. Thanks for your help!

PR #49, opened 2018-04-16
Support system-assigned port via two part listen()

### LINKED ISSUE #46: Method to retrieve port after starting with port 0?
I was trying to use cpp-httplib for running tests, so I'd like to start it on a system-selected port, and then have clients connect to that port. I don't want to pre-select a port number to avoid collisions with other things on the system, or when tests are run in parallel.

When I do the common thing of passing 0 as the server port everything seems to work, and I was able to hunt down the actual port by running netstat, and when I then connect to that port things work just fine.

Would it be possible to have a method to retrieve port that the server's running on? (Or is that already possible and I missed how to do so?)

sgraham:
This fixes #46 by allowing the user to separate the port bind from the
blocking listen(). Two new API functions bind_to_any_port() (which
returns the system-assigned port) and listen_after_bind() are equivalent
to the existing listen().

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -196,6 +196,9 @@ public:
     void set_error_handler(Handler handler);
     void set_logger(Logger logger);
 
+    int bind_to_any_port(const char* host, int socket_flags = 0);
+    bool listen_after_bind();
+
     bool listen(const char* host, int port, int socket_flags = 0);
 
     bool is_running() const;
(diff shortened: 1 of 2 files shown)


yhirose:
@sgraham, thanks for the PR! When I have time, I'll looking to #48.

sgraham:
Thank you so much for the quick review and merge!

PR #50, opened 2018-04-16
Support Content-Encoding: gzip on server side

sgraham:
If the client specifies Content-Encoding: gzip for POST requests,
decompress the body before attempting to parse it.


yhirose:
@sgraham, looks very good!

I just wonder if we should respond with '406 Not Acceptable' when the server isn't built with `CPPHTTPLIB_ZLIB_SUPPORT` and a client POST request contains `Content-Encoding: gzip`. What do you think?

sgraham:
Some error seems like a good idea.

From what I can see looking at https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/406 and https://httpd.apache.org/docs/current/content-negotiation.html I'm not sure if it's appropriate for an unexpected `Content-Encoding: gzip` or not. But I guess as long as it's a 4XX it's fine and making it more specific might be helpful to track down what's wrong.

Done!

yhirose:
It seems like the response should be '415 Unsupported Media Type'.
https://tools.ietf.org/html/rfc7231#page-13
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/415

Thanks again for the great contribution!

sgraham:
Aha! Great, thanks.

PR #51, opened 2018-04-16
Read \r\n terminator after chunked encoding

sgraham:
I believe according to https://en.wikipedia.org/wiki/Chunked_transfer_encoding, after the terminating zero-sized chunk of "0\r\n", there's an additional "\r\n" that should be read.

(I was seeing 400 errors due to this following a chunked upload because the server would try to dispatch an empty method+resource due to the additional blank line.)


yhirose:
@sgraham, good catch! I'll add a test case for it later.

PR #54, opened 2018-04-17
Make 'chunked' in Transfer-Encoding case-insensitive

sgraham:
NSURLConnection spells chunked Transfer-Encoding as "Chunked".


yhirose:
@sgraham, thanks for the fix!

PR #55, opened 2018-04-17
Don't fail chunked read if buffer not yet filled

sgraham:
I found on large chunked uploads on Windows 7, the buffer might not yet have received the entire chunk's data. Instead of bailing when all the bytes aren't available, I believe this needs to loop until the full specified size has been read, or the read returns -1.

(Perhaps a looping read should be in a lower level helper function? I'm not certain if there's other places this might realistically be needed.)

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -817,10 +817,8 @@ bool read_content_chunked(Stream& strm, std::string& out)
     auto chunk_len = std::stoi(reader.ptr(), 0, 16);
 
     while (chunk_len > 0){
-        std::string chunk(chunk_len, 0);
-
-        auto n = strm.read(&chunk[0], chunk_len);
-        if (n <= 0) {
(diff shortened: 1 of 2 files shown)


yhirose:
@sgraham, another good catch!

Regarding your comment about 'a lower level helper function', there is already such a function. (Please see the latest [read_content_with_length()](https://github.com/yhirose/cpp-httplib/blob/master/httplib.h#L770-L788) function.) Do you think the helper function can work for you?

sgraham:
Aha, yes! Especially now that it's std::string& instead of T&. :) Updated to use that and added a test. Unfortunately, I couldn't get a test to fail on other OSs but at least this tries to exercise that case.

yhirose:
Thanks a lot for your contribution!

PR #59, opened 2018-04-26
add keyword inline to some function definitions to avoid linking error

adikabintang:


yhirose:
@adikabintang, Great! Thanks for your contribution.

PR #67, opened 2018-05-09
Define and use kInvalidSocket

jperaza:
SOCKET on Windows is an unsigned type, which generates warnings when compared with -1. This change makes the comparisons against INVALID_SOCKET instead.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -39,6 +39,7 @@
 #endif
 
 typedef SOCKET socket_t;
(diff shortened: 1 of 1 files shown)


yhirose:
@jperaza, thanks for the PR! I changed the code slightly to be able to compile on the VS2013 old compiler which I still have to support. (VS2013 doesn't properly handle `constexpr`... I changed to define INVALID_SOCKET for POSIX based operating systems.)

PR #70, opened 2018-05-29
Handle port==0 when socket is bound on ipv6

sgraham:
I discovered
https://github.com/yhirose/cpp-httplib/commit/0515c6aad6834a2a329db226b9312294d0002a13
doesn't work when the server is bound on an AF_INET6 address on Windows
due to the getsockname() call failing.


yhirose:
@sgraham, thanks for another contribution! I have tested on Mac, and on Windows with VS2015 and VS2017, and all worked fine.

PR #75, opened 2018-06-01
More status messages for http codes

crtxcr:
I don't know whether it would be more practical to more or less add all http status codes. For now, I have  added some which are commonly used by applications (e. g. redirects).


yhirose:
@crtxcr, thanks for the congribution!

yhirose:
@crtxcr, thanks for the contribution!

PR #76, opened 2018-06-02
Fixed bound checks for #74

### LINKED ISSUE #74: Out of bounds array access in decode_url and from_hex_to_i
decode_url is used in Server::parse_request_line 
`        req.path = detail::decode_url(m[3]);`

First out of bounds access happens if the path ends with %:
```

 if (s[i] == '%') {
            if (s[i + 1] && s[i + 1] == 'u') {

```

Furthermore, **from_hex_to_i** gets the index, but trusts it, so here some checks are also necessary.

crtxcr:
Thank you for your quick reactions. However, I'm afraid 2bb27aa25d51020ad8b3570c880530ba0bff833b to fix #74 is a little bit incomplete. Checking `s[i]` in the for loop does not appear to suffice, because for example the line ` i += 5; // 'u0000'`  will cause an out-of-bounds access on the next iteration.

 Please consider the following patch.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -1001,9 +1001,9 @@ inline std::string decode_url(const std::string& s)
 {
     std::string result;
 
-    for (size_t i = 0; s[i]; i++) {
+    for (size_t i = 0; i < s.size(); i++) {
         if (s[i] == '%' && i + 1 < s.size()) {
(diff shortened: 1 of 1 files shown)


yhirose:
@crtxcr, thanks for the follow up! You are absolutely correct. I'll merge it right away.

PR #82, opened 2018-07-27
Fix Keep-Alive behavior and http version parsing

mkolinichenko:
First of all, thank you for your work. 
I was testing a simple app based on your server and found two issues. First one, I beleive, is a typo: after regex matching you're looking for a protocol version in m[4] while it resides in m[5]. Second one addresses Keep-Alive behavior. I was sending requests with ApacheBenchmark (ab) utility. AB is able to utilize Keep-Alive connections, but being HTTP/1.0 client it requires "Connection: Keep-Alive" explicitly specified in response header. Neither it received "Connection: close", so ab just waited for whole 5 seconds on each request. Also I believe it's not correct to set "Connection: close" in response to each HTTP/1.0 request as it is now in [write_response](https://github.com/yhirose/cpp-httplib/blob/222f49a1252b0dc289818dfb9b4c67bb5dbd6715/httplib.h#L1563). Thankfully it wasn't working because of the first issue :)
So my proposal is to set close, if close was requested or if we're about to close according to last_connection. And to set Keep-Alive if it was requested and it's not the last connection.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -1528,7 +1528,7 @@ inline bool Server::parse_request_line(const char* s, Request& req)
 
     std::cmatch m;
     if (std::regex_match(s, m, re)) {
-        req.version = std::string(m[4]);
+        req.version = std::string(m[5]);
         req.method = std::string(m[1]);
         req.target = std::string(m[2]);
         req.path = detail::decode_url(m[3]);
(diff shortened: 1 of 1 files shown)


yhirose:
@mkolinichenko, thanks for the PR. I checked those problem and your fixes, and all look good. Thanks for your contribution!

PR #84, opened 2018-07-31
Fix small issues in tests and added some extra checks.

davidgfnet:
It did not work out of the box using "make" in test/, therefore added dependency on cert in Makefile and automated cert generation, since we do not care about the cert really.
Minor other stuff too.

### DIFF
--- a/test/Makefile
+++ b/test/Makefile
@@ -2,19 +2,19 @@
 CC = clang++
 #CC = g++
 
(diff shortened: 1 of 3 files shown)


yhirose:
@davidgfnet, thanks for the pull request. It looks very good. I discovered a long-standing bug in `listen_internal` method when reviewing your pull request. Thanks for your contribution!

PR #85, opened 2018-07-31
Implementing streaming Responses

davidgfnet:


yhirose:
@davidgfnet, thanks for another contribution. ca343ae seems to be for #84, right?

davidgfnet:
Yeah! I'm not sure about what's the best interface for this but I went for simple-is-better

yhirose:
This interface for the streaming response looks good to me. I just merged it.
Also is it possible for you to add simple usage after the section ''multipart/form-data' POST data' in README? Thanks for the great contribution!

PR #89, opened 2018-08-06
Request cancelation feature

dolphineye:
This commit modifies the signature of the `Progress` callback so that its return value will indicate whether the request shall continue to be processed by returning `true`, or if it shall be aborted by returning `false`. Such modification will allow one to cancel an ongoing request before it has completed.

When migrating, developers should modify there `Progress` callbacks to always return `true` by default in case there do not want to benefit from the cancelation feature.

A few unit tests use cases were provided, but anyone should feel free to provide additional uses cases that they find relevant.


yhirose:
@dolphineye, thanks for the implementation and unit tests. All looks good. I'll adjust README soon. Thanks again for your contribution!

PR #90, opened 2018-08-17
Refactor the examples to compile with a C++11 compiler

mrexodia:
The current examples don't compile with a C++11 compiler. Since this is a C++11 library I refactored them to do.


yhirose:
@mrexodia, thanks for the contribution!

PR #92, opened 2018-08-17
Change size_t to time_t where applicable

mrexodia:


yhirose:
@mrexodia, I tested on MacOS and works fine. Thanks for the contribution!

PR #98, opened 2018-10-15
Fix undefined behavior + make some ifdefs more readable

mrexodia:
Nothing big here, just some minor changes I had in my local repo

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -5,38 +5,39 @@
 //  MIT License
 //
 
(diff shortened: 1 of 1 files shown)


yhirose:
@mrexodia, thanks for the feedback. Which part of your change is related to 'undefined behavior'?

yhirose:
@mrexodia, thanks for the information!

PR #99, opened 2018-10-15
travis and appveyor configuration

mrexodia:
This will need some extra work, you have to register the project on https://travis.ci.org and https://ci.appveyor.com for this to trigger. Currently it will build+test on Windows+macOS (because std::regex doesn't appear to be working on ubuntu 14).

Once the projects are created badges can be added to the README

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ Release
 ipch
 *.dSYM
 .*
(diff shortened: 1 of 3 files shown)


yhirose:
@mrexodia, sorry for the late response. Thanks for your contribution!

PR #109, opened 2018-11-22
Set response content-length of zero to prevent client hangs

zpalmtree:
If we have an empty body, the content length doesn't get set. This can cause hangs in some clients, for example, curl. A simple program to illustrate:

```cpp
#include "httplib.h"

using namespace httplib;

int main()
{
    Server svr;

    svr.Get("/", [](const Request &req, Response &res)
    {
        res.status = 200;
    });

    svr.listen("127.0.0.1", 1234);
}
```

If we then perform `curl http://127.0.0.1:1234 -v` - We can see:

```
< HTTP/1.1 200 OK
* no chunk, no close, no size. Assume close to signal end
< 
* Closing connection 0
```

On my machine, this hangs for approximately 3 or so seconds.
If you then apply the patch in my PR, and try again, it should complete instantly:

```
< HTTP/1.1 200 OK
< Content-Length: 0
< 
* Connection #0 to host 127.0.0.1 left intact
```

Environment: Arch Linux 4.19.1, Curl 7.62.0

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -1643,6 +1643,11 @@ inline void Server::write_response(Stream& strm, bool last_connection, const Req
         res.set_header("Connection", "Keep-Alive");
     }
 
+    /* Set length even if it's empty - connections sometimes hang if no length
+       given with an empty response */
+    auto length = std::to_string(res.body.size());
+    res.set_header("Content-Length", length.c_str());
+
     if (!res.body.empty()) {
 #ifdef CPPHTTPLIB_ZLIB_SUPPORT
         // TODO: 'Accpet-Encoding' has gzip, not gzip;q=0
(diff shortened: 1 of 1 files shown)


yhirose:
@zpalmtree, thanks for your contribution! But it breaks the following unit test...

```
[ RUN      ] ServerTest.GetStreamedChunked
test.cc:836: Failure
Value of: res->body == "aaabbb"
  Actual: false
Expected: true
```

Could you try this new version? https://github.com/yhirose/cpp-httplib/commit/5ad4311fb0257c1ced9d2f8e4fe32b0c478120da
It should fix your problem and pass all the unit test cases.

Thank you!

zpalmtree:
My bad, I should have run the tests. Your fix works perfectly, thank you!

PR #129, opened 2019-02-24
Encoding LF and CR

geraldolsribeiro:


yhirose:
@geraldolsribeiro, thanks for your fine contribution!

PR #130, opened 2019-02-25
Fix build warning with Boost::asio

Whitetigerswt:
This commit fixes a minor build warning if including boost::asio before cpp-httplib:

1>c:\\...\src\httplib.h(32): warning C4005: 'NOMINMAX': macro redefinition (compiling source file src\APIManager.cpp)
1>C:\\...\boost_1_67_0\boost/asio/detail/config.hpp(954): note: see previous definition of 'NOMINMAX'


yhirose:
@Whitetigerswt, good catch! Thanks for your contribution.

PR #132, opened 2019-03-07
small fix to allow fronting

archivaldo:
Hi!,

this is an small fix to be able to set the Host header manually.
In this way you can use this library to do requests using domain fronting.

thanks for your hard work!

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -2066,19 +2066,22 @@ inline void Client::write_request(Stream& strm, Request& req)
         path.c_str());
 
     // Headers
-    if (is_ssl()) {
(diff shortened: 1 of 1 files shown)


yhirose:
@archivaldo, thanks for the contribution! It looks good. Only think is that the coding style doesn't match the surroundings. Could you make an adjustment to use '4 spaces indent' and the same curly brace style in the code? Then, I'll merge it right away. Thanks!

archivaldo:
done!, 
thanks for you!

yhirose:
@archivaldo, thanks for your contribution!

PR #135, opened 2019-03-27
Add WinSock2 dependency

batist73:
The library uses WinSock2 on Windows. So, you need to link ws2_32.lib manually to projects, that use the library. I suggest you use pragma comment directive to tell linker to add WinSock2 as dependency. So, the library can be used without additional actions on Windows with VS compiler.


yhirose:
@batist73, thanks for the contribution!

PR #148, opened 2019-05-14
Update httplib.h

marknelson:
CPPHTTPLIB_PAYLOAD_MAX_LENGTH is only used in one place, that is to initialize payload_max_length_. Because payload_max_length_ is a size_t, attempts to intialize it with a 64 bit integer max will generate warnings, notably on 32 bit compilers. Changing the max to the same size as the target resolves that, and this should be no risk.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -88,7 +88,7 @@ typedef int socket_t;
 #define CPPHTTPLIB_KEEPALIVE_TIMEOUT_USECOND 0
 #define CPPHTTPLIB_KEEPALIVE_MAX_COUNT 5
 #define CPPHTTPLIB_REQUEST_URI_MAX_LENGTH 8192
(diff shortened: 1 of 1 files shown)


yhirose:
@marknelson, thanks for the pull request! Your name somehow sounded very familiar to me. Then I realized that I have your book, the Japanese translation of the first edition of "The Data Compression Book" that I got 25 years ago. I got rid of lots of my computer books, but I still keep this one in my bookshelf because it's one of my favorite books! :)

marknelson:
>. Then I realized that I have your book, the Japanese translation of the
first edition of "The Data Compression Book" that I got 25 years ago

Oh that's so cool. You can't imagine how excited I was when the publisher
arranged for translations of that book to German and Japanese! I still have
my copies too, of course :-)


------------------------------------------------------------------------------

Mark Nelson – mark@nelson.us - http://marknelson.us

Issue #151, opened 2019-05-19
Client POST parameter values are not encoded properly


hans-erickson:
When a POST request is received by the server, the function decode_url() is called on the parameter values to properly decode any encoded characters. This behavior seems to be correct.

However, on the client side, a matching encode_url() is not called on the parameter values. This can result in the key/value pairs being sent by the client in a way that can be parsed incorrectly by the server.

A suggested patch is attached.

[encode_patch.txt](https://github.com/yhirose/cpp-httplib/files/3195593/encode_patch.txt)

yhirose:
@hans-erickson, thanks for the report. Could you send it as a pull request?

yhirose:
Closed by #151.

PR #152, opened 2019-05-20
Add encode_url() call to Client class when adding parameter values. (issue #151)

hans-erickson:
As requested, here is a pull request for issue #151.


yhirose:
@hans-erickson, thanks for your contribution!

Issue #154, opened 2019-05-25
Support SSL client certificates


PixlRainbow:
Does cpp-httplib currently support?

1. (server side) requesting the client to submit a "client certificate" or
2. (client side) submitting a "client certificate" if requested to do so by a server
3. (server side) if (1.) is true, rejecting client if the client did not submit a "client certificate"

yhirose:
@PixlRainbow, you are asking for the feature like `curl -E CERT_FILE`, right?
I haven't implemented the feature, yet... I put the 'enhancement' label for my reminder.
Thanks for your feedback!

PixlRainbow:
I've implemented this feature in #162 but through testing I've noticed that Firefox does not appear to be able to handle intermediary-signed client certificates.

## Tested Conditions
| Certificate  | Client Type                                               | Result                            |
|--------------|-----------------------------------------------------------|-----------------------------------|
| Signed by CA | Firefox                                                   | SSL_ERROR_HANDSHAKE_FAILURE_ALERT |
| Self Signed  | Firefox                                                   | OK                                |
| Signed by CA | cURL,<br>OpenSSL s_client,<br>Google Chrome,<br>cpp-httplib client | OK                                |
| Self Signed  | cURL,<br>OpenSSL s_client,<br>Google Chrome,<br>cpp-httplib client | OK                                | 

Searching online for related issues brings up results that mention Firefox has a strange policy where the client certificate's issuer must match the host's SSL certificate issuer, so it may just be me not creating the certificates correctly.

yhirose:
@PixlRainbow, thanks for the PR. I'll put comments on #162 instead of here.

yhirose:
Fixed by #162

PR #162, opened 2019-06-04
client certificates support

PixlRainbow:
#154

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -391,7 +391,8 @@ private:
 
 class SSLClient : public Client {
 public:
-  SSLClient(const char *host, int port = 443, time_t timeout_sec = 300);
+  SSLClient(const char *host, int port = 443, time_t timeout_sec = 300,
+            const char *client_cert_path = nullptr, const char *client_key_path = nullptr);
 
   virtual ~SSLClient();
 
@@ -2390,7 +2391,8 @@ inline bool SSLServer::read_and_close_socket(socket_t sock) {
 }
 
 // SSL HTTP client implementation
-inline SSLClient::SSLClient(const char *host, int port, time_t timeout_sec)
+inline SSLClient::SSLClient(const char *host, int port, time_t timeout_sec,
+                            const char *client_cert_path, const char *client_key_path)
     : Client(host, port, timeout_sec) {
   ctx_ = SSL_CTX_new(SSLv23_client_method());
 
@@ -2398,6 +2400,13 @@ inline SSLClient::SSLClient(const char *host, int port, time_t timeout_sec)
                 [&](const char *b, const char *e) {
                   host_components_.emplace_back(std::string(b, e));
                 });
+  if(client_cert_path && client_key_path) {
+    if (SSL_CTX_use_certificate_file(ctx_, client_cert_path, SSL_FILETYPE_PEM) != 1
+      ||SSL_CTX_use_PrivateKey_file(ctx_, client_key_path, SSL_FILETYPE_PEM) != 1) {
+      SSL_CTX_free(ctx_);
+      ctx_ = nullptr;
+    }
+  }
 }
 
 inline SSLClient::~SSLClient() {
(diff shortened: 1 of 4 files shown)


yhirose:
@PixlRainbow, thank you so much for the fine contribution! I'll review the code when I am not busy.

Meanwhile, could you add at least two unit test cases ('success' and 'failure' cases) in text.cc, so that we can make sure that other contributors won't break the code accidentally? There are already test cases for server certificate check at the bottom of test.cc. So you can add them after that.

Only think so far in the code is about the usage of `SSL_CTX_load_verify_locations`. According to [the OpenSSL API reference](https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_load_verify_locations.html), both `CAfile` and `CApath` are not required when calling the API. It's ok if either one is specified. So I think the sentences like `if(trusted_cert_path && client_cert_path)` should be `if(trusted_cert_path || client_cert_path)`. I found the similar code pattern in curl at [here]( https://github.com/curl/curl/blob/8986d3c6f4f72cb4734f24ef297d31865e3018cc/lib/vtls/openssl.c#L2605-L2608)

Thanks a lot!

PixlRainbow:
@yhirose although `CApath` is optional (for self-signed client certificates), `CAfile` is actually needed in this case because although `SSL_CTX_load_verify_locations` can accept having either one, `SSL_load_client_CA_file` absolutely requires a file and not a directory.
The former loads certificates and builds the chain of trust locally while the latter is the list of acceptable CAs to be requested from the client.

PixlRainbow:
I have found the reason why my tests were failing with desktop Chrome and Firefox with CA-signed certificates. I was testing with the wrong certificate. I took a closer look at the [OpenSSL API reference](https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_client_CA_list.html#NOTES) and realized I should be using the _CA's_ certificate as `client_cert_path` and not the client's public key.
The reason why my wrong test configuration worked with self-signed client certificates is because such certificates serve as root CAs and client public keys in one.

PixlRainbow:
it appears the build failed on Travis CI because c_rehash was removed from Mac OS X starting with High Sierra

PixlRainbow:
@yhirose you mind initiating a rebuild on AppVeyor? I did not change anything between b2f4120 and 2d532a6 that would have caused _just_ the Windows build to fail on that particular test.

yhirose:
@PixlRainbow, thanks for adding the test cases. I kicked the PR build again. Interestingly, my first and second attempts failed, but the third attempt succeeded.

yhirose:
@PixlRainbow, thanks for adding the test cases. I kicked the PR build again. The first and second attempts failed like you experienced, but the third attempt finally succeeded. It's really interesting, but good news anyway.

PixlRainbow:
might have just been that particular cloud instance being too busy to serve packets or dropping packets

PixlRainbow:
@yhirose I am considering adding the X509's name field to `Request` so that request handlers of `SSLServer` are able to identify multiple specific users. The name is obtained by getting X509 pointer with `SSL_get_peer_certificate(const SSL *s)` and then accessing the `char* name` member of the X509. However, `Request req` is constructed within the scope of `Server::process_request` so `SSLServer` cannot access `req` by calling the inherited function.
**I can think of two ways**
1. Modify `Server::process_request` to take `Request &req` as a parameter like `Client::process_request` so that `SSLServer` can attach a cert name to the `req` before passing to inherited function.
2. Copy and paste the entire contents of `Server::process_request` into `SSLServer::process_request`, then adding one line before the `routing` call to add certificate name to `req`.

I am asking for your suggestion or opinion because I am afraid of making changes that you might consider too drastic and I am uncertain if you have intentionally sandboxed the Request within the base `Server` class for security reasons.

yhirose:
@PixlRainbow, thanks for the great contribution!

PR #165, opened 2019-06-12
access client certificate name in request handler

PixlRainbow:
update for #154 
allows request handlers to identify specific clients

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -129,6 +129,7 @@ struct Request {
   std::string method;
   std::string target;
   std::string path;
+  std::string X509_name;
   Headers headers;
   std::string body;
   Params params;
@@ -249,6 +250,9 @@ public:
 protected:
   bool process_request(Stream &strm, bool last_connection,
                        bool &connection_close);
+  bool process_request(Stream &strm, bool last_connection,
+                       bool &connection_close,
+                       Request &req, Response &res);
 
   size_t keep_alive_max_count_;
   size_t payload_max_length_;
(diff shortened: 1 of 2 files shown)


PixlRainbow:
about e2ecea3: I've noticed that if no client ca file is provided, the server will still start, but it will reject all clients.

PixlRainbow:
Noticed that my PR will not compile on OpenSSL 1.1.0 as they have made some structs opaque. Will push some commits to use accessors instead soon.

yhirose:
@PixlRainbow, thanks for the PR. I implemented what you are trying to do in a different way at https://github.com/yhirose/cpp-httplib/commit/1981e0ccad8a70524e2cbaa9ed7c2098c0207c27. In this implementation, `Request` has `const SSL* ssl` instead of `std::string X509_name`, so that users could get more information out of `SSL *`.

Here is an example to get the common name.

```cpp
    auto peer_cert = SSL_get_peer_certificate(req.ssl);
    ASSERT_TRUE(peer_cert != nullptr);

    auto subject_name = X509_get_subject_name(peer_cert);
    ASSERT_TRUE(subject_name != nullptr);

    std::string common_name;
    {
      char name[BUFSIZ];
      auto name_len = X509_NAME_get_text_by_NID(subject_name, NID_commonName,
                                                name, sizeof(name));
      common_name.assign(name, name_len);
    }

    EXPECT_EQ("Common Name", common_name);

    X509_free(peer_cert);
```

Is this approach is enough for you?

PixlRainbow:
Yeah that works! Thanks!

yhirose:
@PixlRainbow, thanks for your contribution!

PR #169, opened 2019-06-17
CMake support

omaralvarez:
Added modern CMake support to have a standard way of using the library.


yhirose:
@omaralvarez, thanks for the PR. But it has conflicts and `httplib.h` has been modified...

omaralvarez:
Yes. I will also try to keep the original dir structure, so that tests don't fail. I didn't have more time yesterday.

PR #172, opened 2019-06-18
CMake support

omaralvarez:
Added modern CMake support to have a standard way of using the library.


yhirose:
@omaralvarez, thank you for the quick response! I just noticed that `httplib.h` is now moved to `include` directory that breaks all the project files (Makefile and *.vcxproj files) in `example` directory. Is it requirement to have `include` directory for the Conan support? I would like to keep `httplib.h` right under the root directory, because some of my other git projects contains 'cpp-httplib' via git-submodule. So I have to fix up the include paths as well...

omaralvarez:
It has to do with CMake, I will try to find a solution. But the standard way of doing header only in CMake is that one. I can fix the examples, sorry for not noticing. I'll write here with whatever I find.

yhirose:
@omaralvarez, here is another example  (Chromium Crashpad) that depends on the structure in addition to my personal projects.

https://github.com/chromium/chromium/blob/2ca8c5037021c9d2ecc00b787d58a31ed8fc8bcb/third_party/crashpad/crashpad/util/net/http_transport_test_server.cc#L42

As you can see, once `httplib.h` moves into `include` directory, the code has to be fixed to point to the new location.

omaralvarez:
Everything is ready, I don't know why the appveyor test failed. Right now the file hierarchy is the same as before, so nothing should break.

yhirose:
@omaralvarez, thanks for the update. I kicked the AppVeyor test again, and it succeeded this time. Thanks for your great contribution!

omaralvarez:
@yhirose No problem, thanks to you for this awesome library!

PR #180, opened 2019-06-30
use 0.0.0.0 replace localhost to avoid some mistake

how0723:


PixlRainbow:
@how0723 `0.0.0.0` is *not a valid host address for clients*. It's only valid for servers as a way to say "bind/listen on all interfaces". And even then it is not recommended to use it for testing since you are opening yourself up to potential attack from outsiders.

yhirose:
@how0723, I agree with the above @PixlRainbow's comment. Also this pull request breaks the unit tests on CI servers.

Please see the following AppVeyer log: https://ci.appveyor.com/project/yhirose/cpp-httplib/builds/25639417

Issue #139, opened 2019-04-01
Can I specify an output stream for large downloads?


spyang:
So far, I really like your library!

My use case is with the SSLClient.  The callback for displaying download progress is working.  However, I only know how to wait for the entire result first before writing to disk (by getting the body attribute from the result).  In my case, the file I may be downloading could be 1GB in size, so I'd prefer to write the result to disk as it comes in.  Is there a way to do this?

yhirose:
@spyang, thank you for your feedback.

Sorry that there is no easy, built-in way... But if you check where `strm.read` is called, you may find a way to do it. Also you should probably remove `std::string body` from `Response` and add a callback function like `std::function<int (Stream& strm)> data_receiver` to take over the control of where the data should be stored.

Your request sounds a good idea anyway. Hope I can find a way to do it without changing the current httplib interface, so that existing users won't be affected and we can benefit from the new feature. Thanks!

ivanenko:
Im also voting for this feature!

yhirose:
The read handler should be`std::function<int (Stream& strm, size_t content_length)> data_receiver`.

We expect that `data_receiver` could be called in 3 different ways based on HTTP request type:
1. With content length (`content_length` should be greater than  -1)
2. Without content length ('content_length` should be -1)
3. Chunked transfer encoding (2. could be called more than once)

There is a big problem remaining still though, that is **'gzip' encoding support**. Right now, cpp-httlib inflates data after it receives entire data in `body` member. So if a user sets the `data_receiver`, it becomes now the users' responsibility to inflate data which is not an easy task...

PixlRainbow:
I've made an implementation similar to your third suggested way in #181, and I've set it up such that gzip encoded data will be inflated chunk-by-chunk before being given to a handler.
To create a handler, for example, you could do this
```cpp
size_t write_offset = 0;
FILE *testfile = fopen("download_file", "wb");
//the callback handler
bool recvData(std::string chunk){
    usleep(1000000U);
    std::cerr << "Chunk No. " << count++ << ": " << chunk << std::endl;
    fwrite(chunk.c_str(), chunk.size(), write_offset, testfile);
    write_offset += chunk.size();
    return true;
}
```
and in your `main` code
```cpp
httplib::Client cli("somehost.com");
//path, progress callback, chunk callback
cli.Get("/HTTP/ChunkedScript", nullptr, recvData);
```
[...]

PR #181, opened 2019-07-02
Chunked transfer handlers for client

PixlRainbow:


PixlRainbow:
I have just realized that process_request only checks the Request object for user-defined stream handlers very late in the processing, after it has already done reading from the incoming stream. This makes it difficult to implement chunked handlers for _incoming_ chunks without doing a major refactor

yhirose:
@PixlRainbow, thanks for the pull request! As you may have already noticed, I have been actually pondering over good user interface of the chunked transfer encoding support on client as well as other similar data transfer scenarios at #139. I'll try to understand your implementation and learn idea from it. Thanks for your efforts.

yhirose:
@PixlRainbow, I have just fixed and closed #139. I learned a lot from your pull request to know how to implement the chunked transfer handler effectively. Thanks for your contribution!

PixlRainbow:
No problem! Glad I could help. Just a pity that I have just found I cannot make use of this because neither Chrome not Firefox support streamed upload callbacks in JavaScript yet, even though it is in ES6 specifications.

Issue #160, opened 2019-06-04
Race condition between listen and stop, fortify failure


gemarcano:
This seems to be the related to problem #48, but I'll open this issue to bring attention to it.

The problem is brought about due to calling `stop()` immediately after calling `bind_to_any_port` and then launching `listen_after_bind()` on its own thread. I only get this when compiling the program in `Release` mode (CMake), since I'm assuming that's only when it's fast enough that the race manifests itself.

The problem is a race condition between two threads, let's call them T1 and T2. T1 is the "main" thread, and T2 is the one running `listen_after_bind()`:
1. T1 prepares the server (`bind_to_any_port`, etc.).
2. T1 then launches T2, with T2 running `listen_after_bind()`.
3. T1 waits (a simple loop with a sleep) for `is_ready()` to return true.
4. T2 sets `is_ready_` to true in `listen_internal()`, which causes `is_ready()` to return true for T1.
5. T1 calls `stop()` before T2 can continue, and shuts down `svr_sock_` and sets it to -1. At this point T1 calls `join()` on T2.
6. T2 now finally begins to proceed, and calls `detail::select_read(svr_sock_, 0, 100000);` in `listen_internal()`, which calls `FD_SET(-1, fds)` since `svr_sock_` is now -1. This causes the program, with fortify enabled, to blow up with an `*** buffer overflow detected ***:` error.

I noticed the code in `listen_internal` has an if statement meant to handled a case of when `svr_sock_` is set to -1, but if it's already -1 due to a race condition by the time `detail::select_read` is called, if the fortify compiler flags are used when compiling the program, that if statement will never be reached, the program will be aborted before then.

The [POSIX standard's page on select and pselect](https://pubs.opengroup.org/onlinepubs/9699919799/functions/pselect.html) has the following to say about FD_SET and negative fd arugments:

> The behavior of these macros [FD_SET] is undefined if the fd argument is less than 0 or greater than or equal to FD_SETSIZE, or if fd is not a valid file descriptor, or if any of the arguments are expressions with side-effects

I'm not 100% sure what's the right way to fix this. I think as it stands, `stop` is meant to be called from another thread from `listen_internal` (right?), so some kind of threading synchronization needs to happen between the two. I can't think of anything that can be done from outside httplib to fix this issue since `listen_internal` and any listen functions that call it effectively block, so there's no way to know when it's safe to call `stop()` from the outside (currently).

gemarcano:
[...]

PR #184, opened 2019-07-05
Made svr_sock_ and is_running_ variables atomic

jaspervandeven:
As discussed in our last email, here the last try, I will do to ensure that there is no data race condition. This is related to issue #160, which was closed, but the clang sanitizers still issued a warning for an existing data race. Hope, this fixes that. And thanks for the fantastic work on this library.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -70,6 +70,7 @@ typedef int socket_t;
 #include <string>
 #include <sys/stat.h>
 #include <thread>
+#include <atomic>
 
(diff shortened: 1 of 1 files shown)


yhirose:
@jaspervandeven, the code looks good to me. Thanks for your contribution!

brobusta:
This change didn't fix the data race, right?
you still access/close the underlying socket without any protections.

Issue #170, opened 2019-06-17
Versioning


omaralvarez:
Would it be possible for you to release a initial version (even if it is 0.1)? Using tags and the github release system. I would like to create a conan package that points to your repo for easy compilation using CMake.

I have made a pull request to add CMake support. Once you are satisfied with it, it would be nice if you made a release so I can properly add Conan support for master and the first version number.

omaralvarez:
@yhirose Now that CMake support is added is a perfect time for a release! :wink: I will have another pull request with Conan support as soon as it is created.

yhirose:
@omaralvarez, I just added 'v0.2.0' tag. Let me know what else should be done on my side.

omaralvarez:
@omaralvarez Perfect! Now the ball is in my court. 👍

yhirose:
@omaralvarez, any progress?

omaralvarez:
I actually was working on it today, but CMake gave me issues, I may have to open another pull request.

omaralvarez:
CMake error fixed. Would it be possible for you to accept it and recreate `v0.2.0`? The conan package is ready, and will just need these changes to work.

yhirose:
@omaralvarez, thanks for the progress. Only thing is the version number. Does 'recreate `v0.2.0`' mean remove the tag and add it to the latest commit, or something else? The reason why I am concerned with the version number is that someone kindly registered cpp-httplib to vcpkg library manager at https://github.com/Microsoft/vcpkg/pull/7037. So I just wonder if the `v0.2.0` should stay at the current commit (2f8479016fa9af9a63056571992b31c32741d89b).

omaralvarez:
Sounds good. Just release a new version then. e.g. v0.2.1

yhirose:
@omaralvarez, I added `v0.2.1` tag.

omaralvarez:
@yhirose Thanks a lot and sorry for my silly mistake that made us create another tag. Another pull request coming with information about using Conan with the library.

omaralvarez:
@yhirose Done! #195

yhirose:
@omaralvarez, @Pospelove, thanks for your contribution.

In fact, a serious bug has been reported after `v0.2.1` has been released, and `v0.2.2` may be coming soon. Also there will be even more coming one after another for bug fixes and new features. :)

[...]

PR #195, opened 2019-07-26
Add package manager info to README.md

omaralvarez:
I have just added instructions to use Conan to manage httplib as a dependency.


JoshuaBehrens:
Nice!

yhirose:
Close it for now. This issue could be reconsidered when #170 is reopened.

PR #198, opened 2019-07-31
add thread pool supporting.

xinyu391:
add thread pool supporting.
default no thread pool.
active thread pool by call this srv.set_thread_pool_size(n).

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,13 @@ svr.listen_after_bind();
 svr.set_base_dir("./www");
 ```
 
+
+### Using thread pool
+
+```cpp
+svr.set_thread_pool_size(4);
+```
+
 ### Logging
 
 ```cpp
--- a/httplib.h
+++ b/httplib.h
@@ -177,7 +177,7 @@ public:
             std::unique_lock<std::mutex> lck(mtx);
             quit = true;
         }
-        codv.notify_all(); // 唤醒线程池所有线程
+        codv.notify_all(); 
 
         for (auto& t : threads) {
             t.join();
@@ -187,8 +187,8 @@ public:
 public:
     std::mutex mtx;
     std::queue<std::function<void()>> queue;
-    std::condition_variable codv; // 唤醒线程池
-    std::condition_variable codv_que; // 唤醒add 等待队列
+    std::condition_variable codv;
+    std::condition_variable codv_que;
     bool quit;
 
 private:
--- a/httplib.h
+++ b/httplib.h
@@ -72,6 +72,11 @@ typedef int socket_t;
 #include <string>
 #include <sys/stat.h>
 #include <thread>
+#include <condition_variable>
+#include <future>
+#include <queue>
+#include <stdexcept>
+#include <vector>
 
 #ifdef CPPHTTPLIB_OPENSSL_SUPPORT
 #include <openssl/err.h>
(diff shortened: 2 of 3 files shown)


yhirose:
@xinyu391, thanks for the great pull request. In fact, I have been pondering over supporting the thread pool in cpp-httplib for longtime, but I haven't figure out the best way to implement it. One of the reason is that there are lots of thread pool implementations, and each one has its own unique characteristics and I don't want to insist my own implementation to others. (For instance, on Windows environment, people may want to use the [PPL](https://docs.microsoft.com/en-us/cpp/parallel/concrt/parallel-patterns-library-ppl?view=vs-2019) based implementation rather than C++11 thread library.

So I am now thinking to make the thread strategy pluggable in cpp-httplib and provide my implementation as a default. I have learned how to integrate a thread pool library a lot from your example. I'll get back to you when I come to have something to show you. Thanks!

yhirose:
@xinyu391, I have just implemented the pluggable thread pool interface (`class TaskQueue` and `new_task_queue` method on `Server`). I also provided my own implementation of thread pool as default which supports graceful shutdown feature.

If we want to override the default thread pool, we can do it as below:

```cpp
class OctopusTaskQueue : public TaskQueue {
public:
  OctopusTaskQueue() : pool_(new octopus::Pool(12)) {}

  virtual ~OctopusTaskQueue() {}

  virtual void enqueue(std::function<void(void)> fn) override {
    thread_pool->enqueue(fn);
  }

  virtual void shutdown() override { delete pool_; }

private:
  octopus::Pool *pool_;
};
```

```cpp
httplib::Server svr;
svr.new_task_queue = [] { return new OctopusTaskQueue(); };
```

In other words, we can use whatever thread pool library in cpp-httplib now.

xinyu391:
@yhirose , that's a great idea.
but   I don'nt understanding your default thread pool.
in your struct worker,  if you lock pool_.mutex_ in the for loop
mulity thread can not run `fn()` at same time.

``` C++
  struct worker {
    worker(ThreadPool &pool) : pool_(pool) {}

    void operator()() {
      for (;;) {
       //  ### lock the pool_.mutex
        std::unique_lock<std::mutex> lock(pool_.mutex_);

        pool_.cond_.wait(
            lock, [&] { return !pool_.jobs_.empty() || pool_.shutdown_; });

        if (pool_.shutdown_) { break; }

        auto fn = pool_.jobs_.front();
        pool_.jobs_.pop_front();

        assert(true == (bool)fn);
        fn(); // ### only one thread in the pool can run  fn() at same time.
      }

      std::unique_lock<std::mutex> lock(pool_.mutex_);
      pool_.remaining_--;
    }
```

second, in the `shutdown()`
  if there are still many request  in the jobs_ queue, the first for loop is not necessary, every thread in the pool will check the queue by itself.
you could just set `shutdown_ = true;` ,and check job_.empty() and shutdown in thread 。

```C++
 virtual void shutdown() override {
    // Handle all remaining jobs...
    {
      std::unique_lock<std::mutex> lock(mutex_);
      shutdown_ = true;
    }

    cond_.notify_all();

    // Join...
    for (auto t : threads_) {
      t->join();
    }
}
```


```C++
struct worker {
 ....
 void operator()() {
  for (;;) {
        std::unique_lock<std::mutex> lock(pool_.mutex_);

        pool_.cond_.wait(
            lock, [&] { return !pool_.jobs_.empty() || pool_.shutdown_; });

        if (pool_.shutdown_&&pool_.jobs_.empty() ) { break; }
...
```

yhirose:
@xinyu391, thanks for finding the bug in the worker class. When I ported the code form my thread pool code snippet, I removed some necessary code... I made it back. Here is the current code.

https://github.com/yhirose/cpp-httplib/blob/3629f8762784e5db24551d301f12ca4b668ab701/httplib.h#L336-L357

When it comes to the shutdown code, I'll look into it too see what you mentioned. Thanks for your valuable information!

yhirose:
@xinyu391, I checked your suggestion and looks very good. The latest cpp-httlib has the fix.
I made a separated repository called [cpp-threadpool](https://github.com/yhirose/cpp-threadpool), and I confirmed the thread pool does graceful shutdown successfully with [my test code](https://github.com/yhirose/cpp-threadpool/blob/master/test.cc). Thanks for your great contribution!

PR #214, opened 2019-09-14
New Client::Get variant combining a ContentReceiver with a ResponseHandler

gulrak:
While trying to implement streaming of internet radio, where a ContentReceiver is needed to handle the audio data, I had the problem, that important information about the stream data is part of the HTTP header (e.g. size of audio chunks between meta data), so I added a ResponseHandler and a new Get variant, to gain access to the header before handling the first chunk of data.

The ResponseHandler can abort the request by returning false, in the same way as the ContentReceiver.

A test case was also added.


yhirose:
@gulrak, thanks for the fine improvement! I added one line to complete this change and merged it into master.

PR #223, opened 2019-09-26
Fix several -Wold-style-cast warnings in Clang-9

Zefz:


yhirose:
@Zefz, thanks for your contribution!

PR #224, opened 2019-09-26
Allow configuration to be overriden without source editing

Zefz:
This allows users of `cpp-httplib` to override the default configuration (e.g number of threads) without needing to edit `httplib.h`.

e.g in my CMake project file:
```
target_compile_definitions(httplib INTERFACE 
    CPPHTTPLIB_THREAD_POOL_COUNT=0
   CPPHTTPLIB_READ_TIMEOUT_SECOND=3
)
```


yhirose:
@Zefz, thanks for the improvement. Only thing is the following:

```cpp
#ifndef CPPHTTPLIB_USE_POLL
#define CPPHTTPLIB_USE_POLL 1
#endif
```

Since `CPPHTTPLIB_USE_POLL` is boolean, I don't feel it's needed and I would like to keep `#ifdef CPPHTTPLIB_USE_POLL`. Does it cause any problem on your side?

Zefz:
We would need to change it to something like:

```

#ifndef CPPHTTPLIB_USE_POLL
#define CPPHTTPLIB_USE_POLL
#elif CPPHTTPLIB_USE_POLL == 0
#undef CPPHTTPLIB_USE_POLL
#endif
```

since we need to be able to tell configuration to explicitly not define the macro through the preprocessor.

Zefz:
And then you could use `#ifdef` instead of `#if` in the rest of the code.

yhirose:
@Zefz, sorry for the late reply. I pondered over it and decided to remove CPPHTTPLIB_USE_POLL and add CPPHTTPLIB_USE_SELECT instead. Since the `poll` implementation is the default behavior, `httplib.h` doesn't include the macro symbo. When a user explicitly wants to use the `select` behavior, he has to define the CPPHTTPLIB_USE_SELECT symbol in his processor setting. I think this approach is better than the previous.

So the code for CPPHTTPLIB_USE_POLL is not necessary any more. Could you make rebase the code in your pull request? Then, I'll merge it to the master right away. Thanks!

Zefz:
Makes sense! I've updated the PR.

yhirose:
Thanks for the adjustment!

PR #225, opened 2019-09-27
Fix compile error in android ndk

TangHuaiZhe:
Fix a compile error in android NDK.

httplib.h:2794:11: error: return type 'bool' must match previous return type 'int' when lambda expression has unspecified explicit return type
            return ret;
            ^

I'm not familiar with c++,try to declare the  lambda function returns a bool can fix the error.

Looks like a [C++11 restrictions on lambda return type:](https://stackoverflow.com/questions/26692637/c11-restrictions-on-lambda-return-type)

PS:My Android NDK Project using C++11 and clang tool chain.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -2773,7 +2773,7 @@ inline bool Client::send(const std::vector<Request> &requests,
 
     if (!process_and_close_socket(
             sock, requests.size() - i,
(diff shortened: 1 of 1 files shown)


TangHuaiZhe:
image

yhirose:
@TangHuaiZhe, thanks for your contribution!

PR #226, opened 2019-09-27
Fix compilation on Mingw-64

Zefz:
This minor change fixes compilation with the Mingw-64 compiler on Windows 7


yhirose:
@Zefz, thanks for your another contribution!

Issue #115, opened 2018-12-15
Crash on too much requests


Olegkravec:
When client sends about 15000 requests per ~10 sec the server is crashing. I think that there is end of memory allocation  for thread pool. When server crash, he doesnt respond any exception just return with code 11

Olegkravec:
How I know SIGSEGV signal is sending to application when he tries to appeal to unused page or when page access violation - SEGV_MAPERR and SEGV_ACCERR

yhirose:
@Olegkravec, thanks for the feed back. In order to reproduce it on my machine, I need more detailed information. Could you show me your server code? I'll try to reproduce it with the code by using Apache Bench. Thanks for your assistance!

xg0:
I have the same issue with more than 2000 requests:

*** buffer overflow detected ***: ./server terminated
======= Backtrace: =========
/lib/x86_64-linux-gnu/libc.so.6(+0x777e5)[0x7f6731c807e5]
======= Backtrace: =========
/lib/x86_64-linux-gnu/libc.so.6(__fortify_fail+0x5c)[0x7f6731d2215c]
/lib/x86_64-linux-gnu/libc.so.6(+0x117160)[0x7f6731d20160]
/lib/x86_64-linux-gnu/libc.so.6(+0x1190a7)[0x7f6731d220a7]
./server(_ZNSt6thread5_ImplISt12_Bind_simpleIFZN7httplib6Server15listen_internalEvEUlvE_vEEE6_M_runEv+0xda)[0x44b45a]
/usr/lib/x86_64-linux-gnu/libstdc++.so.6(+0xb8c80)[0x7f67325aac80]

yhirose:
I couldn't reproduce this problem on my machine with Apach Bench. I would really appreciated if someone finds an easy way to reproduce it.

PixlRainbow:
I wonder if it would fix the issue to add an atomic integer tracking number of active requests handlers/threads, then if some threshold is reached just start sending http 429 and connection close to any other requests.
Http 429 means "too many requests"

yhirose:
@Olegkravec and @xg0 , can I kindly ask you to give me more information please when you have time?

I tested with the sample server in the example folder and sent `ab -n 20000 -c 1000 https//localhost:8080/hi` command to the server. But it worked on Ubuntu 18.04 without any problem. Can you give your result by doing the same thing as I did? Thanks a lot!

PixlRainbow:
[...]

PR #227, opened 2019-09-30
Don't exit if accept fails due to no more file descriptors

ha11owed:
A burst of connections that causes the maximum number of file descriptors to be reached will cause the server to stop.
Instead of stopping, the server could retry after a short timeout (the pull request). More complex mechanisms could also be implemented.

Steps to simulate the problem:
1. Trigger more than 1024 parallel connections.
for example by calling "ab -r -n 90000 -c 1000 http://localhost:8888/index.html" in two separate processes (terminals).
2. Wait for the burst to be done (now no more connections to the server are active)
3. The server stopped! NOK

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -98,6 +98,7 @@ typedef int socket_t;
 #include <assert.h>
 #include <atomic>
 #include <condition_variable>
+#include <errno.h>
 #include <fcntl.h>
 #include <fstream>
 #include <functional>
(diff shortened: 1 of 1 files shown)


yhirose:
@ha11owed, thanks for your fine contribution!

PixlRainbow:
thanks! this may help reduce chance of #115 recurring

PR #228, opened 2019-09-30
Modernize code

Zefz:
* Replace C-style arrays with std::array since we are using C++11 anyways
* Fix some minor warnings reported by static code analysis
* Replace string literals with constants

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -230,7 +230,10 @@ enum class HttpMethod {
     METHOD_TRACE,
     METHOD_OPTIONS,
     METHOD_CONNECT,
-    METHOD_PATCH
+    METHOD_PATCH,
+
+    //HTTP/2
+    METHOD_PRI
 };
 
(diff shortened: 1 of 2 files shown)


yhirose:
@Zefz, sorry for the late reply, and thanks for your fine efforts to improve the code base. I checked the changes, I am interested in the first and second parts of your PR ('Replace C-style arrays with std::array since we are using C++11 anyways' and 'Fix some minor warnings reported by static code analysis'). Could you make respective pull requests for these, then I'll merge them.

As for the third ('Replace string literals with constants'), I understand this change is more type-safe code, but I just would like to keep the code simple. Also I don't want to break users' code using `Client::send` API with `std::string method` member under `struct Request`.

Thanks!

Zefz:
Ok, I'll split away the commit and update the PR when I have time.

yhirose:
Closed by #256.

PR #231, opened 2019-10-03
Fix a sample code of multipart/form-data POST data in the README.md.

sux2mfgj:
I tested by this code.
```
#include <iostream>
#include "httplib.h"

int main(int argc, char const* argv[])
{
    using namespace httplib;

    Server svr;

    svr.Post("/multipart", [&](const auto& req, auto& res) {
        auto size = req.files.size();
        auto ret = req.has_file("name1");
        const auto& file = req.get_file_value("name1");
        // file.filename;
        // file.content_type;
        auto body = req.body.substr(file.offset, file.length);
    });

    svr.listen("localhost", 1234);

    return 0;
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,12 +77,13 @@ svr.set_error_handler([](const auto& req, auto& res) {
 ```cpp
 svr.Post("/multipart", [&](const auto& req, auto& res) {
     auto size = req.files.size();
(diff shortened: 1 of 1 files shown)


yhirose:
@sux2mfgj, thanks for finding the mistakes!

PR #232, opened 2019-10-03
Fix a test, ConnectionErrorTest::InvalidPort. 

sux2mfgj:
currently, the `abcde.com` is valid, so I change it. The hyphen is not permitted at the first byte in hostname.


yhirose:
@sux2mfgj, thanks for the good catch and the fix!

PR #240, opened 2019-10-15
Add Out type

Cansisti:
Examples in README.md uses Out type, but it was not defined.


yhirose:
@Cansisti, thanks for the report. Actually README is wrong. I forgot updating the examples in README when I introduced `DataSink` type...

yhirose:
README has been updated. Thanks for your contribution!

PR #242, opened 2019-10-18
added server streamhandler to handle large post data. 

sophana:
It is now possible to register a stream handler which has access to the request stream.
The handler is able to call read_content on the stream with its own handler.
Updated simplesvr example

### DIFF
--- a/example/Makefile
+++ b/example/Makefile
@@ -1,7 +1,9 @@
 
 #CXX = clang++
 CXXFLAGS = -std=c++14 -I.. -Wall -Wextra -pthread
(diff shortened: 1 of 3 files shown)


PixlRainbow:
referenced PR: #242

PixlRainbow:
Could you not have simply re-used the existing `httplib::detail::ContentReceiver` instead of creating a new type?

yhirose:
@sophana, thanks for your contribution. I just took a look at your code, but I cannot accept this PR. It is because the fix exposes `Stream` and user code needs to call `detail::read_content`. Both are 'internal' items which are not supposed to be used by users.

But I understand what you are trying to do with the PR and it's useful for large data. I'll implement this feature when I have time. Thanks for bringing up this matter!

sophana:
You may want to hide read_content and stream by providing a binded version of read_content in the handlers arguments?

Issue #192, opened 2019-07-24
Deprecate OpenSSL versions below 1.1.1


yhirose:
https://www.openssl.org/policies/releasestrat.html

Version 1.1.0 will be supported until 2019-09-11.
Version 1.0.2 will be supported until 2019-12-31 (LTS).
Version 1.0.1 is no longer supported.
Version 1.0.0 is no longer supported.
Version 0.9.8 is no longer supported.

yamashi:
Does this mean currently the lib doesn't support the latest openssl versions ? I am running into build issues with openssl 1.1.x

yhirose:
@yamashi, I have just copied the chart as it is from the link. ('will be' looks a bit confusing...)
Anyway, it supports 1.1.1, and I confirmed again all the unit tests passed with 1.1.1. Thanks for asking.

yhirose:
Mac OS used by Github Action still uses an OpenSSL version less than 1.1.1...

PR #243, opened 2019-10-19
Allow use of OpenSSL 1.1.1, fix compile errors

Sil3ntStorm:


yhirose:
@Sil3ntStorm, thanks for the pull request. Does openssl 1.1.1 on your machine really cause the compile errors? I can compile httplib.h without any error even if I am using openssl 1.1.1d on my MacBook. I just would like to know why there is such a difference.

Also as you can see at #192, I am dropping the openssl support prior to 1.1.1 soon. (1.1.0 has already been deprecated, and 1.0.2 will be ending on 2019-12-31.). So I'll use `OPENSSL_init_ssl` for the initialization as your PR does and remove `SSL_library_init ` and `SSL_load_error_strings` calls completely from httplib.h.

Thanks!

Sil3ntStorm:
Those functions are not in OpenSSL any longer if you compile it without deprecated features. As such yes it does cause the error.
The static_casts are there to silence compiler warnings, as otherwise you assign int to char.

When you're going to remove < 1.1.1 support, you can just toss out the ifdefs at that time.

yhirose:
Thanks for the answer. I merged it.

PR #244, opened 2019-10-19
Added bind_to_port()

aaronalbers:
- This compliments the existing `bind_to_any_port()`
  where you can determine if the bind succeeded prior
  to calling `listen_after_bind()` but allows you to
  specify the port.


yhirose:
@aaronalbers, thanks for your contribution!

PR #246, opened 2019-10-23
html files may be .htm

BastienDurel:


yhirose:
@BastienDurel, thanks for your contribution!

Issue #248, opened 2019-10-26
Per Request read timeout


Sil3ntStorm:
The only option to set a read timeout is at compile time using the `CPPHTTPLIB_READ_TIMEOUT_SECOND` and `CPPHTTLIB_READ_TIMEOUT_USECOND` defines.  
This is incredibly lacking, as there are times where you would want to set the read timeout to a higher value for long running server requests, but not block unnecessarily long for other requests that are expected to return quickly.

yhirose:
@Sil3ntStorm, thank you for the feedback!
Would you like to set read timeout value as parameters in client APIs such as `Get`? If so, it's not too difficult to implement.
Or are you talking about changing read timeout value per each server handler? In this case, i couldn't find any easy way to do it...

Sil3ntStorm:
Talking about the client part of this yes.  
My first idea was also to add a `read_timeout` parameter to the `Get`, `Post` etc. functions which seemed the most logical / convenient place. However the actual read timeout is in the (SSL)SocketStream::read function that gets called from stream_line_reader::getline, so it would end up getting passed around quite a bit 😞.

yhirose:
@Sil3ntStorm, I have pondered over where this feature can implemented. My conclusion is to add`set_read_timeout` method instead of added another parameter to `Get` and other client methods. Since most users don't need this feature, this approach seems to be better to me. This is similar to `set_keep_alive_max_count`. Hope this method will solve the issue you have.

tuxflo:
Sorry to comment on such an old issue, but how does this actually work? Do we need to set all three timeouts listed in the Readme?
```
cli.set_connection_timeout(0, 300000); // 300 milliseconds
cli.set_read_timeout(5, 0); // 5 seconds
cli.set_write_timeout(5, 0); // 5 seconds
```
I want to set the timeout to the maximum time `std::numeric_limits<time_t>::max();` to talk to an API which has a "wait for changes" parameter (etcd v2). And with the basic example code the request returns after 5 seconds even if I set something like this:
```
time_t max_time = std::numeric_limits<time_t>::max();
cli.set_connection_timeout(0, max_time);
cli.set_read_timeout(200, 0);
```

PR #252, opened 2019-10-27
Add read_timeout to Get and Post requests

Sil3ntStorm:
Possible solution for #248 

Let me know what you think (probably missed something) and if this is desirable for other methods as well.


yhirose:
@Sil3ntStorm, thanks for your suggestion. I have actually implemented in a different way. Please see my comment in #248.

PR #253, opened 2019-10-27
Add clang-format file

Sil3ntStorm:
Roughly in line with current style, with some tweaking


yhirose:
@Sil3ntStorm, thank you for the PR. But I am ok with the default behavior of clang-format.

PR #254, opened 2019-10-30
Fixed the warning using marcro

hoathienvu8x:
…ttplib::Client::send_with_content_provider(const char*, const char*, const Headers&, const string&, size_t, httplib::ContentProvider, const char*, bool)’:

./httplib.h:3176:70: warning: unused parameter ‘compress’ [-Wunused-parameter]
     ContentProvider content_provider, const char *content_type, bool compress) {
                                                                      ^~~~~~~~
fixed the warning using marcro

```
\#ifdef CPPHTTPLIB_ZLIB_SUPPORT
, bool compress = false
\#endif
```

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -622,78 +622,127 @@ public:
   std::shared_ptr<Response> Head(const char *path, const Headers &headers);
 
   std::shared_ptr<Response> Post(const char *path, const std::string &body,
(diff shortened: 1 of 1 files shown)


yhirose:
@hoathienvu8x, thank you for bringing up the issue! I fixed in a simpler way.

PR #257, opened 2019-11-01
Update README.md

danielzehe:
added return true to the content provider get example, doesn't compile without it


yhirose:
@danielzehe, thank you for your contribution!

PR #280, opened 2019-12-03
Whitespace trimming and libc++ compatibility workaround

matvore:
I found a crashing bug when sending malformed input to the header-parsing code. It is particular to libc++ (i.e. it doesn't repro with other std C++ library implementations)

The exact error I saw was:

libc++abi.dylib: terminating with uncaught exception of type std::__1::regex_error: The complexity of an attempted match against a regular expression exceeded a pre-set level.
Abort trap: 6

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -1357,6 +1357,26 @@ inline bool is_connection_error() {
 #endif
 }
 
+inline socket_t create_client_socket(
+    const char *host, int port, time_t timeout_sec) {
+  return create_socket(
+      host, port, [=](socket_t sock, struct addrinfo &ai) -> bool {
+        set_nonblocking(sock, true);
+
+        auto ret = ::connect(sock, ai.ai_addr, static_cast<int>(ai.ai_addrlen));
+        if (ret < 0) {
+          if (is_connection_error() ||
+              !wait_until_socket_is_ready(sock, timeout_sec, 0)) {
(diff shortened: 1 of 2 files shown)


yhirose:
@matvore, thanks for the contribution. I just took a look at the code a little bit. But I found that all the examples in example/ directory cannot be built anymore. This is due to the gtest `FRIEND_TEST` macro in httplib.h. If I added the following code in httplib.h, then I am now able to build though...

```cpp
#ifndef FRIEND_TEST
#define FRIEND_TEST(a, b)
#endif
```

Personally, I don't want to have this kind of code in httplib.h, because that is not directly related to the httplib.h functionalities. I understand that you wanted to test the private methods, but I just wonder if we could do so without touching httplib.h itself. I am thinking to make a test HTTP client class which feeds the bad test data described in your test cases. Anyway, thank you for finding the bug and showing the excellent solution!

matvore:
I've pushed a new version. This is more black-box, as you suggested. I do agree this is a better approach. However, I did expose some socket-creation logic - once internal to the Client class - so that it could be used in tests to trigger requests.

Note that I didn't create a test-Client class, but I did try to keep the redundant boilerplate to a reasonable amount.

yhirose:
@matvore, great! Thanks for your contribution!

matvore:
Thanks for the review :)

PR #283, opened 2019-12-06
Remove use of exceptions.

barryam3:
Some people, including myself, compile C++ code with exceptions disabled. I just tried upgrading to the newest version of this library and discovered that it's using exceptions now. But looking at the code, it seems there's only one exception use, and it's an implementation detail because the thrown exception is immediately caught. This change rewrites that section to avoid requiring people to have exceptions enabled.


yhirose:
@barryam3, thanks for the pull request, and it looks good to me. Thanks for your fine contribution!

PR #288, opened 2019-12-12
Fix multiple files upload issue

DJBOUBOU:
When multiple files are sent through POST request they all get concatenated in one file because name_ field is the same for all contents.


DJBOUBOU:
Fixed in last tag

PR #297, opened 2019-12-21
Fix redundant Progress copy in Get methods

hyperxor:
Hello, I have found sevetal redundant copy of `Progess` objects in `Get` methods. We can use moving here.


yhirose:
@hyperxor, thank you for your contribution!

hyperxor:
@yhirose you are welcome

PR #317, opened 2020-01-07
Tiny improvement in DataSink class

hyperxor:
Hello, I have decided to make a tiny improvement in `DataSink` class that you created recently. As far as I know copy construction and copy assignment should be in pair (the same thing with move construction and move assignment). Do you agree with me?


yhirose:
@hyperxor, thanks for the improvement!

hyperxor:
@yhirose you are welcome

PR #318, opened 2020-01-07
Remove redundant code

hyperxor:
In my humble opinion `new_task_queue` field is redundant and `ThreadPool` can be created directy in `Server::listen_internal` without using lambda function.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -509,8 +509,6 @@ public:
   bool is_running() const;
   void stop();
 
-  std::function<TaskQueue *(void)> new_task_queue;
-
 protected:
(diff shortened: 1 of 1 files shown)


yhirose:
@hyperxor, thanks for the pull request. The reason why I took the redundant way is to allow users to use their own thread pool library, because the default `ThreadPool` implementation may not satisfy their needs. Please see [this](https://github.com/yhirose/cpp-httplib#override-the-default-thread-pool-with-yours) and #198.

hyperxor:
@yhirose You are right, I see.

PR #322, opened 2020-01-08
Fix suppressed Response class objects moving

hyperxor:
I have found that each time when we want to move `Response` class objects copying operation is executed.  For instance,
```
...
    if (!process_and_close_socket(sock, requests.size() - i,
                                  [&](Stream &strm, bool last_connection,
                                      bool &connection_close) -> bool {
                                    auto &req = requests[i++];
                                    auto res = Response();
                                    auto ret = handle_request(strm, req, res,
                                                              last_connection,
                                                              connection_close);
                                    if (ret) {
                                      responses.emplace_back(std::move(res));
                                    }
...
```

There is an used defined `~Response()` destructor. It causes compiler to suppress the default move constructor/ assigment generation and as a result the optimization is missed. See more details [here](https://www.modernescpp.com/index.php/c-core-guidelines-constructors-assignments-and-desctructors).

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -300,7 +300,7 @@ struct Request {
 
 struct Response {
   std::string version;
-  int status;
+  int status = -1;
   Headers headers;
   std::string body;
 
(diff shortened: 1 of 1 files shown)


yhirose:
@hyperxor, thanks for another fine contribution!

PR #324, opened 2020-01-09
Use sleep_for and chrono for delays in tests

hyperxor:
Hello, @yhirose! What do you think about using `sleep_for` and `chrono` for delays in tests and deleting platform-dependent defines?


yhirose:
@hyperxor, it's a fine improvement. Thanks for your contribution!

PR #325, opened 2020-01-10
Properly implement BufferStream::read

rockwotj:
The existing implementation did not advance the buffer at all.


yhirose:
@rockwotj, oops, what a serious bug... Thanks for your contribution!

PR #326, opened 2020-01-10
Add status message for 202 Accepted

vvanelslande:


yhirose:
@vvanelslande, thanks for the contribution!

PR #333, opened 2020-01-14
std::thread::hardware_concurrency

stupedama:
If std::thread::hardware_concurrency is 0 use 2 threads. Also -1 thread because we already have one thread from the main function.


yhirose:
@stupedama, looks good to me. Thanks for your contribution!

PR #337, opened 2020-01-25
Update split.py

TheMaverickProgrammer:
now works with legacy Python

Added a version check before invoking os.makedirs(). Now this can run on older python versions as well as 3


TheMaverickProgrammer:
forgot something

PR #338, opened 2020-01-25
fixed missing import

TheMaverickProgrammer:
same as other pull (how do I cancel the other one?) but added missing `import sys`


yhirose:
@TheMaverickProgrammer, thanks for your contribution!

PR #340, opened 2020-01-28
HEAD support for static file server

vitalyster:


yhirose:
@vitalyster, thanks for your contribution!

PR #355, opened 2020-02-14
Added Client::Post with ContentReceiver param

evg82:
I do not know if this is allowed in HTTP standar, but i find it usefull in my project.

I added Client::Post than allow send static content and receive response with ContentReceiver. For example, in my project i send a post with json encoded information describing "topic" to susbcribe, then keep listening in this conection por SSE messages.

It works really well (client and server use this lib) and maybe useful for someone

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -628,6 +628,10 @@ public:
   std::shared_ptr<Response> Post(const char *path, const Headers &headers,
                                  const MultipartFormDataItems &items);
 
+  std::shared_ptr<Response> Post(const char *path, const Headers &headers,
+                                 const std::string &body, const char *content_type,
+                                 ContentReceiver content_receiver);
+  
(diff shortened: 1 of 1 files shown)


yhirose:
@evg82, thanks for the pull request. It seems acceptable. According to [this stackoverflow post](https://stackoverflow.com/questions/1226810/is-http-post-request-allowed-to-send-back-a-response-body).
The only thing is that the pull request doesn't come with a unit test. Could you provide at least one unit test in 'test/test.cc'? Then, I'll review it. Thanks!

evg82:
Im trying to write unit test code (i have no experience with test), however i see send is public in client. The function i need, and other combinations params/methods can be made without change the lib. This is enought for me porpuse. I do not know if is better to add it or keep lib clean.

PR #356, opened 2020-02-17
Fix a few shadowed variable compilation warnings.

samhocevar:
These warnings are minor but verbose (with `-Wshadow`), and I think fixing them may help with clarity, especially in `Response::set_content_provider` where it isn’t immediately obvious what `length` refers to.


yhirose:
@samhocevar, thanks for the fine refinement!

PR #358, opened 2020-02-20
Fix issue with Windows getnameinfo

brianclinkenbeard:
Converting from `size_t` to `DWORD` gives a compilation error: https://gist.github.com/bunnei/9c09b2853e8a3b5c7c66777d496192ca


yhirose:
@brianclinkenbeard, thanks for the report. As you can see my comment above, I fixed it by casting the value to `socklen_t`. Thanks for your contribution!

PR #360, opened 2020-02-21
Support for deflate compression

rymis:
As long as httplib has gzip support it is easy to add deflate. I did it with automatic encoding detection because some clients could compress content with GZip and to set deflate compression instead of correct gzip.

As a discussion starting point, I'd like to add ContentTranscoder abstraction to encode and decode custom streams, like LZ4, Snappy, etc.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -1667,14 +1667,15 @@ inline bool compress(std::string &content) {
 class decompressor {
 public:
   decompressor() {
+    std::memset(&strm, 0, sizeof(strm));
     strm.zalloc = Z_NULL;
     strm.zfree = Z_NULL;
     strm.opaque = Z_NULL;
 
-    // 15 is the value of wbits, which should be at the maximum possible value
-    // to ensure that any gzip stream can be decoded. The offset of 16 specifies
-    // that the stream to decompress will be formatted with a gzip wrapper.
-    is_valid_ = inflateInit2(&strm, 16 + 15) == Z_OK;
+    // 31 is the value of wbits, which should be used for compression auto
+    // detection. In such mode zlib will be able to decompress both gzip
+    // and deflate data.
+    is_valid_ = inflateInit2(&strm, 32 + 15) == Z_OK;
   }
 
   ~decompressor() { inflateEnd(&strm); }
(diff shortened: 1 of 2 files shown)


yhirose:
@rymis, thank you for the pull request. Could you explain more about "because some clients could compress content with GZip and to set deflate compression instead of correct gzip"? I would like to stick to the HTML spec unless the behavior is a widely accepted common practice. Thanks!

rymis:
Sorry, I thought a little and realized, I was wrong. I read that some client library was sending incorrect Content-Encoding somewhere on stackoverflow, but I think it is not wise to support this behavior. I'll fix the MR to be standard complient. The actual RFC looks like:
```
4.2.2.  Deflate Coding

   The "deflate" coding is a "zlib" data format [RFC1950] containing a
   "deflate" compressed data stream [RFC1951] that uses a combination of
   the Lempel-Ziv (LZ77) compression algorithm and Huffman coding.

      Note: Some non-conformant implementations send the "deflate"
      compressed data without the zlib wrapper.


4.2.3.  Gzip Coding

   The "gzip" coding is an LZ77 coding with a 32-bit Cyclic Redundancy
   Check (CRC) that is commonly produced by the gzip file compression
   program [RFC1952].  A recipient SHOULD consider "x-gzip" to be
   equivalent to "gzip".
```

httplib.h
```diff
@@ -1667,14 +1667,16 @@ inline bool compress(std::string &content) {
 class decompressor {
 public:
   decompressor() {
+    std::memset(&strm, 0, sizeof(strm));
     strm.zalloc = Z_NULL;
     strm.zfree = Z_NULL;
     strm.opaque = Z_NULL;
 
     // 15 is the value of wbits, which should be at the maximum possible value
     // to ensure that any gzip stream can be decoded. The offset of 16 specifies
```
yhirose: Since '32' is now used for the automatic stream type detection, this comment is no longer valid. Could you update it?

yhirose:
@rymis, I checked the revised code and looks good to me except the comment for `inflateInit2(&strm, 32 + 15)` . Could you update it? Thanks!

yhirose:
Also one more request. Could you squash all the commits into a commit, so that the commit history could be simpler?

httplib.h
rymis: Yes, for sure

rymis:
It's because of squashing of commits. It came from master :( I'll change the comment.

yhirose:
Actually I can squash the commits on my side. Thanks for the adjustments.

yhirose:
I have just merged it into the master. Thanks for your contribution!

rymis:
Thanks!

PR #363, opened 2020-02-24
Remove code duplication in converting params to query

hyperxor:
I have extracted a free function for converting `Params` to query string and added a unit test.


yhirose:
@hyperxor, thanks for the code improvement!

hyperxor:
@yhirose you are welcome

PR #368, opened 2020-02-28
Fix exception that occurs with libc++ regex engine

matvore:


yhirose:
Wow, fantastic! Thanks for the improvement.

PR #376, opened 2020-03-10
Need to catch handler exceptions

ralequi:
Hi,

In many situations a handler may throw an exception.
If this exceptions are not catch, all http-server process crash too.

This is the easiest way I have found to do this.

Tested and working.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -3462,14 +3462,23 @@ inline bool Server::routing(Request &req, Response &res, Stream &strm,
 
 inline bool Server::dispatch_request(Request &req, Response &res,
                                      Handlers &handlers) {
-  for (const auto &x : handlers) {
-    const auto &pattern = x.first;
-    const auto &handler = x.second;
 
-    if (std::regex_match(req.path, req.matches, pattern)) {
-      handler(req, res);
-      return true;
+  try {
+    for (const auto &x : handlers) {
+      const auto &pattern = x.first;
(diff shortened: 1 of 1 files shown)


yhirose:
@ralequi, thanks for bringing up the matter! The PR looks good. The only thing is that I don't want to send the custom header `EXCEPTION_WHAT` to client, because the exception message is internal information for the server, not for client. But t is helpful for server administrator. So it would be nice to output the message into log.

yhirose:
I manually merged your change. I made a change not to send the EXCEPTION_WHAT header to client, also added a unit test. Thanks for your contribution!

https://github.com/yhirose/cpp-httplib/commit/e07c5fec015c19ea96305bb8edc3859f8eabff0e
https://github.com/yhirose/cpp-httplib/commit/26deffe0c6a1e177e393416b94cdc0f340ca79bd

ralequi:
Thank you!
I was thinking on add the EXCEPTION_WHAT to the request headers instead of the response one for the same reason... however it had no much sense to do that.
The check you added is useful, but, it has small (or big, depending on the server usage) overhead as far you are comparing many strings on each response, which in fact is expensive in computing terms.

Maybe a cleaner way should be add an internal field to the response data, or maybe something more disruptive. Don't know what do you think.
If I thought in something interesting and cleaner, and have time to implement, I would make a new PR.

Nice project!

PR #377, opened 2020-03-10
Fix conversion to ‘int’ from ‘long int’ warning

miketsts:


yhirose:
@miketsts, thanks for your contribution!

PR #378, opened 2020-03-11
add poll size argument for Server() and SSLServer()

xinyu391:
some time default thread pool size ** std::thread::hardware_concurrency() - 1** is not enough.
especial for IO work, or GPU work,  It shoule  let user decied the thread pool size.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -48,10 +48,6 @@
 #define CPPHTTPLIB_RECV_BUFSIZ size_t(4096u)
 #endif
 
-#ifndef CPPHTTPLIB_THREAD_POOL_COUNT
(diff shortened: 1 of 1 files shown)


yhirose:
@xinyu391, thank you for your pull request. But I can't accept it, because I intentionally designed the Server constructor not to have any optional setting. I provided `set_???` functions instead. Please check `keep_alive_max_count_ ` member variable to see how I tread such option values. Hope it helps!

yhirose:
@xinyu391, thanks for the pull request, but I cannot accept it. `ThreadPool` is just a default thread pool implementation, and cpp-httplib actually doesn't depend on it at all. I think you already know about it because this design came from your suggestion at #198. Hope it helps!

PR #381, opened 2020-03-12
Prevent an implicit capture of 'this' via '[=]' (in C++20)

isnullxbh:


yhirose:
@isnullxbh, thanks for the pull request. I don't mind changing `[this]` to `[=, this]` everywhere directly instead of doing it when a C++20 compiler is used. So could you get red of the `#if __cplusplus > 201703L` condition? Also if we grep httplib.h, there are a number of places where `[this]` is used. Do you think if this pull request also need to handle them? Thanks!

isnullxbh:
Hello, @yhirose!

> changing [this] to [=, this] everywhere directly

This will result in new warnings if users using C++17 (or less). We should use [=, this] since C++20, but the same form of capture will resulting in warnings in C++17 (or less).

Before C++20:
```
// OK
[=](){ handle(event); /* foo - member-function, event - local variable */ }
```

But if we pass "this" we got the warning:
```
// warning: explicit by-copy capture of 'this' redundant with by-copy capture default
[=, this](){ handle(event); /* foo - member-function, event - local variable */ }
```

Since C++20:
```
// compiler generates warning "implicit capture of 'this' via '[=]' 
// is deprecated in C++20
[=](){ handle(event);  }
```

But if we pass this - all ok:
```
// OK
[=, this](){ handle(event);  }
```

> Also if we grep httplib.h, there are a number of places where [this] is used

I think that we should fix only those places, where we use "=" in a capture list, but at the same time, we use "this" inside a lambda.

Thanks for your library and excuse me for my English :)

yhirose:
@isnullxbh, thanks for the clear explanation. I fully now fully understand the situation. I'll merge it soon. Thanks for your contribution! My English is not good either. :)

PR #386, opened 2020-03-15
Update calculation formula for progress percentage

AristoChen:
As I tested with the progress callback function, the output result of the progress percentage will always be zero


yhirose:
@AristoChen, thank you for finding the bug!

PR #391, opened 2020-03-20
Accept content by value to allow moving

agasparovic-sabre:
Previously, calling set_content always resulted in 's' being copied. With this change, there will still be the same number of copies made (1) when doing `set_content(my_value, ...)`, but there will be no copies if a caller elects to do `set_content(std::move(value), ...)` or `set_content(some_function_that_returns_a_temporary(), ...)` instead.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -313,7 +313,7 @@ struct Response {
 
   void set_redirect(const char *url);
   void set_content(const char *s, size_t n, const char *content_type);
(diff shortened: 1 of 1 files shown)


yhirose:
@agasparovic-sabre, thank you for your contribution!

agasparovic-sabre:
Thank you for creating so many great libraries!!!

PR #395, opened 2020-03-22
Minor improvements in httplib classes

hyperxor:
I have made several minor improvements: get rid of unnecessary copy string copy (`set_boundary`) and decrease object size (`MultipartFormDataParser`).


yhirose:
@hyperxor, thanks for the contribution!

PR #399, opened 2020-03-24
option to close socket from another thread

mrpuzzler:
minimum changes, but now it's possible to interrupt connection

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -880,7 +880,7 @@ public:
 
 private:
   bool process_and_close_socket(
-      socket_t sock, size_t request_count,
+      size_t request_count,
       std::function<bool(Stream &strm, bool last_connection,
                          bool &connection_close)>
           callback) override;
@@ -4634,7 +4634,7 @@ inline long SSLClient::get_openssl_verify_result() const {
 inline SSL_CTX *SSLClient::ssl_context() const noexcept { return ctx_; }
 
 inline bool SSLClient::process_and_close_socket(
-    socket_t sock, size_t request_count,
+    size_t request_count,
     std::function<bool(Stream &strm, bool last_connection,
                        bool &connection_close)>
         callback) {
--- a/httplib.h
+++ b/httplib.h
@@ -713,6 +713,8 @@ public:
 
   void set_logger(Logger logger);
 
+  void close_socket();
+
 protected:
   bool process_request(Stream &strm, const Request &req, Response &res,
                        bool last_connection, bool &connection_close);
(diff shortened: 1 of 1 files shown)


yhirose:
@mrpuzzler, thank you for the pull request. I reviewed that code, but honestly I can't merge this code... Once the raw socket becomes exposed to outside world and can be accessed in different threads, you have to give proper protection to.

The Server is already doing it in `stop` method. Please see the following code:

https://github.com/yhirose/cpp-httplib/blob/dc13cde820621806e415034ae45163ebcc39b66e/httplib.h#L3007-L3014

I recommend you to call `shutdown` first, then call `close`, as well as the socket should be properly protected with `std::atomic`. In this way, we can gracefully shutdown/close socket.

https://stackoverflow.com/questions/4160347/close-vs-shutdown-socket
https://docs.microsoft.com/en-us/windows/win32/winsock/graceful-shutdown-linger-options-and-socket-closure-2

Also the most important thing is to add a test case to test/test.cc in order to confirm the change really works. You could do the similar by referring to the test case for `stop` is here:

https://github.com/yhirose/cpp-httplib/blob/dc13cde820621806e415034ae45163ebcc39b66e/test/test.cc#L2139-L2177

Thanks for your contribution and understanding!

mrpuzzler:
Actually I need something like here:

https://stackoverflow.com/questions/18267146/how-to-let-a-thread-which-blocks-on-recv-exit-gracefully

mrpuzzler:
Maybe it's possible to add some shutdown flag, so we can interrupt reading?

yhirose:
@mrpuzzler, the flag is already set in `detail::shutdown_socket` which is called by `Server::stop()`. Please see the following code:

https://github.com/yhirose/cpp-httplib/blob/dc13cde820621806e415034ae45163ebcc39b66e/httplib.h#L1360-L1366

mrpuzzler:
I need Client side.

mrpuzzler:
Possible to make same for Client?

mrpuzzler:
Looks like we have to change mode first:
https://www.codeproject.com/Questions/723382/How-to-close-a-blocking-recv-socket-from-another-t

yhirose:
@mrpuzzler, yes you can implement the feature for Client in the way that I explained above.

I recommend you to call `detail::shutdown_socket` first, then call `detail::close_socket`, as well as the socket should be properly protected with std::atomic. In this way, we can gracefully shutdown/close socket. Also don't forget to document it in README and add a corresponding test in test/test.cc.

After you revise this PR, I'll review it again. Thanks for your help!

mrpuzzler:
This is not a problem to call it as you described. It's hard to separate
socket into own field, as you pass it along all calls. It is possible you
make it as explicit field at Client class?

PR #402, opened 2020-03-26
add http status code 201 to show the right status message in return headers

SoenkeHeeren:
I need to return the http status code 201. As of now, the corresponding message is "Internal Server Error". I propose this code to show the right message for status code 201.

https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#2xx_Success


yhirose:
@SoenkeHeeren, thank you for your contribution!

PR #408, opened 2020-04-01
added quick-start in read me

skramm:
Hi,
Great library, just discovered but I think it is very useful.
I felt there was missing some "quick start" section in the readme, just added some lines of what is needed to make a Hello-world demo, if you think that can be useful (if not, no problem !)

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,22 @@ int main(void)
   svr.listen("localhost", 1234);
 }
 ```
+#### Quick start (Linux)
+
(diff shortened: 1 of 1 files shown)


yhirose:
@skramm, thank you for your suggestion. But since this is a header-only library, I think we don't need it for most of users. I would rather want to focus on showing what this library can do.

If we really need such a 'quick start' page, I would do it in a different page and put usages not only for Linux w/ g++ but also Windows w/ Visual Studio and so on. Then I would  just place the link to the page in README.

Anyway, thanks for your contribution!

PR #413, opened 2020-04-06
SSL_shutdown() only if not already closed by remote

madMAx43v3r:
This fixes SIGPIPE on SSL_shutdown() in case the connection has already been closed by the remote host.


yhirose:
@madMAx43v3r, excellent! I'll merge it to the master soon.

PR #416, opened 2020-04-07
Add steps to install httplib on Linux

kallydev:
This may be helpful to users who use cpp-httplib for the first time.


yhirose:
@kallydev, thanks for the PR. But since this library is a single file header-only library, it's obvious. I would like to put only really necessary information about this library. (#408 is also closed due to the same reason.) Thank you for your understanding!

kallydev:
okay, I get it. thank you. :)

Issue #417, opened 2020-04-07
Add support for DELETE request body


altaitai:
According to MDN web docs (https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/DELETE), the DELETE request may have a body. Hence, support should be added, i.e. I will open a PR with the following changes to implement new functionality:
- Add `HandlersForContentReader delete_handlers_for_content_reader_` private member to `Server`
- Add `Server & Delete(const char *pattern, HandlerWithContentReader handler);` overload to `Server`
- Dispatch request with content to new DELETE handler with content in `Server::routing`

yhirose:
@altaitai, thanks for your contribution! Could also add a corresponding test case in test/test.cc when you submit the PR? Thanks.

yhirose:
Fixed by #418.

PR #418, opened 2020-04-07
Added support for DELETE request body

altaitai:
Addresses issue #417 and adds DELETE test case with content receiver.


yhirose:
@altaitai, thanks for your contribution!

altaitai:
@yhirose well thank you for the fantastic library !

ben-sol:
DELETE request _without_ a body fails now.

PR #432, opened 2020-04-18
Pass certs and keys from memory

ottigeda:
I would like to be able to pass certificates and keys from memory to the SSLServer and SSLClient, rather than using them from files on disk.
Therefore I did the modifications to allow this.
I hope you could have a look at it add this to your really cool library ?
I did run your test-cases and everything seems to be fine.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -848,6 +848,9 @@ public:
             const char *client_ca_cert_file_path = nullptr,
             const char *client_ca_cert_dir_path = nullptr);
 
+  SSLServer(X509 *cert, EVP_PKEY *private_key,
+            X509_STORE *client_ca_cert_store = nullptr);
+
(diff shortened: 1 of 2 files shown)


yhirose:
@ottigeda, thank you for your contribution! I only accepts pull requests with corresponding unit tests in test/test.cc. Could you add a unit test for the pull request? You can refer to my test cases for the file based certificate and key there. Thank you!

ottigeda:
Makes lot of sense, just added an test case.
Do you think it is fine now ?

yhirose:
@ottigeda, excellent! The code looks very good to me. Thanks for your find contribution!

ottigeda:
Thanks a lot, happy you like my contribution!
Was the 1st pull request I did :-)

PR #435, opened 2020-04-20
SSLClient::set_ca_cert_store: mark as inline

ottigeda:
Sorry, but I believe I have forgotten to inline this one new method.


yhirose:
@ottigeda, thanks!

PR #440, opened 2020-04-22
Fix size_t convert to int WARNING with msvc14

zhangyfan:
Fix size_t convert to int WARNING with msvc14

Modify:
httplib.h SocketStream::read
httplib.h SocketStream::write

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -2908,12 +2908,12 @@ inline bool SocketStream::is_writable() const {
 }
 
 inline ssize_t SocketStream::read(char *ptr, size_t size) {
-  if (is_readable()) { return recv(sock_, ptr, size, 0); }
+  if (is_readable()) { return recv(sock_, ptr, (int)size, 0); }
   return -1;
 }
 
 inline ssize_t SocketStream::write(const char *ptr, size_t size) {
-  if (is_writable()) { return send(sock_, ptr, size, 0); }
+  if (is_writable()) { return send(sock_, ptr, (int)size, 0); }
   return -1;
 }
 


yhirose:
@zhangyfan, thanks for the pull request. But, it causes another waring on Mac OS.

```
In file included from test.cc:1:
../httplib.h:2911:48: warning: implicit conversion changes signedness: 'int' to 'size_t' (aka 'unsigned long') [-Wsign-conversion]
  if (is_readable()) { return recv(sock_, ptr, (int)size, 0); }
                              ~~~~             ^~~~~~~~~
../httplib.h:2916:48: warning: implicit conversion changes signedness: 'int' to 'size_t' (aka 'unsigned long') [-Wsign-conversion]
  if (is_writable()) { return send(sock_, ptr, (int)size, 0); }
                              ~~~~             ^~~~~~~~~
\\\

yhirose:
Also `static_cast` should be used.

zhangyfan:
> Also `static_cast` should be used.

i will test this case with macos the next day, thank you.

yhirose:
Linux/BSD uses `size_t` for `len`:
http://man7.org/linux/man-pages/man2/recv.2.html
https://www.freebsd.org/cgi/man.cgi?query=recv(2)

Winsock2 uses `int` for `len`:
https://docs.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-recv

When you make an adjustment, could you make sure that no warning is produced on a**ll the platforms (Linux, Mac OS, Windows)** since this is a cross-platform library.

Thanks for your great help!

PR #441, opened 2020-04-23
Do not throw exceptions when parsing request chunks

matvore:
detail::read_content_chunked was using std::stoul to parse the
hexadecimal chunk lengths for "Transfer-Encoding: chunked" requests.
This throws an exception if the string does not begin with any valid
digits. read_content_chunked is not called in the context of a try block
so this caused the process to terminate.

Rather than use exceptions, I opted for std::stroul, which is similar to
std::stoul but does not throw exceptions. Since malformed user input is
not particularly exceptional, and some projects are compiled without
exception support, this approach seems both more portable and more
correct.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -1889,9 +1889,16 @@ inline bool read_content_chunked(Stream &strm, ContentReceiver out) {
 
   if (!line_reader.getline()) { return false; }
 
-  auto chunk_len = std::stoul(line_reader.ptr(), 0, 16);
+  unsigned long chunk_len;
+  while (true) {
+    char *end_ptr;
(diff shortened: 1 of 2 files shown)


yhirose:
@matvore, thank you for the improvement. The only thing is that it doesn't check the overflow situation. Should it check if `chunk_len` is not `ULONG_MAX`?

https://en.cppreference.com/w/cpp/string/byte/strtoul

matvore:
A check for ULONG_MAX can't hurt. Sent https://github.com/yhirose/cpp-httplib/pull/444

PR #442, opened 2020-04-23
TaskQueue method to internal size adjust

evg82:
I use a custom TaskQueue, with variable number of workers, adding workers on demand is an easy task when new connection arrive (in enqueue function) however i need another funtion to be called even (or better) went no new connections arrives to reduce workers count. I only added a new virtual method in TaskQueue class to allow custom class to adjust workers size over time. Even if this methods is called frequenlty custom class can keep a "last_update" counter to check if need to adjust worker count or any other internal task. Without this function i need an external thread to make this adjust task.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -365,6 +365,7 @@ public:
   TaskQueue() = default;
   virtual ~TaskQueue() = default;
   virtual void enqueue(std::function<void()> fn) = 0;
+  virtual void queue_adjust() { };
   virtual void shutdown() = 0;
 };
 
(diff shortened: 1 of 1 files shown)


yhirose:
@evg82, thank you for your excellent contribution! I made one adjustment to the pull request to rename `queue_adjust` to `on_idle` which seems more suitable name to me. Thanks!

yhirose:
@evg82, I worked on the issue #488, and decided that the `on_idle` should be optional since most users don't need it and affect performance. So you now have to change the default value of `CPPHTTPLIB_IDLE_INTERVAL_SECOND ` or `CPPHTTPLIB_IDLE_INTERVAL_USECOND `, or `set_idle_interval` method in order to use the idle event feature. Hope it hleps.

PR #444, opened 2020-04-23
Fail to read a chunk if its length is >= ULONG_MAX

matvore:
We cannot trivially support such large chunks, and the maximum value
std::strtoul can parse accurately is ULONG_MAX-1. Error out early if the
length is longer than that.


yhirose:
@matvore, thanks for the quick response!

PR #446, opened 2020-04-24
fix warning about send/recv and size_t on 64-bit windows compiler

ottigeda:
Hi

I tried to fix 2 warnings I get when I am compiling httplib.h with Visual Studio 2019 - 64 bit compiler
Microsoft (R) C/C++ Optimizing Compiler Version 19.24.28319 for x64

httplib.h(2939): warning C4267: 'argument': conversion from 'size_t' to 'int', possible loss of data
httplib.h(2944): warning C4267: 'argument': conversion from 'size_t' to 'int', possible loss of data

because 'size' (which is of type size_t) is 64bit wide on this platform,
but send/recv even these days only take an int, which is 32bit wide.

What do you think about this ?

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -2936,12 +2936,20 @@ inline bool SocketStream::is_writable() const {
 }
 
 inline ssize_t SocketStream::read(char *ptr, size_t size) {
+#ifdef _WIN32
+  if (is_readable() && size <= std::numeric_limits<int>::max()) { return recv(sock_, ptr, static_cast<int>(size), 0); }
+#else
   if (is_readable()) { return recv(sock_, ptr, size, 0); }
+#endif
   return -1;
 }
 
 inline ssize_t SocketStream::write(const char *ptr, size_t size) {
+#ifdef _WIN32
+  if (is_writable() && size <= std::numeric_limits<int>::max()) { return send(sock_, ptr, static_cast<int>(size), 0); }
+#else
   if (is_writable()) { return send(sock_, ptr, size, 0); }
+#endif
   return -1;
 }
 
--- a/httplib.h
+++ b/httplib.h
@@ -2937,6 +2937,8 @@ inline bool SocketStream::is_writable() const {
 
 inline ssize_t SocketStream::read(char *ptr, size_t size) {
 #ifdef _WIN32
+#pragma warning( push )
+#pragma warning( disable : 4018 )
   if (is_readable() && size <= std::numeric_limits<int>::max()) { return recv(sock_, ptr, static_cast<int>(size), 0); }
 #else
   if (is_readable()) { return recv(sock_, ptr, size, 0); }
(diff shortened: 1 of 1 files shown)


ottigeda:
BTW: thanks for you latest addition for supporting remote_addr and port in client request, just was I was looking for :-)

ottigeda:
ok, seems clang does not like it. 
I do not see how an better portable version of this would look like ?
I believe you do not want to use old-style-c cast for this.

ottigeda:
and just now the check has turned from red to green.
Great, but I am a little bit confused ?

yhirose:
@ottigeda, I actually re-run the TravisCI process, and it passed all the test this time. (The previous job somehow failed. It's a probably  network timing issue.)

If you take a look at the log of the Mac OS build on the Github Action, it shows the following warnings:

```
../httplib.h:2939:91: warning: implicit conversion changes signedness: 'int' to 'size_t' (aka 'unsigned long') [-Wsign-conversion]
  if (is_readable() && size <= std::numeric_limits<int>::max()) { return recv(sock_, ptr, static_cast<int>(size), 0); }
                                                                         ~~~~             ^~~~~~~~~~~~~~~~~~~~~~
../httplib.h:2944:91: warning: implicit conversion changes signedness: 'int' to 'size_t' (aka 'unsigned long') [-Wsign-conversion]
  if (is_writable() && size <= std::numeric_limits<int>::max()) { return send(sock_, ptr, static_cast<int>(size), 0); }
                                                                         ~~~~             ^~~~~~~~~~~~~~~~~~~~~~
```

This is because the BSD version of `recv` and `send` takes `len` as `size_t`, whereas Winsock2 takes `len` as `int`. So we probably need to use `#ifdef`.

```cpp
#ifdef _WIN32
   // For Winsock2
#else
   // For others
#endif
```

Linux/BSD uses size_t for len:
http://man7.org/linux/man-pages/man2/recv.2.html
https://www.freebsd.org/cgi/man.cgi?query=recv(2)

Winsock2 uses int for len:
https://docs.microsoft.com/en-us/windows/win32/api/winsock/nf-winsock-recv

Thanks!

ottigeda:
understood, better like this ?

yhirose:
On the AppVeyor, now the following warnings are present.

```c:\projects\cpp-httplib\httplib.h(2940): warning C4018: '<=': signed/unsigned mismatch [C:\projects\cpp-httplib\test\test.vcxproj]
c:\projects\cpp-httplib\httplib.h(2949): warning C4018: '<=': signed/unsigned mismatch 
```

Should it be like this?

```cpp
if (is_readable() &&
    size <= static_cast<size_t>(std::numeric_limits<int>::max())) {
  return recv(sock_, ptr, static_cast<int>(size), 0);
}
```

ottigeda:
Hmm, I see no better way as to disable that warning.

As I believe the types of source and destination are a given, 
it is what I intended to do: casting an size_t down to an int.
The code checks that value will fit into the destination variable first,
so I believe it should be safe to do this cast.

yhirose:
@ottigeda, thanks for the update, but I personally don't like to suppress warnings with `#pragma warning`. (You can see there are a number places where cpp-httplib uses `static_cast` to handle such type conversion warnings without using `#pragma`.

Anyway, I tried it also here #448 by making some adjustments to your original suggestion, and confirmed that no warning on all the environment. (It also resolves #440.) Thanks a lot!

PR #448, opened 2020-04-25
Fixed Visual Studio compiler warnings with x64 platform

yhirose:


yhirose:
This resolve #440 and #446.

PR #450, opened 2020-04-27
server: support dual-stack server socket

ottigeda:
According to RFC 3493 the socket option IPV6_V6ONLY
should be off by default, see
https://tools.ietf.org/html/rfc3493#page-22 (chapter 5.3).

However this does not seem to be the case on all systems.
For instance on any Windows OS, the option is on by default.

Therefore clear this option in order to allow
an server socket which can support IPv6 and IPv4 at the same time.

Would this change be acceptable to you ?
Or should I make it configurable, using an #define or an param on the server class ?

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -1469,6 +1469,11 @@ socket_t create_socket(const char *host, int port, Fn fn,
     int yes = 1;
     setsockopt(sock, SOL_SOCKET, SO_REUSEADDR, reinterpret_cast<char *>(&yes),
                sizeof(yes));
+
+    int no = 0;
+    setsockopt(sock, IPPROTO_IPV6, IPV6_V6ONLY, reinterpret_cast<char *>(&no),
+               sizeof(no));
+
 #ifdef SO_REUSEPORT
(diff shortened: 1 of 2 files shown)


ottigeda:
As it seems to me, the test on "Travis CI" can connect to "127.0.0.1" but not to "::1".
Do you know if that is a problem of my code, or is this rather related to the system the test is running on ? Do you know what could be the difference ?

yhirose:
@ottigeda, sorry for the late reply. I actually don't know why. But ipv6 support could be disabled on the TravisCI.

https://github.com/njh/travis-ipv6-test

ottigeda:
I think it should really be you to decide that. 
However in my opinion supporting  IPv6 becomes more important, since the IPv4 addresses are in short supply these days.
It is for sure still an long way to go for IPv6, but if the infrastructure does not support it,
it is difficult to make progress in applications to make them support it.

yhirose:
@ottigeda, I removed TravisCI test. Thank you for your contribution!

PR #451, opened 2020-04-27
Sketch handling EINTR errors

je-ik:
Syscalls might return EINTR in case of signals occurring. These cases can be handled by restarting the syscall. This PR fixes the problem in my application, but it might be good idea to add the EINTR handling in all syscalls (I added just the cases that were problematic in my case). The PR is only a proposal, there might be better solutions to the problem than I outlined.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -172,6 +172,18 @@ inline const unsigned char *ASN1_STRING_get0_data(const ASN1_STRING *asn1) {
 #endif
 #endif
 
+#define HANDLE_EINTR(method, ...) \
+  ({int res; \
+  while (true) { \
+    res = method(__VA_ARGS__); \
+    if (res < 0 && errno == EINTR) { \
+      continue; \
+    } else { \
+      break; \
+    } \
+  } \
+  res;})
+
 #ifdef CPPHTTPLIB_ZLIB_SUPPORT
 #include <zlib.h>
 #endif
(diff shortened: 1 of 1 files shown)


yhirose:
@je-ik, thanks for the fine contribution! I have just merged it to the master.

I made one adjustment to use inline function instead of macro, so that we can debug the code easier. https://github.com/yhirose/cpp-httplib/commit/8333340e2cdf370d9c6ac4020e5fcaa1b2ece1da

je-ik:
@yhirose :+1:
just to mention, the PR was not meant to be complete. There probably should be the EINTR handling also for `select()` code paths. Preferable for all syscalls (send, recv, getsockopt, ...). Although it is more essential for blocking cases, because there is much higher probability of encountering it.

yhirose:
@je-ik, thanks for the explanation. Since I personally have not experienced the issue, I don't know how much extent this treatment need to be applied to. It sounds good to me to apply it to select, send and recv, but it is too much for getsockopt?

Also I would like to know what situation you see the issue in, also how often other users might experience it. If it's a pretty rare case, I am thinking to make this feature optional with '#define CPP_HTTPLIB_SYSTEM_CALL_RETRY' or 'Server::set_system_call_retry'.

je-ik:
In my environment the issue appeared when the HTTP client was used in a newly created thread (std::thread) in static initializer. The thread was then detached. It seems that linux can under some conditions send SIGUSR1, which cause the poll call to fail and the connection to break. I pretty much think, that applying the retry only on blocking syscalls is pretty much enough. I'm not aware of any downsides of applying it to all syscalls, though (only that is might be time consuming to implement that).

yhirose:
@je-ik, thanks for the detail. I applied HANDLE_EINTR to `send` and `select` system calls as well.
Thank you for your great contribution!

PR #457, opened 2020-05-01
Fix crash caused by header field regex complexity

matvore:


yhirose:
@matvore, thank you for finding the critical problem!

PR #461, opened 2020-05-06
keepalive: support multiple post using content provider

ottigeda:
I would like to use your keepalive functionality in combination with an content provider doing POST.
It greatly increases the performance of my functionality :-)

I extended the related testcase with an simple additional step.
What do you think ?


yhirose:
@ottigeda, looks good to me. Thanks for your contribution!

PR #467, opened 2020-05-09
Fix #465

### LINKED ISSUE #465: Digest authentication error
The authentication code has error, the username is hardcoded to 'hello'.

  auto field = "Digest username=\"hello\", realm=\"" + auth.at("realm") +
               "\", nonce=\"" + auth.at("nonce") + "\", uri=\"" + req.path +
               "\", algorithm=" + algo + ", qop=" + qop + ", nc=\"" + nc +
               "\", cnonce=\"" + cnonce + "\", response=\"" + response + "\"";

changed to below and it works.
  auto field = "Digest username=\"" + username + "\", realm=\"" + auth.at("realm") +
               "\", nonce=\"" + auth.at("nonce") + "\", uri=\"" + req.path +
               "\", algorithm=" + algo + ", qop=" + qop + ", nc=\"" + nc +
               "\", cnonce=\"" + cnonce + "\", response=\"" + response + "\"";

PixlRainbow:
update digest header username to use username parameter instead of "hello" test value

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -2733,7 +2733,7 @@ inline std::pair<std::string, std::string> make_digest_authentication_header(
                  ":" + qop + ":" + H(A2));
   }
 
-  auto field = "Digest username=\"hello\", realm=\"" + auth.at("realm") +
+  auto field = "Digest username=\"" + username + "\", realm=\"" + auth.at("realm") +
                "\", nonce=\"" + auth.at("nonce") + "\", uri=\"" + req.path +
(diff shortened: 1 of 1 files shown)


PixlRainbow:
testing using google chrome at https://httpbin.org/digest-auth/auth/hello/world shows that httpbin does not correctly reject nonexistent user, only checks for password, which is why test is failing.

yhirose:
@PixlRainbow, thanks for the fix! I acknowledged the bug in httpbin.org. Until httpbin.org fixes the bug, I'll comment out the check.

Issue #421, opened 2020-04-09
using httplib cmake target


guteksan:
Hi,
I am having a problem with linking httplib library to my project.
I have built httplib and installed it in a certain destination <httplib-install>

My CMakeList.txt contains:
```
find_package(httplib CONFIG REQUIRED)
...
add_executable(${PROJECT_NAME} ${SRC})
target_link_libraries(${PROJECT_NAME} PUBLIC httplib)
```

This works, and the Cmake cache contains valid entry for httplib_DIR where the httplibConfig.cmake is located. But when I try to compile my project, it can't find the httplib.h, which I include. I looked into httplibConfig.cmake and there is created an INTERFACE target httplib (without any namespace), so I believe linking targets it the way I did should be enough to pass all directories. Could you please point out what is missing?

yhirose:
@guteksan, thank you for the feedback. Did you install httplib via CMakeLists.txt in this project? If so, this CMakeLists.txt has been contributed by other user, and I am not familiar to what it actually do (, even I am thinking of removing the file from this project).

Since this library is a 'header-only' library, it doesn't need to link anything. So once you manually copy httplib.h to one of include directories in your system, it much be compiled without any other treatment.

Hope it help!

guteksan:
Hi, @yhirose , thank you for prompt reply.
Yes, I did use CMakeLists.txt from this project to build and install the library. It went smooth, so I thought it is the right way, 
I know that header only libraries do not need to be linked, anyway I always prefered to separate my original code from the 3rd party dependencies, and using targets installed in the system is a nice way to do it. You might want to have a look at the Eigen library, which is also header only and allows to be used as targets in own projects.

yhirose:
@guteksan, if "It went smooth, so I thought it is the right way" is true,  what do you think causes the problem, in CMakeLists.txt in cpp-httplib, or the one in your project?

Since I haven't used the CMakeLists.txt at all and even tested it, I am tending to remove it from this repo though...

PixlRainbow:
Header only libraries on cmake are not linked against. Instead `target_include_directory` is used with the found directory.

guteksan:
[...]

PR #470, opened 2020-05-10
Bringing Cmake back

sum01:
I find having Cmake support very useful, so I reverted the deletion, and fixed the issue someone had with it in #421 

Note: This PR namespaces the Cmake target, so it's now `httplib::httplib`

I also took the time to improve it in a couple ways.
* Added some Cmake options to allow for strict or relaxed dependency requirements.<sup>[1]</sup>
* Added in a way for it to properly find its (optional) dependencies for the user when `find_package(httplib)` is called so they don't have to remember to do that.
* Added a way for users to require a version in the `find_pacakge` call.<sup>[2]</sup>
* Fixed the wrong install path that was causing issue #421 
* Changed the install path to allow for end-users to change it as desired.<sup>[3]</sup>
* Added an alias target to `httplib::httplib` so people building in-tree can link against the same target as users who installed.
* Added additional compiler feature detection. This is pretty minor, just might catch a few rare failed cmake builds on old compiler versions.
* Added some documentation on all this to the README

---

[1]. Specifically the following:
- `HTTPLIB_REQUIRE_OPENSSL` (default off)
- `HTTPLIB_REQUIRE_ZLIB` (default off)
- `HTTPLIB_USE_OPENSSL_IF_AVAILABLE` (default on)
- `HTTPLIB_USE_ZLIB_IF_AVAILABLE` (default on)

[2]. I was unsure if you follow [SemVer](https://semver.org/) or not, so I made the versioning allow anything >= request version.

[3]. See [GNUInstallDirs](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html), specifically `CMAKE_INSTALL_INCLUDEDIR`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,29 @@
+cmake_minimum_required(VERSION 3.7.0)
+project(httplib)
+
+set(CMAKE_CXX_STANDARD 11)
+
+# Include
+include(GNUInstallDirs)
+include(ExternalProject)
+
+add_library(${PROJECT_NAME} INTERFACE)
+target_compile_features(${PROJECT_NAME} INTERFACE cxx_std_11)
+
+target_include_directories(${PROJECT_NAME} INTERFACE
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
+    $<INSTALL_INTERFACE:include>)
+
+install(TARGETS ${PROJECT_NAME} EXPORT httplibConfig
+    ARCHIVE  DESTINATION ${CMAKE_INSTALL_LIBDIR}
+    LIBRARY  DESTINATION ${CMAKE_INSTALL_LIBDIR}
+    RUNTIME  DESTINATION ${CMAKE_INSTALL_BINDIR})
+
+install(FILES httplib.h DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/${PROJECT_NAME})
+
+install(EXPORT httplibConfig DESTINATION share/httplib/cmake)
+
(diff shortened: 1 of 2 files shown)


sum01:
Anything you want me to improve upon with this PR?

I think it doesn't really have any more to add to it, but I'm open to suggestions.

yhirose:
@sum01, thank you very much for spending time for CMake support. I don't come up with any suggestion since I personally don't know much about CMake. But it looks good and complete to me.

I would like to ask you to handle the following two things before merge the PR:

1. Remove the version description (`VERSION 0.6.3`) from CMakeLIsts.txt
2. Revert the change in README

The reason why I am asking these things is to reduce necessary maintenance work as much as possible because I don't personally use it and I don't have time to support users about CMake. (That's why I removed the previous CMake support.)

When it comes to README, I would like to put only information about features and examples in README. (I have actually received similar pull requests such as #416, #408, #195. But I didn't accepted any of them with the same reason.)

I appreciate your understanding and thank you very much for your great job!

sum01:
I'm a little bummed that you don't want the version string, as it allows specific version requirements in a `find_package` call, but very well.

As for the README, would it make sense to put that info somewhere else?
It seems odd to leave no info at all.

sum01:
Ok I removed the version file & README changes.

The only thing that might ever need changing is the OpenSSL minimum version string, which I doubt you'll be doing often anyways.

yhirose:
@sum01, thank you for the adjustments even if you don't fully agree with.

The OpenSSL minimum version string (1.1.1) is good with me. Also, it's actually a good idea to include the installation guide in the CMakeLists.txt file itself, so that users don't need to look around in README to find the info.

Hope it helps many users who depend on cmake. Thank you so much for the fine contribution!

PR #472, opened 2020-05-11
add missing 'done' in write_request

ottigeda:
DataSink::done was not assigned in write_request. 
Therefore it was not possible to call "done" in an content provider. 
I expected that should be the way to go, in case of failure to provide the content.
As an example, this could happen because the file to provide has vanished from disk in mean time. 
I tried to implement it the same way as in other places, I hope it makes sense to you ?

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -4112,7 +4112,9 @@ inline bool Client::write_request(Stream &strm, const Request &req,
         written_length = strm.write(d, l);
         offset += static_cast<size_t>(written_length);
       };
-      data_sink.is_writable = [&](void) { return strm.is_writable(); };
+      data_sink.is_writable = [&](void) {
+        return strm.is_writable() && written_length >= 0;
+      };
       data_sink.done = [&](void) { written_length = -1; };
 
       while (offset < end_offset) {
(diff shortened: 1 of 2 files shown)


yhirose:
@ottigeda, thank you for your contribution! The code looks almost good to me. Could you handle two things? Than, I'll merge it. Thanks!

1) Change `data_sink.is_writable`  implementation as below?

```cpp
    data_sink.is_writable = [&](void) {
      return strm.is_writable() && written_length >= 0;
    };
```

2) Apply the same change to the following section?

https://github.com/yhirose/cpp-httplib/blob/ba685dbe48e350944d00b588eb6d0e2868e3803d/httplib.h#L4138-L4160

ottigeda:
Is it that, what you wanted to have in addition ?

yhirose:
@ottigeda, thanks for the update. I have been pondering over this change, and came to remember that `done` is only used for stopping chunked transfer data stream. This is not for canceling or aborting the streaming process.

So instead of implementing `done` method in this case. I have rather added 'boolean return value' to `ContentProvider` type for this purpose. If we want to cancel the streaming process, the functor should return `false`. It actually corresponds to the behavior of `ContentReceiver`.

This is a breaking change to the ContentProvider interface. But it's necessary for the API consistency.

Sorry that I couldn't use your changes. But thank you very much for bringing this matter!

ottigeda:
No problem, your alternative solution is perfectly fine with me. Thanks a lot!

PR #475, opened 2020-05-12
 A detail about Gzip support

FFC12:
Just like for OpenSSL support (`libcrypto` and `libssl`), `libz` must be linked while building for Gzip support.


yhirose:
@FFC12, thanks for the corrections.

PR #477, opened 2020-05-13
make write timeout configurable (like the read timeout already is)

ottigeda:
In case we want to send a lot of data,
and the receiver is slower than the sender.

This will first fill up the receivers queues and after this
eventually also the senders queues,
until the socket is temporarily unable to accept more data to send.

select_write is done with an timeout of zero,
which makes the select call used always return immediately:
(see http://man7.org/linux/man-pages/man2/select.2.html)

This means that every marginal unavailability will make it return false
for is_writable and therefore httplib will immediately abort the transfer.

Therefore make this values configurable in the same way
as the read timeout already is.

Set the default write timeout to 5 seconds,
the same default value used for the read timeout.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -32,6 +32,14 @@
 #define CPPHTTPLIB_READ_TIMEOUT_USECOND 0
 #endif
 
+#ifndef CPPHTTPLIB_WRITE_TIMEOUT_SECOND
+#define CPPHTTPLIB_WRITE_TIMEOUT_SECOND 5
+#endif
+
+#ifndef CPPHTTPLIB_WRITE_TIMEOUT_USECOND
+#define CPPHTTPLIB_WRITE_TIMEOUT_USECOND 0
+#endif
+
 #ifndef CPPHTTPLIB_REQUEST_URI_MAX_LENGTH
 #define CPPHTTPLIB_REQUEST_URI_MAX_LENGTH 8192
 #endif
@@ -509,6 +517,7 @@ public:
 
   void set_keep_alive_max_count(size_t count);
   void set_read_timeout(time_t sec, time_t usec);
+  void set_write_timeout(time_t sec, time_t usec);
   void set_payload_max_length(size_t length);
 
   bool bind_to_port(const char *host, int port, int socket_flags = 0);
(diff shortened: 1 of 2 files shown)


ottigeda:
What do you think about it ?
However I find it difficult to add an test case for this one, as behavior of write buffers heavily depend on the hardware, the operating system and its current system load.
Could you help me a bit on what you would like to have in terms of test step ?

yhirose:
@ottigeda, looks good to me. When it comes to making a unit test, I actually don't know how to test it, since I have not encountered the situation ever. You may try to cause the situation with the content provider with huge data on server and content receiver with `std::this_thread::sleep_for` on client?

FFC12:
Even if we make timeout for write process, when the data is large, it does not send anything by resetting the connection. So it does not work. Maybe it should be investigated.

ottigeda:
Hi FFC12

I am sorry for this, but I do not understand what should be investigated exactly ?
Do you have some example code I could look at, or some step-by-step instructions on how to reproduce the issue ?

The intention of the write timeout I added is to allow the Operating System some time to empty its send-buffers in case they are full. 
Before this pull request, we just did not allow the OS any time.
Therefore any transfer was immediately aborted as soon as the OS reported that it could not accept more data to send at this time. 
The intention with this change is to be a bit more forgiving and try to send the data again later, without having to reset the connection. 
Of course we do not want to wait for forever, as this would cause an stalled transfer to never abort.
If the connection is reset for whatever reason, then the write timeout does not have any meaning anymore anyway I would think.

I believe it is basically the same idea as the read timeout has.

ottigeda:
Does this test meet your expectations ?
On my system (Windows 10) I need to post around 10MByte of data, in order to fill the write buffers.

yhirose:
@ottigeda, thanks for your work. I think the following test are enough:

1. Write huge data with appropriate timeout second (This test should succeed.)
2. Write huge data with 0 timeout second (This test should fail.)

I guess your test is for the first case, right? If so, could you make sure 2nd case is handled with your change?

Thank you!

ottigeda:
Yes of course you are correct, my test should now handle case 2 as well.
But I increased the size of the transmitted data to 64 MByte, because it seems 10 MByte was not enough to fill the write-buffers on MacOS.
I hope this test will be stable enough like this.

yhirose:
@ottigeda, thanks for all your efforts! I have just merged it into the master.

PR #480, opened 2020-05-14
Passing progress callback when reading chunked content

jraynal:
Hi there,

I am still working on server-push mjpeg streaming, and I was looking for a way to control when my chunks are received in full without changing assumptions on how the library behave or doing extra parsing.

I propose to pass the `progress` callback to chunks in order to do so.

An alternative would be to not call the user callback until the chunk has been fully parsed, but that would require at least one heap allocation.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -1950,7 +1950,7 @@ inline bool read_content_without_length(Stream &strm, ContentReceiver out) {
   return true;
 }
 
-inline bool read_content_chunked(Stream &strm, ContentReceiver out) {
+inline bool read_content_chunked(Stream &strm, Progress progress, ContentReceiver out) {
   const auto bufsiz = 16;
   char buf[bufsiz];
 
@@ -1969,7 +1969,7 @@ inline bool read_content_chunked(Stream &strm, ContentReceiver out) {
 
     if (chunk_len == 0) { break; }
 
-    if (!read_content_with_length(strm, chunk_len, nullptr, out)) {
+    if (!read_content_with_length(strm, chunk_len, progress, out)) {
       return false;
     }
 
@@ -2029,7 +2029,7 @@ bool read_content(Stream &strm, T &x, size_t payload_max_length, int &status,
   auto exceed_payload_max_length = false;
 
   if (is_chunked_transfer_encoding(x.headers)) {
-    ret = read_content_chunked(strm, out);
+    ret = read_content_chunked(strm, progress, out);
   } else if (!has_header(x.headers, "Content-Length")) {
     ret = read_content_without_length(strm, out);
   } else {
--- a/test/test.cc
+++ b/test/test.cc
@@ -323,18 +323,19 @@ TEST(ChunkedEncodingTest, WithResponseHandlerAndContentReceiverAndProgress) {
 
 #ifdef CPPHTTPLIB_OPENSSL_SUPPORT
   auto port = 443;
-  httplib::SSLServer server;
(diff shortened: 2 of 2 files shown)


yhirose:
@jraynal, the change looks good to me. Could you add a simple unit test in `test/test.cc`?

jraynal:
I think I inadvertently pushed an infinite loop for Unix systems. Sorry.

Also it occurs to me that my PR may break the semantics of the Progress callback, as one would expect to show the progress of the pulling of a full entity. It seems that the better way to do what I intend is to allocate a dynamic array for each chunk that would be the size of the current chunk. Thoughts?

yhirose:
@jraynal, you are right. The progress callback doesn't work with the chunked data transfer, because we cannot know total size of content.

I probably don't fully understand what you are trying to do. Could you explain what exactly you want to do? Also what does it mean by 'allocate a dynamic array for each chunk'?

jraynal:
It's the second time I may be asking something that is not quite right, thanks for bearing with me.

My needs are really close to the test I pushed in this PR: I am sending big chunks of data on the wire, bigger than the default receive buffer size `CPPHTTPLIB_RECV_BUFSIZ`. The chunks are send are conveniently JPEG images, and are of unknown sizes, so I can't just hardcode a bigger buffer size.

I am using formats that should tell me when a message begin and end, `mutlipart/x-mixed-replace` and `text/event-stream`, however it is convenient to send the whole message as a chunk, and use the chunk protocol specification to read the correct size.

My current submission allows you to measure the progress of the reception of a chunk bigger than `CPPHTTPLIB_RECV_BUFSIZ`. Which works fine for my purpose, but does not seem to be conform to the library design (progress should be for the reception of the whole object).

The alternative would be to replace the `CPPPHTTPLIB_RECV_BUFSIZE` when reading a chunk, and instead using a heap allocated array of the size of the chunk. So https://github.com/yhirose/cpp-httplib/blob/master/httplib.h#L1913-L1933 could become:

```cpp
inline bool read_content_with_length(Stream &strm, uint64_t len,
                                     Progress progress, ContentReceiver out) {
  auto buf = std::make_unique<char[]>(len);
  uint64_t r = 0;
  while (r < len) {
    auto read_len = static_cast<size_t>(len - r);
    auto n = strm.read(&buf[r], (std::min)(read_len, len));
    if (n <= 0) { return false; }

    if (!out(buf, static_cast<size_t>(n))) { return false; }

    r += static_cast<uint64_t>(n);

    if (progress) {
      if (!progress(r, len)) { return false; }
    }
  }

  return true;
}
```

Something of that nature would work for me, but force one heap allocation per chunk. I do not know what kind of performance impact it would have for other users, it would probably depends on their use case. An intermediate solution could be designed to be in between, like a sort of "small array optimization" that would use `CPPPHTTPLIB_RECV_BUFSIZE` if the `len` is smaller, or allocate an array otherwise.

Of course, the later possibility is to forget about this PR and just parse the data to detect the end of my messages using my protocol conventions.

yhirose:
@jraynal, thank you for the clear explanation. I now understand what 'allocate a dynamic array for each chunk' means and your intention with it. I am not going to merge your solution due to not only performance impact as you mentioned, but also other concerns. For instance, what if a bad user send a number of chunks of huge gigabytes?  What if a user wants to access received data as soon as possible even if whole data has not come yet?

Anyway, I think the dynamic array solution should work great in your situation. Or if you know the largest image size, you can set the size or larger to `CPPHTTPLIB_RECV_BUFSIZ`. It doesn't require any code change.

Hope your project goes well!

PR #481, opened 2020-05-14
Add {Client,Client2}::set_decompress

vvanelslande:
I need the compressed body.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -800,6 +800,7 @@ protected:
 #endif
     follow_location_ = rhs.follow_location_;
     compress_ = rhs.compress_;
+    decompress_ = rhs.decompress_;
     interface_ = rhs.interface_;
     proxy_host_ = rhs.proxy_host_;
     proxy_port_ = rhs.proxy_port_;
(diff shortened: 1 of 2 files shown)


yhirose:
@vvanelslande, the change looks good to me except one place. Could you make sure the `decompress_` is copied in `copy_settings`? Also could you add a simple unit test in test/test.cc?

https://github.com/yhirose/cpp-httplib/blob/63a96aeb20d8e1ec6d35655f91083cb3f1c26a89/httplib.h#L784

vvanelslande:
1. done
2. no

yhirose:
@vvanelslande, it seems like the unit test on GitHub Action fails... After you fix this failure and add a unit test for `set_decompress `, then I'll merge it right away. Thanks!

vvanelslande:
idk why it fails

yhirose:
@vvanelslande, when you are still interested in the pull request, please fix the problem and add a simple unit test case as I always ask contributors. I'll close it for now until then. Thank you for your efforts!

PR #487, opened 2020-05-18
Add CPPHTTPLIB_DO_NOT_DECOMPRESS

vvanelslande:
if CPPHTTPLIB_DO_NOT_DECOMPRESS is defined, the library will not decompress bodies

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -2056,6 +2056,7 @@ bool read_content(Stream &strm, T &x, size_t payload_max_length, int &status,
     return receiver(buf, n);
   };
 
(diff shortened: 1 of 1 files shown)


yhirose:
@vvanelslande, thank you for the pull request, but I cannot accept it. I don't think it's a compile time decision, but it should be a runtime decision with a client method like `set_decompress_for_response_body`.

vvanelslande:
@yhirose I tried but it broke the tests and I don't know why

vvanelslande:
can you add `set_decompress_for_response_body`?

PR #493, opened 2020-05-24
Add HTTPLIB_COMPILE option to Cmake

sum01:
This option (default OFF) automatically splits the file (with split.py) into a header & source file, then compiles it as a shared/static library. This requires an installed Python v3 executable to work.

This also adds a HTTPLIB_IS_COMPILED boolean that's available after a find_package(httplib) call.

Note that the minimum Cmake version increased to 3.12 because of FindPython3. Hopefully this isn't a problem, as it's already 3 years old at this point.


yhirose:
@sum01, thank you for another fine contribution!

PR #505, opened 2020-05-28
Add automatic versioning to Cmake

sum01:
It pulls the version from the user-agent string in the header, so it will not need to be manually adjusted. This version file is installed so that you can check for a specific version with find_package(httplib)

Also added HTTPLIB_INCLUDE_DIR (root path without header name), and HTTPLIB_LIBRARY (only if compiled).

Added HTTPLIB_VERSION, and HTTPLIB_FOUND (although it's recommended to check if the target exists).

Updated CMakeLists documentation for all this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,11 +32,15 @@
 
 	-------------------------------------------------------------------------------
 
-	These three variables are available after you run find_package(httplib)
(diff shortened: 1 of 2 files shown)


yhirose:
@sum01, looks awesome!

PR #509, opened 2020-06-02
fix get value function

wanggao1990:
An wrong result may be returned by the functions ( `get_header_value`,  `get_param_value`)  when the parameter id specified >0.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -2268,9 +2268,10 @@ inline bool has_header(const Headers &headers, const char *key) {
 
 inline const char *get_header_value(const Headers &headers, const char *key,
                                     size_t id = 0, const char *def = nullptr) {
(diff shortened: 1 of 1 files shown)


wanggao1990:
Request req;
    req.headers.emplace("a", "1");
    req.headers.emplace("b", "2");
    req.headers.emplace("c", "3");
    req.headers.emplace("b", "4");

    auto res = req.get_header_value("b",2);


if the id is 2 which larger than the numers of element of key "b",  res may be 3 ( "c","3" ).

yhirose:
@wanggao1990, thanks for fixing the bug! Could you rename `itRange` to `rng` to match my (undocumented) coding standard? Then, I'll merge it right away.

PR #511, opened 2020-06-03
check for [[deprecated]] support via feature test macro

nioncode:
The [[deprecated]] specifier is a C++14 feature, so it might not always
be available on a C++11 compiler.


yhirose:
@nioncode, thanks!

PR #512, opened 2020-06-03
fix cast warning

nioncode:
This fixes a few casting warnings I had on my CI machines.


yhirose:
@nioncode, thanks for catching the bug!

PR #513, opened 2020-06-03
replace usage of [[deprecated]] with CPPHTTPLIB_DEPRECATED

nioncode:
Sorry, my previous PR missed this, since I was basing it off of v0.6.5 instead of master.


yhirose:
@nioncode, never mind. Thanks for the update.

PR #518, opened 2020-06-09
Add ability to parse the fragment part of an URI

ReiserFlorian:
Add ability to parse the fragment part of an URI (this is all stuff following the #-sign, e.g. http://foo.bar?var=value#anyfragment)

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -3693,7 +3693,7 @@ inline void Server::stop() {
 inline bool Server::parse_request_line(const char *s, Request &req) {
   const static std::regex re(
       "(GET|HEAD|POST|PUT|DELETE|CONNECT|OPTIONS|TRACE|PATCH|PRI) "
-      "(([^?]+)(?:\\?(.*?))?(?:#(.*?))?) (HTTP/1\\.[01])\r\n");
+      "(([^?#]+)(?:\\?(.*?))?(?:#(.*?))?) (HTTP/1\\.[01])\r\n");
 
   std::cmatch m;
   if (std::regex_match(s, m, re)) {
--- a/httplib.h
+++ b/httplib.h
@@ -338,6 +338,7 @@ struct Request {
   // for server
   std::string version;
   std::string target;
+  std::string fragment;
   Params params;
   MultipartFormDataMap files;
   Ranges ranges;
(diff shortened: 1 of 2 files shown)


yhirose:
@ReiserFlorian, thank you for the pull request. I have a question. In what situation, can a server receive a request with the fragment in an URI? According to my understanding, any browsers and curl don't include the hash fragment  when making a HTTP request.

ReiserFlorian:
Hi @yhirose,
maybe the browsers and curl don't propagate it, but I've used your httplib in combination with unittests to test a http REST client module, which sends URI requests generated by the POCO lib (see [https://pocoproject.org/docs/](https://pocoproject.org/docs/)). The POCO lib has its own URI generation module (see [https://pocoproject.org/docs/Poco.URI.html](https://pocoproject.org/docs/Poco.URI.html)), which passes the fragment part to the server. My unittest uses all possibilities to add a fragment (or non fragment) to an URI:
 * http://host.de
 * http://host.de#fragment
 * http://host.de?param=value
 * http://host.de?param=value#framgent
and it doesn't fail. So I guess, the fragment parsing implementation in the httplib should be fine.

ReiserFlorian:
hi, 
sure. Please give me a short while

gatopeich:
Since the fragment does not identify the resource I don't see why the server code should parse it by default.
Specially when it is ignoring the "?query" parameters which are more relevant.

The fact that `Poco::URI` supports parsing all parts of an URI does not mean they are all meaningful to an HTTP server.

gatopeich:
Furthermore `Poco::Net::HTTPRequestHandler` passes the URI as a plain `std::string` to the user's request handler. It is up to the user to parse it, as expected.
https://pocoproject.org/docs/Poco.Net.HTTPRequest.html#23479

ReiserFlorian:
@gatopeich 
I understand your reservations. But please consider, the following request string has lead to nonsense in the server:
`http://host.de?param=value#fragment -> has been resulted in the req.params entry param = value#fragment`
My pullrequest just ensures, that this will not happen. There is no harm done, if the fragement is not passed through the server (what seems to be the default). But because we have revealed at least one example by finding the POCO lib (where this is the case), we should handle this situation.

By the way, `?query` parameters are handled as you can verify by yourself by studying the following regex: `"(([^?]+)(?:\\?(.*?))?(?:#(.*?))?) (HTTP/1\\.[01])\r\n")`: The part `(?:\\?(.*?))?` parses the `?query` part of the URI and the part `(?:#(.*?))?` the fragement part of the URI. The parameters are than inserted into a std::multimap by calling `detail::parse_query_text(m[4], req.params);`

yhirose:
@ReiserFlorian, thanks for adding the unit test. I didn't more research about it. Please see the following pages:

https://github.com/curl/curl/issues/4412
https://sourceforge.net/p/curl/bugs/1159/
https://stackoverflow.com/questions/2181186/how-to-access-url-hash-fragment-from-a-django-request-object

I now feel that it is a bug of POCO library. CURL treaded it as a bug, and fixed it. I think the POCO library should so the same.
But before I make a decision, I'll ponder over it a bit more time. Thank you.

ReiserFlorian:
I agree, if it is really no standard behaviour, we shouldn't provide it in the server. I will consider to send a bug report to the POCO lib owners. Thank you for the investigation.

yhirose:
@ReiserFlorian, no problem. I learned more about this thing anyway, and thanks again for your efforts to prepare the code and  unit test. Hope POCO library will fix the issue soon. Thanks!

Issue #499, opened 2020-05-27
Better support for Keep-alive on the client side


gatopeich:
Not being able to keep a connection alive renders this library unusable for many projects.

The so-called "keep-alive Client API" (see #369) is not usable for anything beyond very basic testing of the server.
```
bool Client::send(const std::vector<Request> &requests, std::vector<Response> &responses)
```
This just sends a pre-canned batch of requests... Can't see how to use that in real life.

The client should just not send "Connection: close" by default. Let the user, or the server, decide what to do if it is too hard to do it in the client code.

In this case just don't add Connection headers if user doesn't. As usual, less is more.

yhirose:
@gatopeich, thanks for the feedback. I am sorry that I haven't found time to work on it so far... It also relates to the longstanding issue #6.

If you need the Keep-alive connection on client now, I recommend you to use libcurl which supports Keep-alive connection as default. It's anyway more robust, and it has more nice features. :)

Thanks for bringing it up again. I'll leave this issue open for a while.

P.S. I'll quote my comment in #369 for your reference:

> thank you for the detail. Yes, you are totally right! I understand the regular Client API should set 'keep-alive' to 'Connection' header. But it's difficult to implement to keep the same connection over more than one requests right now due to the defect of the regular Client API design. I would like to fix it, but it requires fundamental changes in this library. In the future, I may get back to this. Sorry for causing the inconvenience...
> 
> Because of that reason, I provided the 'Keep-Alive' Client API as a workaround. Of course, I understand that this workaround doesn't solve all the problems. @PixlRainbow has already mentioned a problem that the batch client API cannot solve in his comment #45 (comment). That's why I still cannot solve #45 (SSE (Sever-Sent Events) support on client). (As a matter of fact, it was pretty easy for me to handle 'Keep-Alive' connection in the server and the batch client APIs.)
> 
> Thank you for your consideration for the problem. :)

gatopeich:
Yep. It is a very promising library but Keep-alive should work by default.

Why don't you just leave it on the server's hands?

On the client side I would assume connections always keep alive unless the server sends `Connection: close` header, or the user adds `Connection: close` explicity.

Would that really require big changes?

gatopeich:
I don't get the resource management issue. You can always close the socket on destruction.

Honestly if your "Client" became a "ClientSession" (Client+KeepAlive), we would have here a very attractive HTTP library... Maintenance costs would increase due to the growth in user base!

[...]

PR #520, opened 2020-06-11
manual override for connection management

PixlRainbow:
experimental interface for advanced users, for #499


bluecataudio:
That's an interesting initiative, thanks! But doesn't this require an extra method on the client to manually close/reopen the socket?

bluecataudio:
This seems to work pretty well, thanks! May I suggest to add a call to stop() in the Client destructor to make sure that the socket is properly shutdown?

PR #521, opened 2020-06-15
Fix #499

### LINKED ISSUE #499: Better support for Keep-alive on the client side

yhirose:
New Keep-Alive support on Client which supersedes the old batch style Keep-Alive API.

### DIFF
--- a/README.md
+++ b/README.md
@@ -482,29 +482,15 @@ httplib::make_range_header({{0, 0}, {-1, 1}})        // 'Range: bytes=0-0, -1'
 ### Keep-Alive connection
 
 ```cpp
-cli.set_keep_alive_max_count(2); // Default is 5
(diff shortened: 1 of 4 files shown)


yhirose:
Manually merged to the master.

PR #534, opened 2020-06-21
Use OpenSSL default when certificate location not set

jp39:
Also, do not disable certificate verification when using default OpenSSL
certificate location. This is a pretty reasonable default setting for
security.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -5548,7 +5548,8 @@ inline bool SSLClient::initialize_ssl(Socket &socket) {
       socket.sock, ctx_, ctx_mutex_,
       [&](SSL *ssl) {
         if (ca_cert_file_path_.empty() && ca_cert_store_ == nullptr) {
-          SSL_CTX_set_verify(ctx_, SSL_VERIFY_NONE, nullptr);
+          SSL_CTX_set_verify(ctx_, SSL_VERIFY_PEER, nullptr);
+          SSL_CTX_set_default_verify_paths(ctx_);
         } else if (!ca_cert_file_path_.empty()) {
           if (!SSL_CTX_load_verify_locations(ctx_, ca_cert_file_path_.c_str(),
                                              nullptr)) {


yhirose:
@jp39, thanks for the pull request, but there are some test cases failed. (Some tests are failing right now due to [the httpbin.org issue](https://github.com/postmanlabs/httpbin/issues/617), but it seems like other failures are caused by this pull request.)

yhirose:
@jp39, any update about the build errors? Thanks!

jp39:
@yhirose, there are a few tests failing already for me even on the master branch:

```
[  FAILED  ] 3 tests, listed below:
[  FAILED  ] ChunkedEncodingTest.WithResponseHandlerAndContentReceiver
[  FAILED  ] BaseAuthTest.FromHTTPWatch
[  FAILED  ] DigestAuthTest.FromHTTPWatch

 3 FAILED TESTS
```

Now, with my patch, I see these failing:

```
[  FAILED  ] 8 tests, listed below:
[  FAILED  ] ServerTest.GetMethod302Redirect
[  FAILED  ] ServerTest.PostMethod1
[  FAILED  ] ServerTest.PostMethod2
[  FAILED  ] ServerTest.PutMethod3
[  FAILED  ] ServerTest.PostMethod303Redirect
[  FAILED  ] ServerTest.KeepAlive
[  FAILED  ] PayloadMaxLengthTest.ExceedLimit
[  FAILED  ] SSLClientTest.ServerCertificateVerification2

 8 FAILED TESTS
```

This is very curious because my change is only located in the SSLClient class. So I can't see why it would make the Server fail.

The  SSLClientTest.ServerCertificateVerification2 test should probably be removed though because it's asserting that the certificate verfication fails when doing a GET on a google URL.

yhirose:
@jp39, thanks for the report. But all unit tests on CIs (Github actions and AppVeyor) pass successfully.

jp39:
Yeah, I'll take a deeper look into it when I get the chance. Also, I intend to update my PR a little if you agree:

1) What do you think about enabling server certificate verification by default (`server_certificate_verification_` to `true`) ? I suppose we would need to update some of the test case, and some of the examples in the README too.

2) The way I understand what `SSL_CTX_set_verify` does (but I may be wrong), is that it forces the TLS handshake to fail if the server certificate verification fails, when set to SSL_VERIFY_PEER. But this is redundant with verifying the server certificate manually, as it's currently done in `SSLClient::initialize_ssl` when we call `SSL_get_verify_result`. What do you think of using `SSL_VERIFY_NONE` everywhere in `SSLClient::initialize_ssl` (unless my understanding of `SSL_CTX_set_verify` is not correct)  ?

yhirose:
@jp39, thanks for your attention to this issue.

> 1. What do you think about enabling server certificate verification by default (server_certificate_verification_ to true) ? I suppose we would need to update some of the test case, and some of the examples in the README too.

I basically agree with you. Only thing (but a big thing) is that users now have to `set_ca_cert_path` or `set_ca_cert_store' explicitly, and they have to change their existing code to match this new behavior.

To mitigate the migration pain, we may need to use the system default ca-bundle if a user doesn't call `set_ca_cert_path` or `set_ca_cert_store'. This is the curl's behavior. But it doesn't look easy to do in all the platform Mac, Windows, Unix/Linux OSs...

> 2. The way I understand what SSL_CTX_set_verify does (but I may be wrong), is that it forces the TLS handshake to fail if the server certificate verification fails, when set to SSL_VERIFY_PEER. But this is redundant with verifying the server certificate manually, as it's currently done in SSLClient::initialize_ssl when we call SSL_get_verify_result. What do you think of using SSL_VERIFY_NONE everywhere in SSLClient::initialize_ssl (unless my understanding of SSL_CTX_set_verify is not correct) ?

You may be correct.  I'll try to test if it works on my Mac as well.

yhirose:
@jp39, I just tried to implement both you suggested. Everything works on my Mac OS, but it doesn't on Ubuntu somehow. Also I'll check on Windows when I have time.

jp39:
@yhirose Thanks for looking into it.

PR #539, opened 2020-06-22
fix documentation typo

kleinron:
fix "adress" --> "address"


yhirose:
@kleinron, good catch!

PR #544, opened 2020-06-29
Fixed comparison of integers of different signs.

rundong08:
This fixed a warning of "comparison of integers of different signs" when compiled using clang with `-Wsign-compare`.


yhirose:
@rundong08, thanks for your contribution!

PR #550, opened 2020-07-01
fix-the-code-won't compile-with-sdl-checks-on

### LINKED ISSUE #547: The code won't compile in VS2015 with SDL checks on
Hello.
I'm trying to compile in VS2015 a code that uses your library and getting error C4996.

I created a new project from scratch in Visual Studio 2015 Update 3 and left 'SDL checks' checkbox on (security development lifetime; it's a set of restrictive rules the MS is enforcing on source code). I get the following error when trying to compile my code using your library:
```httplib.h(3592): error C4996: 'std::basic_string<char,std::char_traits<char>,std::allocator<char>>::copy': Call to 'std::basic_string::copy' with parameters that may be unsafe - this call relies on the caller to check that the passed values are correct. To disable this warning, use -D_SCL_SECURE_NO_WARNINGS. See documentation on how to use Visual C++ 'Checked Iterators'```
I know I can disable SDL but I'd rather keep it on.
Perhaps you could change the following condition to something else (I'm not sure what)?
https://github.com/yhirose/cpp-httplib/blob/3d47a5143024d73830d1e28c4978c7aeb8ad67bd/httplib.h#L3589-L3593
Please note that in VS2015 Update 3 `_MSC_VER` is defined as 1900 and the code tries to call 'unsafe' (in the sense MS putting in it) version of the `copy` method.

tsilia:
Fixes #547

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -3586,7 +3586,7 @@ inline bool BufferStream::is_readable() const { return true; }
 inline bool BufferStream::is_writable() const { return true; }
 
 inline ssize_t BufferStream::read(char *ptr, size_t size) {
-#if defined(_MSC_VER) && _MSC_VER < 1900
+#if defined(_MSC_VER) && _MSC_VER <= 1900
   auto len_read = buffer._Copy_s(ptr, size, size, position);
 #else
   auto len_read = buffer.copy(ptr, size, position);


yhirose:
@tsilia, thanks for your contribution!

PR #552, opened 2020-07-02
[CHG] solve SSL_read/SSl_write timeout blocking problem

hucongquan:
ssl_read/ssl_write core uses sockets for data transmission，so I think we can solve this problem by setting the socket timeout

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -1,4 +1,4 @@
-//
+﻿//
 //  httplib.h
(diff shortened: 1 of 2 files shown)


yhirose:
@hucongquan, thanks you for the pull request. I implemented the same fix in a simpler way, and it supports both client and server. Please see #556.

PR #568, opened 2020-07-15
include <wincrypt.h> as otherwise CertOpenSystemStoreW can not be found

ottigeda:
- visual studio 2019, version 16.6.3
- 64 bit target

It is not clear for me, why it seems only I run into this problem.


yhirose:
@ottigeda, thanks for the contribution!

PR #580, opened 2020-07-26
Fix Cmake build for MinGW

sum01:
Seems certain targets/hosts failed without these, as `_MSC_VER` is
undefined on MinGW, which caused the `#pragma comment(lib "libname")` to
fail.

Note that the build actually succeeds without linking `cryptui` as long as we link `crypt32`. I just added it because you were originally linking it, and I didn't really investigate the reason. But it might be able to be removed?


yhirose:
@sum01, thanks for the quick fix!

PR #589, opened 2020-07-30
Update httplib.h

ThePiso:
When you disconnect and reconnect from the network, your network stack rewrites and updates /etc/resolv.conf accordingly. This configuration file is needed by the DNS resolver in the C library. The C library reads the DNS configuration from /etc/resolv.conf the first time, and caches it. It doesn't check, with every lookup, if the contents of /etc/resolv.conf have changed.
the solution is to add a call to res_init(), defined in resolv.h.

It would seem that the getaddrinfo (...) function initializes the info only the first time.
With the use of the res_init () function one forces to reinitialize its internal structure.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -156,6 +156,7 @@ using socket_t = SOCKET;
 #include <ifaddrs.h>
 #include <netdb.h>
 #include <netinet/in.h>
+#include <resolv.h>
 #include <netinet/tcp.h>
 #ifdef CPPHTTPLIB_USE_POLL
 #include <poll.h>
@@ -1821,6 +1822,7 @@ socket_t create_socket(const char *host, int port, int socket_flags,
   auto service = std::to_string(port);
 
   if (getaddrinfo(host, service.c_str(), &hints, &result)) {
+    res_init();
     return INVALID_SOCKET;
   }
 


yhirose:
@ThePiso, thanks for your contribution! It seems like builds on macOS and Windows failed. Could you fix the build errors? (That's the difficult part of maintaining 'cross-platform' library. 😅 )

PixlRainbow:
the bug and the fix only apply to Linux, so the builds failed on other platforms because the fix is Linux-specific. The line changes were not correctly isolated inside the platform flags.

ThePiso:
@yhirose<mailto:notifications@github.com>


Unfortunately I am aware that the change is only for linux, but I have no way to try it on windows nor above all on mac.


________________________________
Da: yhirose <notifications@github.com>
Inviato: giovedì 30 luglio 2020 14:14
A: yhirose/cpp-httplib <cpp-httplib@noreply.github.com>
Cc: ThePiso <mirko.pisoni@hotmail.it>; Mention <mention@noreply.github.com>
Oggetto: Re: [yhirose/cpp-httplib] Update httplib.h (#589)


@ThePiso<https://github.com/ThePiso>, thanks for your contribution! It seems like builds on macOS and Windows failed. Could you fix the build errors? (That's the difficult part of maintaining 'cross-platform' library. 😅 )

PixlRainbow:
You don't need to. The mechanism of the bug, as you explained, is unique to Linux. You only need to surround your changes with #ifdef flags for Linux.

PR #593, opened 2020-07-31
Fix FindBrotli for static libs

sum01:
It wasn't linking them.


yhirose:
@sum01, thanks for the update!

PR #594, opened 2020-07-31
Fix for Cmake on systems without Git

sum01:
If you didn't have Git installed, execute_process never declared the
error variable, which led to it never parsing the header for a version.

So if Git wasn't installed, the version variable never got declared,
which caused errors.

This fixes that.


yhirose:
@sum01, thanks again for another update!

Issue #582, opened 2020-07-28
Brotli support in CMake


yhirose:

yhirose:
@sum01, I have recently added Brotli compression support on client which can be configured with `CPPHTTPLIB_BROTLI_SUPPORT`. It requires brotli libraries. (https://github.com/google/brotli)

If it possible to support it in the CMakeLists.txt as `CPPHTTPLIB_ZLIB_SUPPORT` when you have time?
Thank you for your help!

sum01:
@yhirose Do you need all 3 libs (decode, encode, and common) or just specific ones?

yhirose:
@sum01, currently, the library supports only decoder, but I am thinking to support encoder soon as well. So it would be nice to include 3 libraries. As for the minimum version, I have installed 1.0.7 with homebrew on macOS and confirmed it works. Thanks!

```[master] ~/Projects/cpp-httplib$ brew info brotli
brotli: stable 1.0.7 (bottled), HEAD
Generic-purpose lossless compression algorithm by Google
https://github.com/google/brotli
/usr/local/Cellar/brotli/1.0.7 (25 files, 2.2MB) *
  Poured from bottle on 2020-07-25 at 11:39:48
From: https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/brotli.rb
License: MIT
==> Dependencies
Build: cmake ✘
==> Options
--HEAD
	Install HEAD version
==> Analytics
install: 23,343 (30 days), 71,928 (90 days), 319,606 (365 days)
install-on-request: 378 (30 days), 1,261 (90 days), 5,550 (365 days)
build-error: 0 (30 days)
```

yhirose:
@sum01, thanks for the quick support! I have just merged the pull request.

ElonH:
I got a problem when `HTTPLIB_USE_BROTLI_IF_AVAILABLE` was `on`
```
CMake Error at /usr/local/share/cmake-3.17/Modules/FindPackageHandleStandardArgs.cmake:261 (message):
  No REQUIRED_VARS specified for FIND_PACKAGE_HANDLE_STANDARD_ARGS()
Call Stack (most recent call first):
  cmake/FindBrotli.cmake:173 (find_package_handle_standard_args)
  CMakeLists.txt:146 (find_package)
```

sum01:
@ElonH Hmm. Do you not have Brotli installed? I think I might see where that could happen if so.

ElonH:
[...]

PR #598, opened 2020-08-01
Fix FindBrotli when no Brotli installed

sum01:
Woops.

Ref https://github.com/yhirose/cpp-httplib/issues/582#issuecomment-667537002


yhirose:
@sum01, thank you!

PR #605, opened 2020-08-09
Fix nullptr_t issue

freesurfer-rge:
Clang complains that `nullptr_t` should be `std::nullptr_t


yhirose:
@freesurfer-rge, thanks for catching it!

PR #622, opened 2020-08-24
Fix cmake execute_process working directory

felix-kaestner:
Correct the working directory of the execute_process call, to get the version from git tags.
The working directory should be the the directory of this libary. Otherwise, if you include httplib
as a git submodule and call add_subdirectory, the execute_process command will be run in a wrong
directory leading to error.


yhirose:
@felix-kaestner, thank you for your contribution.
@sum01, when you have time, could you review the pull request?

sum01:
Looks good to me

yhirose:
@felix-kaestner, @sum01, thanks for your contribution!

PR #625, opened 2020-08-25
Fix typo in README

Fefer-Ivan:
In README.md in section about content provider without content length there is missing `sink.` for a `done()` member function call.

Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,7 +197,7 @@ svr.Get("/stream", [&](const Request &req, Response &res) {
         // prepare data...
         sink.write(data.data(), data.size());
       } else {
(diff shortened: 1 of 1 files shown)


Fefer-Ivan:
```
 [ RUN      ] ServerTest.SlowPost
test.cc:2051: Failure
Value of: !res
  Actual: false
Expected: true
[  FAILED  ] ServerTest.SlowPost (4463 ms)
```
Looks like the problem is not connected to my changes.

yhirose:
@Fefer-Ivan, thanks for catching it! The failure has nothing to do your change of course.

PR #626, opened 2020-08-25
Example code refactoring

willem640:
Moved the using namespace to global scope, as that's where it usually is, and also remove the void in int main(void) as that's not required in C++

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,9 +13,11 @@ Server Example
 ```c++
 #include <httplib.h>
 
-int main(void)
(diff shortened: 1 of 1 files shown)


yhirose:
@willem640, thank you for the pull request. Since compilers don't complain with the current README example, I'll leave it as it is. Thank you anyway!

willem640:
Aight, I was just changing it bc it occurred to me. Of course, it doesn't matter to the example itself.

Issue #624, opened 2020-08-25
Non-copyable but movable objects can't be used as ContentProviders


Fefer-Ivan:
Hi.

I see this as a problem, because we can't use std::unique_ptr and other non-copyable objects as captures in ContentProvider.

Maybe, it is possible to replace using ContentProvider = std::function<..> with a class with virtual operator() and make set_content_provider into function templates that accept anything callable (for backward compatibility)?
Also, it will make resource_releaser obsolete (because all releasing will be done in ContentProvider destructor), so maybe it will be possible to deprecate it.

If you think this is a good idea, I will make a pull request.

Thanks.

yhirose:
@Fefer-Ivan, thanks for your feedback. Could you show a code example that explains the problem?

Fefer-Ivan:
For example, if I have several threads that return chunks of response, I might want to make a ContentProvider functor that accepts a queue of futures and waits and writes them one by one 
``` 
  class AsyncContentProvider {
  public:
    AsyncContentProvider(
        std::queue<std::future<std::string>> data_queue
    ):
        data_queue_(std::move(data_queue)) {}

    bool operator() (size_t /*offset*/, DataSink& sink) {
      if (data_queue_.empty()) {
        sink.done();
        return true;
      }
      std::string data = data_queue_.front().get();
      data_queue_.pop();
      sink.write(data.c_str(), data.size());
    }

  private:
    std::queue<std::future<std::string>> data_queue_;
  };
```

But it won't compile because std::future is not copyable, but std::function require anything passed to it to have a copy constructor.

Also, if my ContentProvider contains a lot of data (for example, large vector with different places to get data from), I want to avoid it being copied needlessly.

PixlRainbow:
isn't there a referencing capture that will not copy objects?

Fefer-Ivan:
In case of capture by reference, user will have to keep ownership of the resources and release it in resource_releaser function, which works, but doesn't look like RAII, which is preferred way of handling stuff in modern C++, in my opinion. 

Also, shared_ptr will do the trick too.

yhirose:
@Fefer-Ivan, thank you for explaining your idea. I actually don't have any inconvenience with this interface. But I now understand what you are struggling with. Could you implement your idea and send me a pull request? Then, I'll review it. (Please don't forget to add a unit test.) Thanks!

yhirose:
@yhirose, I'll close for now, but a pull request from you is always welcome!

PR #627, opened 2020-08-25
Avoid copying of content provider if possible

Fefer-Ivan:
Right now all content providers are copied inside set_content_provider or set_chunked_content_provider calls.
I replaced explicit copying in capture with move, so it should avoid coping when possible.

Because I can't move-capture in C++11, I had to create ContentProviderAdapter class explicitly.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -428,6 +428,19 @@ struct Response {
   ContentProvider content_provider_;
   std::function<void()> content_provider_resource_releaser_;
   bool is_chunked_content_provider = false;
+
+  class ContentProviderAdapter {
+  public:
+    explicit ContentProviderAdapter(ContentProviderWithoutLength&& content_provider):
+        content_provider_(content_provider) {}
+
+    bool operator()(size_t offset, size_t, DataSink& sink) {
+      return content_provider_(offset, sink);
+    }
+
+  private:
+    ContentProviderWithoutLength content_provider_;
+  };
 };
 
 class Stream {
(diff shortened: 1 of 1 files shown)


yhirose:
@Fefer-Ivan, thanks for the pull request. The code looks very good to me. Could you add a unit test which cannot be compiled with the old httplib.h? (The test case doesn't need to have any test assertion.) Thanks!

Fefer-Ivan:
@yhirose, I would love to, but even after this changes, I can't pass move-only object as ContentProvider (aka std::function<bool(size_t offset, size_t length, DataSink &sink)>) because std::function requires it to be copyable: https://en.cppreference.com/w/cpp/utility/functional/function/function: `F must meet the requirements of Callable and CopyConstructible.`

I can add a test with a class, that throws an exception in copy constructor or counts the number of times copy constructor is called and asserts that it is zero in destructor, but neither option seems good to me.

yhirose:
> I would love to, but even after this changes, I can't pass move-only object as ContentProvider 

@Fefer-Ivan, does it mean that the changes in this pull request are not enough to accomplish your intention?

Fefer-Ivan:
@yhirose, yes, this changes are more like performance optimization, which is still looks good, so I decided to create a pull request.

yhirose:
@Fefer-Ivan, thanks for the clear explanation. (I thought this PR is directly for #624.) I think a unit test is not needed. Thanks!

PR #629, opened 2020-08-28
Fix query parsing issues

omjego:
httplib::detail::split wasn't doing proper boundary checks. Inputs like "/? " (query with only spaces) or "/?= =" were crashing the server cause of illegal memory access done by the process, creating a potential security vulnerability, as this could be leveraged to make a DoS attack.
Have added checks for such cases and test cases as well.


yhirose:
@omjego, the code looks good to me. Thanks for your contribution!

PR #632, opened 2020-08-31
Fix server crash caused due to regex complexity while matching headers.

omjego:
While parsing content-type header in multipart form request the server crashes due to the exhaustion of max iterations performed while matching the input string with content-type regex.
Have removed the regex which might use backtracking while matching and replaced it with manual string processing.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -2979,7 +2979,7 @@ public:
             std::smatch m;
             const std::string header_name = "content-type:";
             if (contains_header(header, header_name)) {
-              header.erase(header.begin(), header.begin() + 13);
+              header.erase(header.begin(), header.begin() + header_name.size());
               trim(header);
               file_.content_type = header;
             } else if (std::regex_match(header, m, re_content_disposition)) {
--- a/httplib.h
+++ b/httplib.h
@@ -281,7 +281,7 @@ public:
 private:
   class data_sink_streambuf : public std::streambuf {
   public:
-    data_sink_streambuf(DataSink &sink) : sink_(sink) {}
+    explicit data_sink_streambuf(DataSink &sink) : sink_(sink) {}
 
   protected:
     std::streamsize xsputn(const char *s, std::streamsize n) {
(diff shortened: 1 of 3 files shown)


yhirose:
@omjego, thanks for the pull request, but first, could you explain this problem more clearly, and show me an example header text that can cause the problem? Thanks!

httplib.h
```diff
@@ -2913,8 +2932,6 @@ class MultipartFormDataParser {
 
   template <typename T, typename U>
   bool parse(const char *buf, size_t n, T content_callback, U header_callback) {
-    static const std::regex re_content_type(R"(^Content-Type:\s*(.*?)\s*$)",
```
omjego, line 2916: `^Content-Type:\s*(.*?) \s*$`
The above regex tries to parse string of following type (in order mentioned), 
1. "Content-type:" 
2. list of spaces
3. zero or more characters (could be anything) - matched lazily 
4. List of trailing spaces 

If the part 3 has spaces in between e.g. (application/json   some-string   another-string ) then as it is being lazily matched, it will start considering the spaces in between as part of (4) but then when it encounters any non space character like ('s' from "some-string") it will go back and add it to (3)
This can cause issues if there are lots of spaces in between. And eventually crashes the server. I couldn't think of any simpler regex so added simple string processing instead.

omjego:
Hey @yhirose  sorry, I should've added an example in the description. Have added comment which explains the change.  Thank you.

httplib.h
```diff
@@ -2960,8 +2977,11 @@ class MultipartFormDataParser {
           auto header = buf_.substr(0, pos);
           {
             std::smatch m;
-            if (std::regex_match(header, m, re_content_type)) {
-              file_.content_type = m[1];
+            const std::string header_name = "content-type:";
+            if (contains_header(header, header_name)) {
+              header.erase(header.begin(), header.begin() + 13);
```
Fefer-Ivan, line 2982: Maybe replace magic number 13 with header_name.size()?
```suggestion
              header.erase(header.begin(), header.begin() + header_name.size());
```

yhirose:
@omjego, and @Fefer-Ivan, thanks for your congribution!

yhirose:
There are some warnings left. I'll fix these later.

PR #636, opened 2020-09-02
Fix gzip_decompressor in case of one chunk being exactly equal to buffer size

Fefer-Ivan:
Previously exit condition for inflate loop was strm_.avail_out != 0, which means, if after inflate we filled output buffer completely, then we should run inflate again.

In case when inflated data chunk size is equal to output buffer size, we would try to inflate again, but we already consumed all input data. This will result in inflate call returning Z_BUF_ERROR and decompress function returning false. If you think it is very improbable, I have triggered it reliably in out testing environment.

To avoid it, we should use while (strm_.avail_in > 0) as our loop condition, which means, inflate until we have consumed all available input.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -2266,7 +2266,7 @@ public:
     strm_.next_in = const_cast<Bytef *>(reinterpret_cast<const Bytef *>(data));
 
     std::array<char, 16384> buff{};
-    do {
+    while (strm_.avail_in > 0) {
       strm_.avail_out = buff.size();
       strm_.next_out = reinterpret_cast<Bytef *>(buff.data());
 
@@ -2281,7 +2281,7 @@ public:
       if (!callback(buff.data(), buff.size() - strm_.avail_out)) {
         return false;
       }
-    } while (strm_.avail_out == 0);
+    }
 
     return ret == Z_OK || ret == Z_STREAM_END;
   }
--- a/test/test.cc
+++ b/test/test.cc
@@ -1032,6 +1032,23 @@ protected:
                    },
                    [i] { delete i; });
              })
+        .Get("/streamed-chunked-large",
+             [&] (const Request & /*req*/, Response &res) {
+               auto i = new int(0);
+               res.set_chunked_content_provider(
+                   "text/plain",
+                   [i](size_t /*offset*/, DataSink &sink) {
+                     EXPECT_TRUE(sink.is_writable());
+                     if (*i < 3) {
+                       sink.os << get_large_test_chunk(*i);
+                     } else {
+                       sink.done();
+                     }
+                     (*i)++;
+                     return true;
+                   },
+                   [i] { delete i; });
+             })
         .Get("/streamed",
              [&](const Request & /*req*/, Response &res) {
                res.set_content_provider(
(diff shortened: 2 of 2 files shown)


yhirose:
@Fefer-Ivan, thank you for finding the problem. I just got back to the zlib example and checked again (https://zlib.net/zlib_how.html), and found some missing lines in my code. The following revised code is almost identical to the zlib example, and I think it can fix the problem that you found. What do you think?

```cpp
  bool decompress(const char *data, size_t data_length,
                  Callback callback) override {
    assert(is_valid_);

    int ret = Z_OK;

    strm_.avail_in = static_cast<decltype(strm_.avail_in)>(data_length);
    // ------- Add Begin --------
    if (strm_.avail_in == 0) {
      return true;
    }
    // -------- Add End --------
    strm_.next_in = const_cast<Bytef *>(reinterpret_cast<const Bytef *>(data));

    std::array<char, 16384> buff{};
    do {
      strm_.avail_out = buff.size();
      strm_.next_out = reinterpret_cast<Bytef *>(buff.data());

      ret = inflate(&strm_, Z_NO_FLUSH);
      assert(ret != Z_STREAM_ERROR);
      switch (ret) {
      case Z_NEED_DICT:
      case Z_DATA_ERROR:
      case Z_MEM_ERROR: inflateEnd(&strm_); return false;
      }

      if (!callback(buff.data(), buff.size() - strm_.avail_out)) {
        return false;
      }
    } while (strm_.avail_out == 0);

    return ret == Z_OK || ret == Z_STREAM_END;
  }
```

Fefer-Ivan:
@yhirose, 

I looked at your proposed solution and at zlib example.

Your proposed fix won't help, because the root of the problem is in do-while loop, not in initialization before. You can check it if you run the test from my pull request with your code.

About zlib example: in the example do-while loop will have a same effect: if a chunk of input will have a decompressed size exactly equal to buffer size, it will call inflate() an extra time without new input, which will return Z_BUF_ERROR. In the annotations it mentions Z_BUF_ERROR, but explains that it is not fatal and why it is not checked.

When we run out of input in decompress in httplib, an extra call to inflate (without any new input) will return Z_BUF_ERROR and decompress will return false because of this line: `return ret == Z_OK || ret == Z_STREAM_END;`

From https://zlib.net/zlib_how.html
> That deflate() call wasn't able to do anything, either consume input or produce output, and so it returns Z_BUF_ERROR. (See, I told you I'd cover this later.) However this is not a problem at all. Now we finally have the desired indication that deflate() is really done, and so we drop out of the inner loop to provide more input to deflate(). 

This is about deflate(), but it is same for inflate():

> Z_BUF_ERROR does not need to be checked for here, for the same reasons noted for def().

So another possible fix is to replace `return ret == Z_OK || ret == Z_STREAM_END;` with `return ret == Z_OK || ret == Z_STREAM_END || ret == Z_BUF_ERROR`

But the reason they check for avail_out == 0 and not avail_in > 0 is stated as following:
> The inner do-loop ends when inflate() has no more output as indicated by not filling the output buffer, just as for deflate(). In this case, we cannot assert that strm.avail_in will be zero, since the deflate stream may end before the file does. 

It is not the case for http response, because we can be sure that all data we received is a single gzip compressed stream, so my original solution still works.

Fefer-Ivan:
I personally think while loop is preferred because solution with do-while loop and ret == Z_BUF_ERROR is harder to understand when reading the code for the first time, especially without reading zlib documentation.

yhirose:
@Fefer-Ivan, thanks for the detailed explanation. I'll merge your pull request soon. Thanks for your contribution!

Fefer-Ivan:
@yhirose, I am glad I could help.

Also, could you please make a new release tag after merge?

Thank you.

yhirose:
@Fefer-Ivan, v7.0.6 is available now.

PR #644, opened 2020-09-08
Add compression buffer size customization

Fefer-Ivan:
While working on a previous pull request, I noticed following note in zlib documentation https://zlib.net/zlib_how.html:
```
CHUNK is simply the buffer size for feeding data to and pulling data from the zlib routines.
Larger buffer sizes would be more efficient, especially for inflate().
If the memory is available, buffers sizes on the order of 128K or 256K bytes should be used.
```

So I decided to add CPPHTTPLIB_COMPRESSION_BUFSIZ macro in line with CPPHTTPLIB_RECV_BUFSIZ macro to allow for buffer size customization on different systems.

Also, I found that brotli decomressor code is written using C-array instead of std::array (as in other [de]compressors), so I refactored it slightly. 

Hope it is useful.

Thank you.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -2381,10 +2381,10 @@ public:
 
     decoder_r = BROTLI_DECODER_RESULT_NEEDS_MORE_OUTPUT;
 
+    std::array<char, CPPHTTPLIB_COMPRESSION_BUFSIZ> buff;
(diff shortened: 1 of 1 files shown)


yhirose:
@Fefer-Ivan, looks very good to me. Thanks for your great contribution!

Issue #359, opened 2020-02-20
use of undeclared identifier 'getifaddrs'


mrpuzzler:
I try to use this code in my Android NDK project.  And get this error:

httplib.h:1495:3: error: use of undeclared identifier 'getifaddrs'; did you mean 'ifaddrs'?
httplib.h:1502:11: error: use of undeclared identifier 'freeifaddrs'

Possible to replace or fix it somehow?

yhirose:
@mrpuzzler, thanks for the report. `getifaddrs` is correct. It seems like Android NDK is lacking 'ifaddrs.h'. Please see the following stackoverflow article for more detailed.
https://stackoverflow.com/questions/26476346/ifaddrs-h-header-not-found-when-compiling-sdl-for-android

So unfortunately there is nothing to do on the cpp-httplib side. If you don't need to call `set_interface` method, you can simply remove the content in `if2ip` for now.

Hope it helps.

mrpuzzler:
I don't need set_interface method ... possible to have some defines to remove it?

Btw, also, it's great to have some defines to remove server or client part of code. For example in my app I don't need any servel logic at all, so just exclude with define.

PixlRainbow:
you don't need defines to remove server part of code. This library is single-header, which essentially means that the code is injected inline into your calling code during compilation and unused code from the library is simply not included in your binaries. At least, that's what normally happens when you have a compiler that is good at optimization.

yhirose:
@mrpuzzler, you may be interested in this:
https://github.com/NAbdulla1/cpp-httplib-with-ifaddrs-fixed-for-android

PR #648, opened 2020-09-10
Undefined if2ip() also on Android since getifaddrs() does not exist.

sasq64:
Se issue #359


yhirose:
@sasq64, thanks for your contribution!

PR #649, opened 2020-09-10
don't replace plus with space in headers

tmahring:
Hello
I ran into a problem uploading a file to httplib using cordovas FileTransfer plugin.
The plugin uses '+' signs in the boundary part of the Content-Type header field

here a capture:
`Content-Type: multipart/form-data; boundary=+++++`

Those + signs are replaced by spaces during the parsing of the headers, causing MultipartFormDataParser to fail.

I have to admin that i don't know wether this is a standard violation of the Filetransfer plugin or a implementation error in httplib, but the attached commit fixes the problem for me.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -4768,7 +4768,7 @@ inline bool ClientImpl::redirect(const Request &req, Response &res) {
     return false;
   }
 
-  auto location = res.get_header_value("location");
+  auto location = detail::decode_url(res.get_header_value("location"), true);
   if (location.empty()) { return false; }
 
   const static std::regex re(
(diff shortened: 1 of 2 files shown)


tmahring:
The Location field now gets parsed the original way, fixing the test.

yhirose:
@tmahring, thanks a lot. Could you also add a simple unit test case? Then, I'll merge it.

tmahring:
I have added a test which emulates the behavior of the FileTransfer plugin by using a hardcoded boundary. It fails without my first commit.

yhirose:
@tmahring, thanks for the fine contribution! I'll merge it.

Issue #661, opened 2020-09-21
parse error when server has two responses (first response code is 100)


liu-yabin:
when server has two responses(first response code is 100):

![image](https://user-images.githubusercontent.com/25415782/93750567-42584e00-fc2e-11ea-8d32-a6844c8d9bb1.png)

the expected result is:

* ignore 100 response.
* parser second response(403 or 200 etc).

![image](https://user-images.githubusercontent.com/25415782/93751106-20130000-fc2f-11ea-9414-0c1fc76a6606.png)

cpp-httplib parses first response and put the second response in the body:

![image](https://user-images.githubusercontent.com/25415782/93751830-4b4a1f00-fc30-11ea-9c34-244a549012b4.png)

yhirose:
@liu-yabin, thanks for the feedback. Could you send the same request from curl to see if curl can handle the situation?
My understanding is that a server sends a response back with `100 Continue`, **only** when a client sends a request with `Expect: 100-continue` header. Do you think my understanding is not correct? Thanks!

PixlRainbow:
>  for compatibility with RFC 2068, a server MAY send a 100 (Continue) status in response to an HTTP/1.1 PUT or POST request that does not include an Expect request-header field with the "100-continue" expectation. This exception, the purpose of which is to minimise any client processing delays associated with an undeclared wait for 100 (Continue) status, applies only to HTTP/1.1 requests, and not to requests with any other HTTP-version value.

Interestingly, it says that some servers *may* send unsolicited 100 continue, and in any case even when a developer explicitly sets "Expect: 100-continue" header in the httplib client, our client does not currently know that it should ignore a 100 Continue status response

yhirose:
@PixlRainbow, thanks for the information!

PR #662, opened 2020-09-21
quick fix for #661

PixlRainbow:
Note: does not yet implement "true" Expect handling as it will not wait for server to send `100 Continue` or `417 Expectation Failed`
This PR is only intended to fix client corrupting the response object when parsing `100 Continue` responses.


yhirose:
@PixlRainbow, thanks for the pull request! I took a look at it, and made it simpler at cc14855ba07908ad414639160a3fc2eb91519a77.

PR #671, opened 2020-09-28
Fix memory leak due caused due to X509_STORE

omjego:
When X509_STORE objects are associated with a SSL_CTX (ctx_), they will automatically be deleted when the parent SSL_CTX is deleted (in the destructor). However, library lazily sets these in load_certs(), only doing it on a valid HTTP request (after the socket connects inside the call to initialize_ssl()). It causes a memory leak when connecting to an invalid URL, since it can't establish the socket.
The current PR has quick fix for the issue, although not sure if this is the ideal solution. Should we associate the X509_STORE with the ctx_t immediately in set_ca_cert_store() ?

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -839,6 +839,10 @@ public:
   void set_proxy_digest_auth(const char *username, const char *password);
 #endif
 
+#ifdef CPPHTTPLIB_OPENSSL_SUPPORT
+  void enable_server_certificate_verification(bool enabled);
+#endif
+
   void set_logger(Logger logger);
 
 protected:
@@ -858,6 +862,8 @@ protected:
                        bool close_connection);
 
   Error get_last_error() const;
+
+  void copy_settings(const ClientImpl &rhs);
 
   // Error state
   mutable Error error_ = Error::Success;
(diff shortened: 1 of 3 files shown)


yhirose:
@omjego, thanks for the pull request. How did you detect memory leak? Can I reproduce it easily on my Mac? Also I wonder if you can add a unit test to test/test.cc.

omjego:
Hey @yhirose  the memory leak was detected with [address sanitiser](https://github.com/google/sanitizers).  Have added a test for repro and address sanitizer flag to the test Makefile which will be useful for finding such bugs in future.

yhirose:
@omjego, thanks for adding the unit test. I now fully understand the situation. I reviewed the pull request, but it seems to be not enough. Here is an example:

```cpp
{
  httplib::SSLClient cli("www.google.com");

  auto cs1 = X509_STORE_new();
  X509_STORE_load_locations(cs1, "/etc/ssl/certs/ca-certificates.crt", nullptr);
  cli.set_ca_cert_store(cs1);

  auto cs2 = X509_STORE_new();
  X509_STORE_load_locations(cs2, "/etc/ssl/certs/ca-certificates.crt", nullptr);
  cli.set_ca_cert_store(cs2);

  // When the destructor of `cli` gets executed, only `cs2` will be freed...
}
```

So, I am thinking to remove `ca_cert_store_` member from `SSLClient`, and modify `set_ca_cert_store` and `load_certs` accordingly:

```cpp
inline void SSLClient::set_ca_cert_store(X509_STORE *ca_cert_store) {
  if (ca_cert_store) {
    if (ctx_) {
      if (SSL_CTX_get_cert_store(ctx_) != ca_cert_store) {
        // The previous cert store pointer will be freed, and the current one
        // wiil be freed in `~SSLClient()`.
        SSL_CTX_set_cert_store(ctx_, ca_cert_store);
      }
    } else {
      X509_STORE_free(ca_cert_store);
    }
  }
}

inline bool SSLClient::load_certs() {
  bool ret = true;

  std::call_once(initialize_cert_, [&]() {
    std::lock_guard<std::mutex> guard(ctx_mutex_);
    if (!ca_cert_file_path_.empty()) {
      if (!SSL_CTX_load_verify_locations(ctx_, ca_cert_file_path_.c_str(),
                                         nullptr)) {
        ret = false;
      }
    } else if (!ca_cert_dir_path_.empty()) {
      if (!SSL_CTX_load_verify_locations(ctx_, nullptr,
                                         ca_cert_dir_path_.c_str())) {
        ret = false;
      }
    } else {
#ifdef _WIN32
      detail::load_system_certs_on_windows(SSL_CTX_get_cert_store(ctx_));
#else
      SSL_CTX_set_default_verify_paths(ctx_);
#endif
    }
  });

  return ret;
}
```

What do you think?

omjego:
Hey @yhirose, Thanks for the example, I kind of missed that.  This approach  sounds good to me!

yhirose:
@omjego, thanks for the info. It seems like clang does't work with `-fsanitize=address` on my Mac. It gives not memory leak with the `SSLClientTest.UpdateCAStore`...

yhirose:
@omjego, could you also fix the conflicts in your pull request? Thanks!

omjego:
@yhirose oh I'm using linux, seems [address sanitizer isn't supported on mac]( https://github.com/google/sanitizers/issues/1026). Should I remove the sanitizer option from the makefile or make it conditional based on os?

yhirose:
@omjego, I am ok with the current Makefile, since it doesn't cause any side effect. Thanks for your contribution!

PR #674, opened 2020-09-29
Added MACRO for server_certificate_verification

gtnardy:
If I have a request (made with universal interface httplib::Client) which redirects (set_follow_location) from HTTP to HTTPS and I had server_certificate_verification_ set to false, the redirection won't copy my server_certificate_verification_ setting and will make an erroneous request.

This solution allows me to keep server_certificate_verification configuration_ = false for my whole code and surrounds my problem.

Also, setting server_certificate_verification_ in the universal interface won't have effect (before it had been redirected from HTTP to HTTPS) and SSLClient instance had been created.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -87,6 +87,10 @@
                       : 0))
 #endif
 
+#ifndef CPPHTTPLIB_SERVER_CERTIFICATE_VERIFICATION
+#define CPPHTTPLIB_SERVER_CERTIFICATE_VERIFICATION true
+#endif
+
 /*
  * Headers
  */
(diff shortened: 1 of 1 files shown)


yhirose:
@gtnardy, thanks for the pull request. It seems "the redirection won't copy my server_certificate_verification_ setting" is a bug. I am thinking to fix the bug instead of introducing `CPPHTTPLIB_SERVER_CERTIFICATE_VERIFICATION` macro.

yhirose:
@gtnardy, could test with the latest httplib.h? It now copies server_certificate_verification_. Thanks for your help!

gtnardy:
Works like a charm! Thank you!

yhirose:
@gtnardy, thanks for testing it!

PR #688, opened 2020-10-10
Add `cache_control` parameter to `set_mount_point`

agasparovic:
Specifies the Cache-Control header value to return when specified. For example:

```
svr.set_mount_point("/assets", "public/assets", "public, max-age=604800, immutable");
```

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -580,7 +580,7 @@ public:
 
   bool set_base_dir(const char *dir, const char *mount_point = nullptr);
   bool set_mount_point(const char *mount_point, const char *dir,
-                       const char *cache_control = nullptr);
+                       Headers headers = Headers());
   bool remove_mount_point(const char *mount_point);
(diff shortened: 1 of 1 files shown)


yhirose:
@agasparovic, thank you for your contribution. It looks good to me, but I think it would be nice to allow users to specify other HTTP headers as well. Is it possible to make the method like `bool set_mount_point(const char *mount_point, const char *dir, Headers headers)`? Thanks!

agasparovic:
That's a great idea! Just updated.

yhirose:
@agasparovic, thanks for the update. I just found that the Windows build gets broken due to the C++17 structured binding feature. :) Could you adjust the code to compatible with C++11?

agasparovic:
Of course!

yhirose:
@agasparovic, excellent! I'll merge it right away. Thanks for your contribution!

agasparovic:
Thank you for all of your excellent projects!

PR #689, opened 2020-10-10
fix MSVC2015 error:  std::tolower to ::lower

wanggao1990:


yhirose:
@wanggao1990, thanks for your contribution!

PR #690, opened 2020-10-10
Added "Transfer-Encoding:Chunked" support when your content length is unpredictable.

berkanuslu:
When you're working with streams, especially if different threads control this stream, you might not know content-length, but you can transfer your data with Transfer-Encoding: Chunked. It has been implemented based on https://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html documentation. I have created an example for transferring your data with chunked. Also, I have added an example of a download file from the internet with the content receiver and ofstream.

### DIFF
--- a/example/chuncked_client.vcxproj
+++ b/example/chuncked_client.vcxproj
@@ -1,0 +1,160 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project DefaultTargets="Build" ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ItemGroup Label="ProjectConfigurations">
+    <ProjectConfiguration Include="Debug|Win32">
+      <Configuration>Debug</Configuration>
+      <Platform>Win32</Platform>
+    </ProjectConfiguration>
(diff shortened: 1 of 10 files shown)


evg82:
Awesome! I ask for this function some days ago. I will try it tomorrow becose i have all my app code ready to use it.

yhirose:
@berkanuslu, thanks for your pull request. However, I feel the pull request is too big to merge for me. Also I did quick check and I have already found  two problems:

* It breaks KeepAlive unit test
* I removed the recent memory leak fix. (https://github.com/yhirose/cpp-httplib/pull/671)

So I am not planning to merge it anytime soon though, at least I understand that you need chunked transfer encoding on client side in addition to server side (https://github.com/yhirose/cpp-httplib#chunked-transfer-encoding). Anyway, thanks for your contribution!

berkanuslu:
@yhirose, thanks for your reply. I don't know what breaks the KeepAlive unit test, but I'm working on fixing this. I just wanted to share this code blocks with the community. But you're right, and it's a little bit big to merge it. It's up to you. By the way, I have closed the pull request. I will split it into a small piece of pull requests.

PR #692, opened 2020-10-11
Use move semantics instead of copy for functions

agasparovic:
In some cases, a few more copies could be prevented by changing function definitions to accept parameters by const-ref, rather than by value, but I didn't want to change public signatures.


yhirose:
@agasparovic, thanks for the improvement. The code looks good to me though, somehow some unit tests failed. Could you take a look at them?

agasparovic:
Hi @yhirose, all fixed!

yhirose:
@agasparovic, thanks for the contribution!

Issue #691, opened 2020-10-10
Multipart form data boundary


lightvector:
If I understand the current code correctly, the construction of the multipart form data boundary might a little low-entropy.

Firstly: currently the implementation `make_multipart_data_boundary` seems to do this:
```
  std::random_device seed_gen;
  std::mt19937 engine(seed_gen());
```
And then concatenates 96 random bits worth of entropy from `engine` on to `"--cpp-httplib-multipart-data-"`.

However by my [understanding](https://en.cppreference.com/w/cpp/numeric/random/random_device), `seed_gen()` only returns a single unsigned int. So there are only 2^32 = ~4 billion possible initializations of mt19937 that can be performed. Four billion is not actually that large for modern computers - I think this implies that with adversarial input, within a few tens of gigabytes of constructed data, it should be possible to have a nontrivial chance of that data conflicting with the boundary. Depending on an application, perhaps this in some cases this could become some sort of minor security risk?

Secondly: the use of std::random_device might be problematic. On some systems+toolchains it seems like the device [may actually be deterministic](https://stackoverflow.com/questions/18880654/why-do-i-get-the-same-sequence-for-every-run-with-stdrandom-device-with-mingw), which I stumbled upon from [this page](https://www.pcg-random.org/posts/cpps-random_device.html).

Any thoughts what to do here? Using more than one draw from seed_gen() in the initialization of mt_19937 would definitely help the first issue and is probably worth doing regardless? For second issue, I'm not sure what one might do, if std::random_device could on some systems in fact be extremely low-entropy.

Perhaps it could be possible to allow a user to specify their own multipart form data boundary that they know or can generate that will not conflict with the data? It also avoids semi-costly initialization of mt19937 (which is quite a big generator to initialize to only then grab 96 bits from), although maybe this initialization cost/time is small anyways compared to any network latencies and such. For my own personal use, I've tried such a hack for the client which seems to work. Replacing this:
```c++
inline Result ClientImpl::Post(const char *path, const Headers &headers,
                               const MultipartFormDataItems &items) {
    auto boundary = detail::make_multipart_data_boundary();
    ...
```

With this:
```c++
inline Result ClientImpl::Post(const char *path, const Headers &headers,
                               const MultipartFormDataItems &items) {
    return Post(path, headers, items, detail::make_multipart_data_boundary());
}
inline Result ClientImpl::Post(const char *path, const Headers &headers,
                               const MultipartFormDataItems &items,
                               const std::string& boundary) {
[...]

PR #694, opened 2020-10-11
Allow client to specify boundary and use more entropy by default (#691)

lightvector:
An attempt to address some of the things in https://github.com/yhirose/cpp-httplib/issues/691

* Increasing the amount of entropy used to initialize the multipart boundary generation to something more commensurate with the amount used.
* Allowing the client to specify the boundary on their own if they wish, as a workaround to any issues with std::random_device, or for any other reason.

The server side also generates multipart boundaries too, but the generation of them is conditional, and is buried deeper in protected methods, so I did not attempt to do the same on the server side as on the client side in this commit. So perhaps the resolution here is only partial and not complete. It's also hard for me to actually think of a specific way a bad boundary could become any kind of security issue unless other separate things have also gone wrong. But in any case, regardless of this or completeness, I hope both of these changes are still non-objectionable and reasonable on their own.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -697,7 +697,8 @@ enum Error {
   Canceled,
   SSLConnection,
   SSLLoadingCerts,
-  SSLServerVerification
+  SSLServerVerification,
+  UnsupportedMultipartBoundaryChars
 };
 
 class Result {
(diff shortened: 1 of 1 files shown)


lightvector:
Also, there are more characters than simply alphanumeric and "-" and "_" allowed, but it seems like different online sources are a bit conflicted in what they allow.

This is supposed to be official:
https://tools.ietf.org/html/rfc2046#section-5.1.1
But Microsoft's page has a different set of allowed characters:
https://docs.microsoft.com/en-us/exchange/troubleshoot/administration/multipart-mixed-mime-message-format

And in any case, the code in httplib currently doesn't double-quote the boundary, which I believe would be needed to avoid running afoul of the tspecials here:
https://www.w3.org/Protocols/rfc1341/4_Content-Type.html

So for now, alphanumeric plus "-" plus "_" seems like a pretty safe set.

yhirose:
@lightvector, the changes looks very good to me. Thanks for your contribution!

PR #696, opened 2020-10-12
Added Client "Transfer-Encoding:Chunked" support when your content length is unpredictable.

berkanuslu:
When you're working with streams, especially if different threads control this stream, you might not know content-length, but you can transfer your data with Transfer-Encoding: Chunked. It has been implemented based on https://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html documentation. I have created an example for transferring your data with chunked.

### DIFF
--- a/example/chunked_client.vcxproj
+++ b/example/chunked_client.vcxproj
@@ -1,0 +1,161 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project DefaultTargets="Build" ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ItemGroup Label="ProjectConfigurations">
+    <ProjectConfiguration Include="Debug|Win32">
+      <Configuration>Debug</Configuration>
+      <Platform>Win32</Platform>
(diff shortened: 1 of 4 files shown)


yhirose:
@berkanuslu, thanks for the new pull request. It looks much simpler! Could you provide a unit test for this feature?  Then, I'll review the code. Thanks!

evg82:
Hi @berkanuslu, Im working with your code. It works ok but have some bugs:
Content-Length is included but chunked enconding conflict with it.
gzip compresion do not work, at least in my code. The code should be very similar to the server one. It will be great if some code can be reused in server and client.

I dont know if you are working in the code or not. If the PR is enough for your needs i can continue your work.

yhirose:
@berkanuslu, any response?

berkanuslu:
Sorry for the late reply, folks.
@evg82 I'll check this error with gzip support. Maybe something is missing for gzip. 
@yhirose Ok, I'll add a new test code for this feature in a few days.

berkanuslu:
I want to close this pull request and open a new one. We can continue to discuss this with the new one.

Issue #699, opened 2020-10-13
could you reopen #697?


hotbroker:
after reviewing the code,im pretty sure this is an error:

reusing the socket in SSL_shutdown but the socket is closed before SSL_shutdown  in SSLClient::close_socket

yhirose:
@hotbroker, glad to hear that. Could you show the smallest possible code snippet, so that I can easily reproduce it on my Mac? Thanks!

hotbroker:
> @hotbroker, glad to hear that. Could you show the smallest possible code snippet, so that I can easily reproduce it on my Mac? Thanks!

well ,i dont think you can reproduc this ? 
sometimes this will may not cause a big prolem,like close a invalid handle,or send data with a invalid handle.
but also may cause a serious issue,like close a handle not belong to it(the secon time it was close,this handle value may be created by other thread for othe type resource like event,file etc *on windows*)

yhirose:
@hotbroker, I think the problem should be reproducible regardless the operating system that you use, because it's OpenSSL issue. Could you anyway try to make a smallest possible snippet? Then, I'll debug it. Thanks!

hotbroker:
> @hotbroker, I think the problem should be reproducible regardless the operating system that you use, because it's OpenSSL issue. Could you anyway try to make a smallest possible snippet? Then, I'll debug it. Thanks!

`	std::string url = "https://www.apple.com/";
	Url u(url);
	string ep = u.scheme() + "://" + u.host();

	string path = url.substr(ep.length());

	auto res1 = httplib::Client(ep.data());

	auto res = res1.Get(path.data());
`
check this out.
**stack as follow:**

00 vrfcore!VerifierStopMessageEx
01 vfbasics!AVrfpVectoredExceptionHandler
02 ntdll!RtlpCallVectoredHandlers
03 ntdll!RtlDispatchException
04 ntdll!KiUserExceptionDispatcher
05 ntdll!NtQueryObject
06 KERNELBASE!GetHandleInformation
07 WS2_32!DSOCKET::FindIFSSocket
08 WS2_32!send
09 !sock_write
0a !bwrite_conv
0b !bio_write_intern
0c !BIO_write
0d !ssl3_write_pending
0e !do_ssl3_write
0f !ssl3_dispatch_alert
10 !ssl3_send_alert
11 !ssl3_shutdown
12 !SSL_shutdown
13 !httplib::detail::ssl_delete
14 !httplib::SSLClient::close_socket
[...]

Issue #697, opened 2020-10-13
error cant pass verifier on windows


hotbroker:
**here is the stack when issue occurs**

00 ntdll!NtQueryObject
01 KERNELBASE!GetHandleInformation
02 WS2_32!DSOCKET::FindIFSSocket
03 WS2_32!send
04 mymodule!sock_write
05 mymodule!bwrite_conv
06 mymodule!bio_write_intern
07 mymodule!BIO_write
08 mymodule!ssl3_write_pending
09 mymodule!do_ssl3_write
0a mymodule!ssl3_dispatch_alert
0b mymodule!ssl3_send_alert
0c mymodule!ssl3_shutdown
0d mymodule!SSL_shutdown
0e mymodule!httplib::detail::ssl_delete
0f mymodule!httplib::SSLClient::close_socket
10 mymodule!httplib::ClientImpl::stop_core
11 mymodule!httplib::ClientImpl::process_request
12 mymodule!httplib::ClientImpl::handle_request
13 mymodule!<lambda_66ee1219a816680a9178e5bd970ecab4>::operator()
14 mymodule!std::invoke<<lambda_66ee1219a816680a9178e5bd970ecab4> &,httplib::Stream &>
15 mymodule!std::_Invoker_ret<bool,0>::_Call<<lambda_66ee1219a816680a9178e5bd970ecab4> &,httplib::Stream &>
16 mymodule!std::_Func_impl_no_alloc<<lambda_66ee1219a816680a9178e5bd970ecab4>,bool,httplib::Stream &>::_Do_call
17 mymodule!std::_Func_class<bool,httplib::Stream &>::operator()
18 mymodule!httplib::detail::process_client_socket_ssl<std::function<bool __cdecl(httplib::Stream &)> >
19 mymodule!httplib::SSLClient::process_socket
1a mymodule!httplib::ClientImpl::send
1b mymodule!httplib::ClientImpl::send_with_content_provider
1c mymodule!httplib::ClientImpl::Post
1d mymodule!httplib::ClientImpl::Post

I check the out,it seems like the socket was closed but the it was used again when delete ssl resource

`    inline void SSLClient::close_socket(Socket& socket, bool process_socket_ret) {
        detail::close_socket(socket.sock);
        socket_.sock = INVALID_SOCKET;
        if (socket.ssl) {
            detail::ssl_delete(ctx_mutex_, socket.ssl, process_socket_ret);
            socket_.ssl = nullptr;
        }
    }
`

[...]

PR #701, opened 2020-10-16
Fix multiple threading bugs including #699 and #697

lightvector:
In the process of fixing #697 and #699, I found that actually the situation is much more complicated. There were multiple issues with the old code, and simply applying the logical fix for those issues alone resulted in failures of test and other things due to the other bugs.

Many of these issues were a bit difficult to untangle individually, and required a nontrivial rework of the shutdown functions, so this pull request reworks all of the shutdown-related functions for the client to fix those thread-safety issues.

Here is the list of things that were fixed:
* In 0743d78c9bab62120422c56868c91eb67bff62ef it looks like a race condition that causes segfaults was buried under the rug by adding a timeout. This isn't a good fix because the segfault can still happen even with the timeout, the timeout only makes it less common. The race condition is that `close_socket` is *unsafe* to call when there are other threads still using the socket because other code might end up attempting to access `INVALID_SOCKET` believing it is a correct socket. And even if that were patched, theoretically the OS could reassign that socket id again causing it to be a valid but completely different socket corresponding to anything else in the application.
\
The proper fix is to add a `socket_requests_in_flight_` variable that is also maintained under `socket_mutex_` that tracks when a request might be actively using the socket despite holding the mutex. See in the new implementation of `inline void ClientImpl::stop()` - when this variable is > 0, we now refrain from closing the socket and instead *only* shutdown the socket (which *should* be thread-safe) to disable read/write on the socket. We then set another new variable `socket_should_be_closed_when_request_is_done_` so that the thread that is still involved in the active request can know to close the socket when it is done. 
\
So now, if there is any ongoing request, the request thread takes full responsibility for closing the socket, later. That way, there can be no race - the same thread making the requests is the one closing the socket.

* In #699 and #697 it was observed that on some OSes, it can cause an error to attempt to SSL_shutdown on a closed socket. In general, one should never use a closed socket with any SSL functions. 
\
The fix is to refactor the shutdown logic into three separate functions `shutdown_ssl` and `shutdown_socket` and `close_socket` because each of them needs to be called in different situations, and we now we no longer ever call `shutdown_ssl` after calling `close_socket`. 
\
It might be a bit confusing to need three functions here, but I don't know of a good way to do it with fewer than three. The general order of calling them should be `shutdown_ssl` then `shutdown_socket` then `close_socket`, but it is also necessary to call the second one alone (as in the first bullet point above), and `shutdown_ssl` needs to be virtual.

* There was a memory and an SSL resource leak in the `~ClientImpl` destructor, because it used to simply call `ClientImpl::close_socket`, which set `socket_.ssl` to nullptr without freeing it. The tricky subtle cause of this bug is that ~ClientImpl runs, virtual functions no longer call SSLClient's version of those functions! Only SSLClient's version of those functions clean up `socket_.ssl` properly. The proper fix, which is implemented now, is to explicitly make sure `~SSLClient` cleans up the `socket_.ssl` first.

* `error_` was accessed from multiple threads without synchronization. Strictly speaking, this is undefined C++ behavior unless this value is made atomic, so it is now made atomic.

All the tests (mostly) pass, for me, at least, and the code seems to work properly in my own personal application, so I think the state of things is good, but of course the logic is still a bit tricky. Let me know what you think.

### DIFF
--- a/README.md
+++ b/README.md
@@ -609,6 +609,12 @@ cli.set_ca_cert_path("./ca-bundle.crt");
 cli.enable_server_certificate_verification(true);
 ```
 
+Note: When using SSL, it seems impossible to avoid SIGPIPE in all cases, since on some operating systems, SIGPIPE
+can only be suppressed on a per-message basis, but there is no way to make the OpenSSL library do so for its
+internal communications. If your program needs to avoid being terminated on SIGPIPE, the only fully general way might
+be to set up a signal handler for SIGPIPE to handle or ignore it yourself.
+
+
 Compression
 -----------
 
--- a/httplib.h
+++ b/httplib.h
@@ -891,7 +891,21 @@ protected:
   };
 
   virtual bool create_and_connect_socket(Socket &socket);
-  virtual void close_socket(Socket &socket, bool process_socket_ret);
+
+  // All of:
+  //   shutdown_ssl
+  //   shutdown_socket
+  //   close_socket
+  // should ONLY be called when socket_mutex_ is locked.
+  // Also, shutdown_ssl and close_socket should also NOT be called concurrently
+  // with a DIFFERENT thread sending requests using that socket.
+  virtual void shutdown_ssl(Socket &socket, bool shutdown_gracefully);
+  void shutdown_socket(Socket &socket);
+  void close_socket(Socket &socket);
+
+  // Similar to shutdown_ssl and close_socket, this should NOT be called
+  // concurrently with a DIFFERENT thread sending requests from the socket
+  void lock_socket_and_shutdown_and_close();
 
   bool process_request(Stream &strm, const Request &req, Response &res,
                        bool close_connection);
@@ -901,7 +915,7 @@ protected:
   void copy_settings(const ClientImpl &rhs);
 
   // Error state
-  mutable Error error_ = Error::Success;
+  mutable std::atomic<Error> error_;
 
   // Socket endoint information
   const std::string host_;
@@ -912,6 +926,11 @@ protected:
   Socket socket_;
   mutable std::mutex socket_mutex_;
   std::recursive_mutex request_mutex_;
+
+  // These are all protected under socket_mutex
+  int socket_requests_in_flight_ = 0;
+  std::thread::id socket_requests_are_from_thread_ = std::thread::id();
+  bool socket_should_be_closed_when_request_is_done_ = false;
 
   // Default headers
   Headers default_headers_;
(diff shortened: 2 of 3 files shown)


lightvector:
Some followup notes:

To my knowledge, there is STILL a minor bug/race in the code, where `ClientImpl::stop` sets `error_ = Error::Canceled;` but even with `error_` being atomic there is no proper synchronization and no guarantee that this error value will be seen by the thread running the requests. Indeed, if you revert 0cc108d45eeb1fe93f795613ebb964c4099d62f8 and go back to 100 threads, for me a small fraction of the time the `ClientStop` test fails due to this bug. However, the way the code is currently written, fixing this bug is very hard.

Some of the synchronization logic that this pull request has modified is very fragile - so it is no surprise that there were some bugs before. But I think that is very hard to avoid given the current way the code was written - a major refactor would be needed to get the code into a state where thread-safety is easier and more natural, instead of tricky.

Especially, a significant amount of complexity in thread-safety reasoning comes from the fact that `request_mutex_` is recursive. Generally, recursive mutexes are very hard to reason about. It took me 3-4 attempts and multiple hours to make this refactor in a way that I could reason to be thread-safe because of the difficulty of the recursive mutex. As far as I can tell, there isn't a good reason for this mutex to be recursive - the reason it's recursive right now is because `redirect` recursively calls `send` nested from inside of `send`'s own calls. Am I missing another purpose of the recursion?

From a thread-safety and code-correctness perspective, it would be much cleaner and simpler if `process_request` *returned* the information that a redirect was necessary instead of calling `send`, so that `send` itself could loop and try the redirected url, instead of recursing. So the recursion is unnecessarily complicated.

The only other thing I can tell that making the mutex non-recursive would lose is for the ability for the user to call something like Get with a ContentReceiver, and *inside the content receiver function start another query*. In theory this could work on the very last call to the ContentReceiver when the message is complete... but this seems kind of crazy and hopefully no user of this library is doing that.

Feel free to let me know if I'm missing something, otherwise a big TODO item for the future would be to get rid of the recursiveness of the mutex. It would be a much bigger refactor than just this pull request, but would make the thread-safety of the code *far* easier to maintain.

lightvector:
Apparently the inline checks here passed but oss-fuzz got a test fail for this pull request. https://github.com/lightvector/cpp-httplib/runs/1266643298
```
[ RUN      ] ServerTest.SlowPostFail
test.cc:2088: Failure
Value of: !res
  Actual: false
Expected: true
[  FAILED  ] ServerTest.SlowPostFail (2186 ms)
```
But then I reran it and it worked: https://github.com/lightvector/cpp-httplib/actions/runs/311486216

And it's not obvious to me how ServerTest.SlowPostFail is related to this pull request directly, since it doesn't interact much with the socket shutdown/close refactors in this pull request all that much more than the other tests. At least, this test does seem to be nondeterministic somehow?

yhirose:
@lightvector, thank very much for the pull request. Since it's a big change, it will definitely take a while for me to fully absorb all the details. If it handles more than one thing, is it possible to split into pieces? Also as long as you could, could you provide corresponding unit tests as well? Thanks!

lightvector:
I haven't been able to get to this. I've still been using my version successfully for my own project, but if you need me to break this up further for your own ability to review, then it might be a little while.

yhirose:
@lightvector, I got some free time this weekend, and was able to look into the pull request. I now understand what it's doing, and looks very good! I'll merge it soon. I think it will also fix #771 as well.

yhirose:
I have merged it after the merge conflict has manually been fixed at 02d3cd59096cad1139ee4eb4ff2ceb30754d9ec0. Thanks for your great contribution!

lightvector:
@yhirose - cool! Despite not getting around to working further on this myself in this case, I'm glad to have been of help. :)

PR #702, opened 2020-10-16
Fix #700 null pointer exception

### LINKED ISSUE #700: Null exception
This took me many hours to figure it out, maybe I am wrong or something off on my side. So this happen when I post a multipart form data and 1 of the field is a file field. And that file is bigger then the buffer size ( > 4096 ).

The problem is from here : 

`return multipart_form_data_parser.parse(buf, n, std::move(multipart_receiver),
                                              std::move(mulitpart_header));`

If the multipart content is bigger then the 4096 which makes those function called multiple times, for the first call is fine because the `multipart_reciever` is still exist, but for the second and so on it become empty (null, not accessible anymore).

To fix it I remove the `std::move` and change the parse parameter from `T content_callback` to` const T &content_callback`. I dont know if it a solid fixed, you should be better on this fix.

I am using VS Community 2017 for the compiler btw, I didnt check with gcc or clang.

apinprastya:


yhirose:
@apinprastya, thanks for the pull request!

Issue #680, opened 2020-10-06
Timeouts don't apply to SSLClient*


FrozenSource:
Hi,
Loving this library so far, however I seem to have found a problem in which the timeouts do not apply to at least the SSLClient.

```cpp
#define CPPHTTPLIB_OPENSSL_SUPPORT
#define CPPHTTPLIB_CONNECTION_TIMEOUT_SECOND 2
#define CPPHTTPLIB_CONNECTION_TIMEOUT_USECOND 0
#define CPPHTTPLIB_READ_TIMEOUT_SECOND 2
#define CPPHTTPLIB_READ_TIMEOUT_USECOND 0
#include <httplib.h>
```

For the following configuration, the ssl implementation waits for about 30s or so....

```cpp
httplib::SSLClient cli(C_ServerRemote, 443);
cli.enable_server_certificate_verification(false);
cli.set_follow_location(true);
std::string sBuffer;
httplib::Headers headers;
headers.emplace("Accept-Encoding", "gzip, deflate");
headers.emplace("Accept", "text/plain");
auto res = cli.Get("/cli/latest", headers);
```

The following returns in 2 seconds as expected.

```cpp
httplib::Client cli(C_ServerRemote, 80);
cli.set_follow_location(true);
std::tring sBuffer;
httplib::Headers headers;
headers.emplace("Accept-Encoding", "gzip, deflate");
headers.emplace("Accept", "text/plain");
auto res = cli.Get("/cli/latest", headers);
```

Wonder if you know a fix for this?

Sincerely,
FrozenSource

yhirose:
@FrozenSource, could you format the code snippets to make it more readable? Thanks!

yhirose:
@FrozenSource, thank you! (I make another simple formatting change to use the GitHub markdown code highlighting feature.)
I would like to clarify the issue. Which has the problem, `CPPHTTPLIB_CONNECTION_TIMEOUT_SECOND` or `CPPHTTPLIB_READ_TIMEOUT_SECOND`?

FrozenSource:
@yhirose I see so you need three of those '`'. Since you get a read error when the connection could not be made since the remote is offline I take it `CPPHTTPLIB_READ_TIMEOUT_SECOND` is not working in an SSL envrionment.

yhirose:
@FrozenSource, sorry that my question confuses you. I am asking, is it a connection timeout problem, or a reading timeout problem?

FrozenSource:
[...]

PR #711, opened 2020-10-20
[oss-fuzz-26439] Fix suffix-byte-range issue

omjego:
Hey @yhirose 
This is a fix for oss -fuzz [issue 26439](https://oss-fuzz.com/testcase-detail/5390822188777472). 
For the request with range header value `{"Range", "bytes=-9999"}` (suffix length is greater than response size),  the server used to crash cause of invalid response offset generated by program.  We're supposed to return the whole response entity. 
From [HTTP Spec](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.35) 
``` If the entity is shorter than the specified suffix-length, the entire entity-body is used```

Let me know if this looks good to you.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -173,6 +173,7 @@ using socket_t = int;
 #define INVALID_SOCKET (-1)
 #endif //_WIN32
 
+#include <algorithm>
 #include <array>
 #include <atomic>
 #include <cassert>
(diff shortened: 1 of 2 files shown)


yhirose:
@omjego, great!

yhirose:
It seems like this change causes this... https://github.com/yhirose/cpp-httplib/issues/680#issuecomment-717316869
I'll take a look at it.

Issue #708, opened 2020-10-19
SSE with long message


evg82:
When i send long chunk with type text/event-stream to server, content receiver is called multiple times, first one containts "data: ....". i suppose i need to reasamble message but this function should be handled in lib, and call content receiver with a full message.

yhirose:
@evg82, thanks for the feedback. I don't understand what "When i send long chunk with type text/event-stream to server" means. `text/event-stream` messages should be sent **from server to client** on SSE protocol. Could you explain more about the issue? Thanks!

evg82:
Sorry for my poor description. I will try to explain better:
I have a server and a client with your lib.
Server sends for example a SSE packet with 10k payload, i see it correctly in wireshark, Client receive this correctly but gets multiple calls to content_receiver with up to CPPHTTPLIB_RECV_BUFSIZ bytes. For normal chunked encoding this is ok, but for SSE i need to receive entire message OR get message lenght to reasamble it in my code.

I dont know how to get message length of each chunk.

yhirose:
@evg82, please see my comment in #712, and try my solution in #713.

evg82:
Hi, its work ok to me. Thanks!

yhirose:
@evg82, thanks for checking!

PR #712, opened 2020-10-20
Added progress in chunked reader

evg82:
This solve sittuation described in #708.

With this PR progress callback is called at chunk begins and after each read. this allow for example read a 100k SSE block in the  client.

I send my current code to handle big json SSE for example:
```
string   Buffer;
Request  Req;
Response Res;
Req.method           = "POST";
Req.path             = Url->Path() + "/getRemote";
Req.headers          = {{"Cookie", Cookie}, { "Accept", "text/event-stream" }, {"Content-Type", "application/json"}};
Req.body             = Content.dump();
Req.response_handler = [&](const Response &response) -> bool {
  // HANDLE HEADERS
};
Req.progress = [&](uint64_t current, uint64_t total) -> bool {
  if((current == 0) && (total > 0x00100000)) return false;
  if(current == 0) {
    Buffer.clear();
    Buffer.reserve(total);
    printf("Current chunk size: %d\n", (int)total);
  } else if(current == total) {
    if((Buffer.size() > 5) && (Buffer.substr(0, 6) == "data: ")) {
      Buffer = Buffer.substr(6);
      json Content = json::parse(Buffer, NULL, false);
      // DO WHAT YOU NEED WITH SSE EVENT
      Buffer.clear();
  }
  return true;
};
Req.content_receiver = [&](const char *data, uint64_t len) -> bool {
    Buffer.append(data, len);
    return true;
};
HTTPClient->send(Req, Res);
```

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -2511,6 +2511,9 @@ inline bool read_content_with_length(Stream &strm, uint64_t len,
   char buf[CPPHTTPLIB_RECV_BUFSIZ];
 
   uint64_t r = 0;
+  if (progress) {
+    if (!progress(r, len)) { return false; }
+  }  
   while (r < len) {
     auto read_len = static_cast<size_t>(len - r);
     auto n = strm.read(buf, (std::min)(read_len, CPPHTTPLIB_RECV_BUFSIZ));
@@ -2554,7 +2557,7 @@ inline bool read_content_without_length(Stream &strm, ContentReceiver out) {
   return true;
 }
 
-inline bool read_content_chunked(Stream &strm, ContentReceiver out) {
+inline bool read_content_chunked(Stream &strm, Progress progress, ContentReceiver out) {
   const auto bufsiz = 16;
   char buf[bufsiz];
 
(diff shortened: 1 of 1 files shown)


yhirose:
@evg82, thanks for the pull request, and I review it. But `Progress` handler is not designed for that purpose, but for allowing users to implement downloading UI for a **finite** size request. That's why I intentionally don't use `Progress` handler in  `read_coutent_without_length` and `read_content_chunked`.

I now understand what you want to do. So I enhanced `ContentReceiver` to take additional parameters `uint64_t offset` and `uint64_t total_lenth` and committed the change in #713. You can use it like this:

```cpp
http::Request req;
req.content_receiver = [&](const char *data, size_t data_length, uint64_t offset, uint64_t total_length) -> bool {
  ...
  return true;
};
```

When `offset` is set to 0, it indicates that a new chunked message arrives, and you can know the size of it with `total_length`.
Please let me know if it works for you. Thanks!

yhirose:
Implemented in #708.

PR #725, opened 2020-10-28
Added Client "Transfer-Encoding:Chunked" support when your content length is unpredictable.

berkanuslu:
When you're working with streams, especially if different threads control this stream, you might not know content-length, but you can transfer your data with Transfer-Encoding: Chunked. It has been implemented based on https://www.w3.org/Protocols/rfc2616/rfc2616-sec3.html documentation. I have created an example for transferring your data with chunked.

### DIFF
--- a/example/chunked_client.vcxproj
+++ b/example/chunked_client.vcxproj
@@ -1,0 +1,161 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project DefaultTargets="Build" ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <ItemGroup Label="ProjectConfigurations">
+    <ProjectConfiguration Include="Debug|Win32">
+      <Configuration>Debug</Configuration>
+      <Platform>Win32</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Debug|x64">
+      <Configuration>Debug</Configuration>
+      <Platform>x64</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Release|Win32">
+      <Configuration>Release</Configuration>
+      <Platform>Win32</Platform>
+    </ProjectConfiguration>
+    <ProjectConfiguration Include="Release|x64">
+      <Configuration>Release</Configuration>
+      <Platform>x64</Platform>
+    </ProjectConfiguration>
+  </ItemGroup>
+  <PropertyGroup Label="Globals">
+    <ProjectGuid>{8D6B6717-EE73-4EC8-B56D-80E23524C408}</ProjectGuid>
+    <Keyword>Win32Proj</Keyword>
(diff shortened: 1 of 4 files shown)


yhirose:
@berkanuslu, thanks for the new pull request. But it looks same as before...

> Hi @berkanuslu, Im working with your code. It works ok but have some bugs:
Content-Length is included but chunked enconding conflict with it.
gzip compresion do not work, at least in my code. The code should be very similar to the server one. It will be great if some code can be reused in server and client.
>
> I dont know if you are working in the code or not. If the PR is enough for your needs i can continue your work.

berkanuslu:
I didn't encounter any issue about gzip and content length conflict. @evg82, can you share any sample about that? Where did you find this problem? Because I think that gzip is about Content-Encoding, and chunked is about Transfer-Encoding. But please correct me if I'm wrong.

evg82:
Im going to test your last pull, to see if i find any problem with compression.

evg82:
Im testing it and see some problems:

You set in lines around 4960 "Transfer-Encoding: chunked" and "Content-Length: 0". However this headers are exclusive as i can read in https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding

I cant make compression works. In client when i enable compresion and send a post request, for example, and empty request is send. Im working in an example to test this problem.

berkanuslu:
@evg82 you're right. I shouldn't force to set the Content-Length variable.  I have made some changes. I have tried with `cli.compress(true);` options, and there is no error now. Can you check the latest changes? If there is any problem, please let me know.

evg82:
It work correctly on my machine

yhirose:
@berkanuslu, I took a look at the code quickly, but I feel I cannot accept this pull request.

First of all, it doesn't support compression. If `cli.set_compress(true);` is enabled, then your unit test fails...

The chunked-encoding shouldn't be handled in `ClientImpl::write_request`, but `ClientImpl::send_with_content_provider` where we can handle both chunked-encoding and compression properly.

On server side, I have already implemented chunked-encoding with compression in `Server::write_content_with_provider` and `write_content_chunked`, and I think we can use the code as much as possible for the client implementation. In this way, we can avoid code duplication and handle chunked-encoding with compression properly.

As for content provider interface, `ContentProviderWithoutLength` needs to be used instead of `ContentProvider`, because `length` doesn't mean anything for the chunked-encoding. You can see how it is used on the server side.
https://github.com/yhirose/cpp-httplib/blob/9c7d841b37b69cde4ffb7c6d5c1aae6d5d6e1d97/httplib.h#L334-L335

Sorry for bringing up the above issues. But I really appreciate you pull request, because it gives me a motivation to support this feature.

Thanks!

berkanuslu:
@yhirose Anyway, it's up to you. If you want to add this feature to the client implementation, I'll be happy with that.

PR #728, opened 2020-11-01
SSL_Connect in non-blocking mode

### LINKED ISSUE #682: Client connection can hang in SSL_do_handshake
I'm trying to improve connection handling in my client application, and recently play a lot with connecting/disconnecting my network adapter. Since I do it manually during the application run, I may disconnect the network at different points of the code.
It happened to me a few times that the code hangs up indefinitely in `SSL_do_handshake` function. I'm not a specialist in openSSL but from some Google reading I understand that it can be related to (non)blocking I/O settings.
Unfortunately, I can't reproduce this situation very easy - it happened to me only once or twice. I also work with system libssl.so, so it doesn't have debug info.
Does my description make sense and you can give an advice about possible changes? Or I need to debug a bit more?

Thanks - and sorry for a possibly unclear description :-)

miketsts:
This PR fixes #682 

SSL connection is performed in two steps:
First, a regular socket connection is established.
Then, SSL_connect is called to establish SSL handshake with the server.

If a network problem occurs during the second stage, SSL_connect may
hang indefinitely.

The non-blocking mode solves this problem.
The same non-blocking handling has been implemented also for server side SSL_accept call.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -5627,7 +5627,9 @@ inline SSL *ssl_new(socket_t sock, SSL_CTX *ctx, std::mutex &ctx_mutex,
   }
 
   if (ssl) {
+    set_nonblocking(sock, true);
     auto bio = BIO_new_socket(static_cast<int>(sock), BIO_NOCLOSE);
+    BIO_set_nbio(bio, 1);
     SSL_set_bio(ssl, bio, bio);
 
     if (!setup(ssl) || SSL_connect_or_accept(ssl) != 1) {
@@ -5636,8 +5638,11 @@ inline SSL *ssl_new(socket_t sock, SSL_CTX *ctx, std::mutex &ctx_mutex,
         std::lock_guard<std::mutex> guard(ctx_mutex);
         SSL_free(ssl);
       }
+      set_nonblocking(sock, false);
       return nullptr;
     }
+    BIO_set_nbio(bio, 0);
+    set_nonblocking(sock, false);
   }
 
   return ssl;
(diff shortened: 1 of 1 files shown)


yhirose:
@miketsts, thank you very much for the excellent pull request! Basically, the pull request looks very good, but one thing. This library allows users to use `poll` instead of `select` because `select` doesn't perform very well in some environments.

So could you make adjustments to the code where `select` is used? You may want to use `select_read` and `select_write` which already supports both `poll` and `select`?

Also could you fully test with two modes with `CPPHTTPLIB_USE_POLL` and non `CPPHTTPLIB_USE_POLL`, since I can't reproduce this problem even manually with lldb?

Thanks!

miketsts:
OK, I will do it.

yhirose:
@miketsts, the new commit looks good to me. Please let me know when you confirm that both modes `poll` and `select` work with the change. I'll merge it, and release the new version v0.7.13. Thanks!

miketsts:
Hi @yhirose 
Yes, I manually tested both select and test mode for connection timeout during SSL_connect - all is working.
I still think about how we can cause the disconnection or delay in server response of SSL handshake in an programmatic way. If we could create a unit test, I would feel much safer with the fix.
But since all current tests are passing - then we can say that at least the current state did not become worse, right?

yhirose:
> If we could create a unit test, I would feel much safer with the fix.
I agree. Please submit another pull request when you come up with the way to manually reproduce it in a unit test.

> But since all current tests are passing - then we can say that at least the current state did not become worse, right?
I think so too. I'll merge it to the master and wait a while to make sure that nothing is reported due to the change. Then, I'll release it as v0.7.13.

Issue #718, opened 2020-10-24
Issues from oss-fuzz


yhirose:
@omjego, could you take a look at the following report for me? (I tried to see the 'Detailed Report: https://oss-fuzz.com/testcase?key=6234711166550016', but it seems I don't have access right to see it...)

https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=26598

Thanks a lot!

yhirose:
@omjego, we got another one. https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=26632. Thanks for your help!

omjego:
@yhirose  Thank you for creating the issue, I'll take look at them. 
Although I was wondering why can't you access the oss-fuzz dashboard, are you using the email id specified [here](https://github.com/google/oss-fuzz/blob/ee423bf3d6195b38fcac59a3b68336bab6dc7001/projects/cpp-httplib/project.yaml#L3) ?

yhirose:
@yhirose, yes. I can of course read emails from oss-fuzz, but when I click the link to the detailed page, nothing happens. So I don't get any practical information for debugging.

yhirose:
@omjego, I now feed it's a bit too much for me to manage the results from oss-fuzz even though the server is great. If you break down results from oss-fuzz and explain the issues easier to me, it will save lots of time and I can focus on fixing the bugs. Thanks for your consideration!

yhirose:
Fixed remaining problems at 041122908ccced5982b48ce209e5a0e1e47a1f84.

PR #729, opened 2020-11-01
Fix issues reported by oss-fuzz

omjego:
Hey @yhirose , 
This PR fixes 2/3 issues reported by oss-fuzz. https://github.com/yhirose/cpp-httplib/issues/718

**Issue [#26529](https://oss-fuzz.com/testcase-detail/5721235667025920).**   and **Issue [#26632](https://oss-fuzz.com/testcase-detail/4912460286656512).**  , both are caused by same issue

The last statement in get_range_offset_and_length() adds 1 to `ssize_t` result, if the value of the expression `r.second - r.first` is exactly equals LLONG_MAX then adding 1 causes integer overflow. 
I’ve typecasted the result of `r.second - r.first` to ssize_t 

Have added fix and corresponding test for this. 

**Issue [#26598](https://oss-fuzz.com/testcase-detail/6234711166550016).** Status: **Added test**

Issue arises when  the offset in range header is greater than content length e.g if range is `bytes=100-200` and content length = 6, in this case the server tries access illegal memory. [code](https://github.com/yhirose/cpp-httplib/blob/17428a8fbfe5bf08ab0b9cd50ad8574d02bd9b29/httplib.h#L4100), results in crash of server. 

From the code it seems that the [range validations](https://github.com/yhirose/cpp-httplib/blob/17428a8fbfe5bf08ab0b9cd50ad8574d02bd9b29/httplib.h#L2925) are only done while parsing the range headers from the request, and aren't dependent on size of the actual response. As the code for range handling is bit complicated, I've just added tests and commented them  (so you can merge and fix it whenever you get time)  

From [HTTP spec:](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.35)
```If a syntactically valid byte-range-set includes at least one byte- range-spec whose first-byte-pos is less than the current length of the entity-body, or at least one suffix-byte-range-spec with a non- zero suffix-length, then the byte-range-set is satisfiable. Otherwise, the byte-range-set is unsatisfiable. If the byte-range-set is unsatisfiable, the server SHOULD return a response with a status of 416 (Requested range not satisfiable)```

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -3171,7 +3171,7 @@ get_range_offset_and_length(const Request &req, size_t content_length,
     r.second = slen - 1;
   }
 
-  if (r.second == -1 || r.second >= slen) {
+  if (r.second == -1) {
     r.second = slen - 1;
   }
   return std::make_pair(r.first, static_cast<size_t>(r.second - r.first) + 1);
--- a/httplib.h
+++ b/httplib.h
@@ -3171,8 +3171,9 @@ get_range_offset_and_length(const Request &req, size_t content_length,
     r.second = slen - 1;
   }
 
-  if (r.second == -1) { r.second = slen - 1; }
-
+  if (r.second == -1 || r.second >= slen) {
+    r.second = slen - 1;
+  }
   return std::make_pair(r.first, r.second - r.first + 1);
 }
 
--- a/httplib.h
+++ b/httplib.h
@@ -3174,7 +3174,7 @@ get_range_offset_and_length(const Request &req, size_t content_length,
   if (r.second == -1 || r.second >= slen) {
     r.second = slen - 1;
   }
-  return std::make_pair(r.first, r.second - r.first + 1);
+  return std::make_pair(r.first, static_cast<size_t>(r.second - r.first) + 1);
 }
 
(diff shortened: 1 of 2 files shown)


yhirose:
@omjego, thank you for the fixes! I noticed that 2 unit test cases failed. (`ServerTest.GetStreamedWithOffsetGreaterThanContent` and `ServerTest.GetStreamedWithRangeMultipart`)
Could you take care of them?

yhirose:
@omjego, any comments?

omjego:
Hey @yhirose , sorry was trying to debug some test cases. Give me a moment, modifying the original comment.

yhirose:
@omjego, what are the purposes of those commented unit tests? If they are valid tests, there should be something wrong in httplib.h, right?

yhirose:
> From the code it seems that the range validations are only done while parsing the range headers from the request, and aren't dependent on size of the actual response. As the code for range handling is bit complicated, I've just added tests and commented them (so you can merge and fix it whenever you get time)

@omjego, thanks for the above comment. I'll take a look at the uncommented test cases.

PR #732, opened 2020-11-02
ssl-verify-host: fix verifying ip addresses containing zero's

ottigeda:
If the subject alternate name contained an ip address with an zero
(like 10.42.0.1) it could not successfully verify.
It is because in c++ strings are null-terminated
and therefore strlen(name) would return a wrong result.
As I can not see why we can not trust the length returned by openssl,
lets drop this check.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -6225,17 +6225,15 @@ SSLClient::verify_host_with_subject_alt_name(X509 *server_cert) const {
         auto name = (const char *)ASN1_STRING_get0_data(val->d.ia5);
         auto name_len = (size_t)ASN1_STRING_length(val->d.ia5);
 
-        if (strlen(name) == name_len) {
-          switch (type) {
-          case GEN_DNS: dsn_matched = check_host_name(name, name_len); break;
+        switch (type) {
+        case GEN_DNS: dsn_matched = check_host_name(name, name_len); break;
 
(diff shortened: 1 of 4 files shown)


ottigeda:
Sorry, but I do not understand why this check did fail ?
It does run just fine, if I try to reproduce this.

yhirose:
@ottigeda, thanks for the pull request. The failure is not your fault. If it possible to add a unit test for this? At least, I would like to see such a situation. Thanks!

ottigeda:
I did come up with something, luckily "127.0.0.1" has zeroes in it.
Did double check that this test actually fails if we revert the fix for it.

ottigeda:
It seems I have broken the macOS build :-(
Would it be possible for you to help me a bit with this "openssl" commands ?
I am not very good at that, and I have no mac to try with.

yhirose:
@ottigeda, I don't exactly why that happens. At least, the error log says `unknown option -addext`. It seems like the version of openssl 1.1 used in GItHub Actions is a bit old, and you can use the option...

You can get more information from the following post.
https://github.com/libressl-portable/portable/issues/544

Hope it helps!

yhirose:
@ottigeda, looks very good to me. Thanks for your contribution!

PR #734, opened 2020-11-04
[oss-fuzz] Enable msan for fuzz tests

omjego:
Currently there's fuzzing is not done with msan : [https://github.com/google/oss-fuzz/blob/master/projects/cpp-httplib/project.yaml] Enabling msan will help us detect uninitialised memory reads done  by server. 
 [Refer](https://github.com/google/sanitizers/wiki/MemorySanitizer)

When the fuzz test written for server  are run with msan enabled it crashes cause of issue with openssl version 1.1.1 when used with sanitizers.  So removing openssl support from fuzz targets to enabling msan.  
The PR changes Makefile used for fuzzing placed in  `test/fuzzing`

### DIFF
--- a/test/fuzzing/Makefile
+++ b/test/fuzzing/Makefile
@@ -19,7 +19,8 @@ all : server_fuzzer
 
 # Fuzz target, so that you can choose which $(LIB_FUZZING_ENGINE) to use.
 server_fuzzer : server_fuzzer.cc ../../httplib.h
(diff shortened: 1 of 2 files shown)


omjego:
@yhirose  FYI, the `ClientStop` test is flaky, I frequently got segmentation faults while running it on my machine (Linux)

yhirose:
@omjego, thanks for the pull request. Could you explain what you found in `ClientStop` more?

PR #740, opened 2020-11-07
Add unit test SSLConnectTimeout

miketsts:
Add unit test for issue #682 fixed in PR#728, which does not contain
the test of its own.

The test creates a fake SSL server which does not create an SSL context.
When an SSL client attempts to send it a request, it gets a timeout
error. Prior to PR#682, the client would wait indefinitely.


miketsts:
@yhirose I'm very sorry. I seem to make a mistake, and GitHub have included the old commit to the new PR.

I will close this PR and resubmit the clean one.
Sorry again.

PR #741, opened 2020-11-07
Add unit test SSLConnectTimeout

miketsts:
Add unit test for issue #682 fixed in PR #728, which does not contain
the test of its own.

The test creates a fake SSL server, inherited from SSLServer, which
does not create an SSL context. When an SSL client attempts to send it
a request, it gets a timeout error. Prior to PR #728, the client would
wait indefinitely


yhirose:
@miketsts, thanks for your contribution! I'll merge it soon.

Issue #746, opened 2020-11-10
Build error on windows using Visual Studio 2015


vawen:
The file httplib.h doesn't compile using Visual Studio 2015 (line 2982)

Error	C2888	'bool MultipartFormDataParser::parse(const char *,std::size_t,const T &,const U &)': el símbolo no se puede definir en el espacio de nombres 'detail'	mmc_proxy	D:\git\VaxAlprDetector\mmc_proxy\cpp-httplib\httplib.h

Looks like Visual Studio 2015 doesn't like the template definition. Changing the parse definition to:
`  bool parse(const char *buf, size_t n, const ContentReceiver &content_callback,
             const MultipartContentHeader &header_callback) {
`
made it work.

yhirose:
@vawen, thanks for the feedback. I no longer use Visual Studio 2015 any more at my work place. So could you send a pull request including the fix? Thanks!

PR #747, opened 2020-11-10
Fix: #746 Compile error under VS2015

vawen:


yhirose:
@vawen, thanks for your contribution!

PR #761, opened 2020-11-19
Unsafe std::tolower

372046933:
> std::tolower is undefined if the argument's value is neither representable as unsigned char nor equal to EOF

from https://en.cppreference.com/w/cpp/string/byte/tolower


yhirose:
@372046933, great catch!

PR #763, opened 2020-11-19
Fix incorrect content_encoding for Brotli

Unkorunk:


yhirose:
@Unkorunk, thanks for your pull request, but I don't think it's correct.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding

Unkorunk:
Brotli uses abbreviation br in Content-Encoding ([source](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding)). Example: `Content-Encoding: br`

yhirose:
@Unkorunk, I am sorry that my code was incorrect... Thanks for catching the mistake!

PR #788, opened 2020-12-09
Special function to encode query params

Yrds:
Hi, first i like to thank you for this project, it's the easiest/simplest library that i found to make my softwares.

So, i am developing a wrapper/frontend for a bittorrent tracker and for now i am trying to proxying http requests between bittorrent client and the tracker. The problem i am facing it's with query params encoding which its not being recognized by the tracker.

I am using cpp-httplib 0.7.15 from conan repository

This is the code:
```
#include "httplib.h"
#include <iostream>
#include <string>

int main(){
  httplib::Server svr;

  svr.Get("/announce", [](const httplib::Request &req, httplib::Response &res) {
      httplib::Client cli("http://0.0.0.0:6869");

      std::string url = "/announce?" + httplib::detail::params_to_query_str(req.params);

      std::cout << "url: " << url << std::endl;

      auto response = cli.Get(url.c_str());

      res.set_content(response->body, "text/plain");

      std::cout << "response from tracker: " << response->body << std::endl;

   });

  svr.listen("0.0.0.0", 6969);

  return 0;
}
```

Parsing params to string its returning the following url:

```
url: /announce?compact=1&corrupt=0&downloaded=0&event=stopped&info_hash=%92%9B%F9y%ED/%C7%CF#}%CCK%F4%BApN%BF%B1&key=D9BFFAD8&left=0&no_peer_id=1&numwant=0&peer_id=-qB4310-(h8pYKLmnJ30&port=36847&`redundant=0&supportcrypto=1&uploaded=0
response from tracker: <title>Invalid Request</title>
```

Seems like characters like `}`, `#`, `/` are not being encoded correctly since are reserved, which makes total sense `encode_url` function not encoding them, but for query params values it's not working correctly.

I made a new function to encode query params values correctly and it's working now:

```
url: /announce?compact=1&corrupt=0&downloaded=0&event=started&info_hash=%92%9B%F9y%17%ED%2F%C7%CF%23%7D%CCK%F4%BA%05pN%BF%B1&key=D9BFFAD8&left=0&no_peer_id=1&numwant=200&peer_id=-qB4310-%28h8pYKLmnJ30&port=36847&redundant=0&supportcrypto=1&uploaded=0`
response from tracker: d8:completei1e10:downloadedi0e10:incompletei0e8:intervali1853e12:min intervali926e5:peers6:��e
```

Maybe this its a special case(since i can't find no one pointing this) so please analyze if this is pertinent to that part of the code

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,22 +120,48 @@ svr.set_file_extension_and_mimetype_mapping("hh", "text/x-h");
 
 The followings are built-in mappings:
 
-| Extension |     MIME Type          |
-| :-------- | :--------------------- |
-| txt       | text/plain             |
-| html, htm | text/html              |
-| css       | text/css               |
-| jpeg, jpg | image/jpg              |
-| png       | image/png              |
-| gif       | image/gif              |
-| svg       | image/svg+xml          |
-| ico       | image/x-icon           |
-| json      | application/json       |
-| pdf       | application/pdf        |
-| js        | application/javascript |
-| wasm      | application/wasm       |
-| xml       | application/xml        |
-| xhtml     | application/xhtml+xml  |
+| Extension  |          MIME Type          |
+| :--------- | :-------------------------- |
+| css        | text/css                    |
+| csv        | text/csv                    |
+| txt        | text/plain                  |
+| vtt        | text/vtt                    |
+| html, htm  | text/html                   |
+| apng       | image/apng                  |
+| avif       | image/avif                  |
+| bmp        | image/bmp                   |
+| gif        | image/gif                   |
+| png        | image/png                   |
+| svg        | image/svg+xml               |
+| webp       | image/webp                  |
+| ico        | image/x-icon                |
+| tif        | image/tiff                  |
+| tiff       | image/tiff                  |
+| jpeg, jpg  | image/jpeg                  |
+| mp4        | video/mp4                   |
+| mpeg       | video/mpeg                  |
+| webm       | video/webm                  |
+| mp3        | audio/mp3                   |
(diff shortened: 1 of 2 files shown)


yhirose:
@Yrds, thanks for the feedback. Yes, it's a bug... In order to fix it completely, I think `encode_query_param` should do what `encodeURIComponent` JavaScript API. What do you think?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent
https://stackoverflow.com/questions/597028/how-to-encode-a-query-string-so-that-it-is-the-value-of-another-query-string-in#answer-597044

Yrds:
> @Yrds, thanks for the feedback. Yes, it's a bug... In order to fix it completely, I think `encode_query_param` should do what `encodeURIComponent` JavaScript API. What do you think?
> 
> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent
> https://stackoverflow.com/questions/597028/how-to-encode-a-query-string-so-that-it-is-the-value-of-another-query-string-in#answer-597044

Yeah, according to the [specification](https://tc39.es/ecma262/#sec-encodeuricomponent-uricomponent) what is needed to be done it's to add ` !  * ' ( )` to the unescaped characters.

Let me know if something more needed to be done.

yhirose:
@Yrds, could you add all the unescaped characters described in the MDN document for `encodeURIComponent`?

```
var set2 = "-_.!~*'()";   // Unescaped Characters
```

Also could you add one or two unit tests for `encode_query_param `?

Thanks for your help!

PR #794, opened 2020-12-15
Handling Integer overflow in client code.

ankitgoogle:
Response error code doesn't support long values. 
Adding support for longs.

Returning 500 in case custom error code is greater than long.

@yhirose How can I write test case for this?

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -4965,7 +4965,7 @@ inline bool ClientImpl::read_response_line(Stream &strm, Response &res) {
   }
   catch (const std::exception &ex) {
     res.status = 500;
-    res.set_header("EXCEPTION_WHAT", "exception detail unavailable in google3");
+    res.set_header("EXCEPTION_WHAT", ex.what());
   } catch (...) {
     res.status = 500;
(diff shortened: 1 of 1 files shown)


yhirose:
@ankitgoogle, thanks for the pull request. But, I don't understand what "Response error code doesn't support long values" means, because the section 6 in RFC 7231 clearly states it should be **three-digit** code between 000 and 999 as below:

> 6.  Response Status Codes
> 
>    The status-code element is a three-digit integer code giving the
>    result of the attempt to understand and satisfy the request.

So we can keep using `std::stoi` instead of `std::stol` safely (even with 16-bit processors).

Since the current code incorrectly accepts a status code which has 4 or more digits, I'll make the following change anyway. 

```cpp
  const static std::regex re("(HTTP/1\\.[01]) (\\d{3}) (.*?)\r\n"); // If the status code is not a three-digit code, this RegEx doesn't match
```

Please let me know if I misunderstand what your pull request tries to do. Thanks!

yhirose:
Please see 7c1c952f5abcfb0a442d7254938504530cb49fab.

ankitgoogle:
Thank you!

PR #799, opened 2020-12-17
Extend built-in extension MIME mapping

00ff0000red:
The extension ".mjs" has been a common file extension used for JavaScript files for roughly the last six years.
It popped up after EcmaScript 2015 made a breaking change by added modules to the languages.

This PR doesn't add any *actual* code, and literally changes less than a line of text.

If you would rather that users set this as a custom header on their own, please feel free to simply reject this PR.

Related, but if I may ask, what should be a default, and what should be a built-in?
Why is, say, `.mp4` a built-in, but `.mp3`, not?

I could update the built-in list to contain around ~50 of the most common extensions that someone might use.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,22 +120,48 @@ svr.set_file_extension_and_mimetype_mapping("hh", "text/x-h");
 
 The followings are built-in mappings:
 
-| Extension |     MIME Type          |
-| :-------- | :--------------------- |
-| txt       | text/plain             |
-| html, htm | text/html              |
-| css       | text/css               |
-| jpeg, jpg | image/jpg              |
-| png       | image/png              |
-| gif       | image/gif              |
-| svg       | image/svg+xml          |
-| ico       | image/x-icon           |
-| json      | application/json       |
-| pdf       | application/pdf        |
-| js, mjs   | application/javascript |
-| wasm      | application/wasm       |
-| xml       | application/xml        |
-| xhtml     | application/xhtml+xml  |
+| Extension  |          MIME Type          |
+| :--------- | :-------------------------- |
+| css        | text/css                    |
+| csv        | text/csv                    |
+| txt        | text/plain                  |
+| vtt        | text/vtt                    |
+| html, htm  | text/html                   |
+| apng       | image/apng                  |
+| avif       | image/avif                  |
+| bmp        | image/bmp                   |
+| gif        | image/gif                   |
+| png        | image/png                   |
+| svg        | image/svg+xml               |
+| webp       | image/webp                  |
+| ico        | image/x-icon                |
+| tif        | image/tiff                  |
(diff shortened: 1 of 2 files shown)


yhirose:
@00ff0000red, thanks for the pull request. When it comes to the question about the build-in extensions, I was simply interested in those I am using in my projects. So I appreciate it if you could add more common extensions to the list. :)

00ff0000red:
Thank you for the quick response!

> When it comes to the question about the build-in extensions, I was simply interested in those I am using in my projects.

Ha, someone who uses XHTML and Wasm, we have something in common.

> So I appreciate it if you could add more common extensions to the list. :)

I could update it to contain about 30 or so common extensions that I have used myself, and seen used across that web.
Ex: .gif, .png, .mp3, 

But the only potential problem would be the long if/else chain.

For something like this, I probably would be using a `std::unordered_map<std::string_view, std::string_view>`, but since this is a C++11 header, I'd probably drop the views to `std::string`s.

If you don't have a problem with this, or any particular concerns about performance, memory consumption, etc, I'll try to refractor the branching to use a statically allocated map of built-in types in its place.

yhirose:
@00ff0000red, I made a change to use used-defined literals to the file extension lookup logic. It allows us to add more and more file extensions to the list without sacrificing performance. Could you add additional common items based on the current code? Thanks!

yhirose:
@00ff0000red, I noticed some build errors. Could you use the latest httplib.h? Sorry for the confusion.

00ff0000red:
> I made a change to use user-defined literals to the file extension lookup logic. It allows us to add more and more file extensions to the list without sacrificing performance. Could you add additional common items based on the current code? Thanks!

Ah, that is much better than my naive approach; sure thing, I can work with that.

> I noticed some build errors. Could you use the latest httplib.h? Sorry for the confusion.

Yes, I had noticed that the build was failing too, and was about to ask about it, but you've already handled it,

00ff0000red:
So, there was only one thing that threw me off when I had first checked this place out, coincidentally, it relates to MIMEs: the server defaults to sending "js" as "application/javascript", yet I have heard that only "text/javascript" should be used.

While looking around, cherry-picking common files and their extensions, I happened across [an MDN page](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types) saying the same (I probably heard it from there first).
Should the default be changed? Do you think that it would break anyone's servers that depend upon this?

00ff0000red:
Btw: seems like what you just did for the built-in extensions could be useful for the `can_compress_content_type` function, but then again, it might just be micro-optimization.

yhirose:
@00ff0000red, looks fantastic. I'll merge it soon. Thanks for your great contribution!

PR #804, opened 2020-12-18
Ensure socket is closed after processing in SSLServer

Ramjii:
SSLServer doesn't release sockets after processing if SSL connection was properly created, which leads to FD exhaustion.


yhirose:
@Ramjii, great catch! I'll merge it soon. Thank you for your fine contribution!

PR #806, opened 2020-12-19
Fix spelling in readme

00ff0000red:
I had initially come to replace "applie" with "apply," but then I just ran it through Grammarly and fixed everything myself.

Still not sure about those capitalized "Content"s though.


yhirose:
@00ff0000red, looks very good to me! Thanks for improving my poor English writing. :)

PR #811, opened 2020-12-25
Fix readme set_chunked_content_provider

00ff0000red:
`Response::set_chunked_content_provider` has two formal parameters, not one.


yhirose:
@00ff0000red, thanks for your contribution!

PR #815, opened 2020-12-29
Added support of older windows versions before Windows 8

iko1:
`std::run_once` can't be used in older Windows versions that `kernel32.dll` doesn't implement the function `InitOnceBeginInitalize`
`inet_pton' is introduced from Windows 8 for older Windows versions the function `WSAStringToAddressA` should be used

After these changes, binary that used the library can be used on Windows XP, 7 and Vista

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -4,7 +4,6 @@
 //  Copyright (c) 2020 Yuji Hirose. All rights reserved.
 //  MIT License
 //
-
 #ifndef CPPHTTPLIB_HTTPLIB_H
 #define CPPHTTPLIB_HTTPLIB_H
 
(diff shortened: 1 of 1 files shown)


yhirose:
@iko1, thank you for the pull request, but I am not going to accept it, since I took out support for Windows 8 or lower anyway. Please see [Note section](https://github.com/yhirose/cpp-httplib#note) in README.

By the way, I think this is not the matter of Windows 8, but the version of compiler that you are using. For instance, according to the Microsoft document, `InitOnceBeginInitialize` is supported from Windows Vista and Windows Server 2008. Also `inet_pton` is also supported from Windows Vista and Windows Server 2008.

https://docs.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-initoncebegininitialize
https://docs.microsoft.com/en-us/windows/win32/api/ws2tcpip/nf-ws2tcpip-inet_pton
https://stackoverflow.com/questions/43033009/does-stdcall-once-work-under-vs2013-and-below

Hope it helps!

PR #816, opened 2020-12-29
Middleware support

yhirose:


httplib.h
```diff
@@ -4240,7 +4253,7 @@ inline bool Server::write_response_core(Stream &strm, bool close_connection,
   std::string boundary;
   if (need_apply_ranges) { apply_ranges(req, res, content_type, boundary); }
 
-  // Headers
+  // Preapre additional headers
```
00ff0000red, line 4256: Did you mean "prepare"?

yhirose: Yes, that's typo.

Issue #817, opened 2020-12-30
Expression: c >= -1 && c <= 255


Fuatnow:
```cpp
wstring str = LR"(中国語)";
std::wstring_convert<std::codecvt_utf8_utf16<wchar_t>> convert;
string xx = convert.to_bytes(str);
httplib::Params params{
	  { "wxid", xx },
	  { "msg", xx }
};
httplib::Client cli("http://localhost:8900");
auto res = cli.Post("/receiveMsg", params);
```

I'm a chinese. When i use chinese char ， this error has appeared。

I found this mistake in `encode_query_param` function.

I have replaced `for (char const &c : value) `  to `for (unsigned char const &c : value) ` . Now，it works fun.

yhirose:
@Fuatnow, thank you for the report.
@Yrds, it seems like this bug is introduced by #801. Could you take a look at it? Thanks!

Yrds:
Hi @Fuatnow i did some tests, including your example code and i can't reproduce your error, i am tring to figure out what is causing this.

Which line is causing this error?

Also code already have static_cast to unsigned char(when the encode is made) in this lines:

```cpp
      escaped << '%' << std::setw(2)
              << static_cast<int>(static_cast<unsigned char>(c));
```

Tests that i made and all passed:

```cpp
TEST(EncodeQueryParamTest, TestUTF8Characters){
  string chineseCharacters = "中国語";
  string russianCharacters = "дом";
  string brazilianCharacters = "óculos";

  EXPECT_EQ(detail::encode_query_param(chineseCharacters),
      "%E4%B8%AD%E5%9B%BD%E8%AA%9E");

  EXPECT_EQ(detail::encode_query_param(russianCharacters),
      "%D0%B4%D0%BE%D0%BC");

  EXPECT_EQ(detail::encode_query_param(brazilianCharacters),
      "%C3%B3culos");
}
```

Fuatnow:
> Hi @Fuatnow i did some tests, including your example code and i can't reproduce your error, i am tring to figure out what is causing this.
> 
> Which line is causing this error?
> 
> Also code already have static_cast to unsigned char(when the encode is made) in this lines:
> 
> ```c++
>       escaped << '%' << std::setw(2)
>               << static_cast<int>(static_cast<unsigned char>(c));
> ```
> 
> Tests that i made and all passed:
> 
> ```c++
> TEST(EncodeQueryParamTest, TestUTF8Characters){
[...]

PR #818, opened 2020-12-31
Added encode_uri_param tests with UTF-8 characters

Yrds:
Due to #817 issue i wrote more tests to encode_uri_param to ensure other UTF8 characters are encoded correctly.


yhirose:
@Yrds, thanks for the additional tests!

PR #823, opened 2021-01-05
No content check

ecleeEEL:


test/test.cc
```diff
@@ -987,6 +1012,7 @@ TEST(RoutingHandlerTest, PreRoutingHandler) {
   {
     Client cli(HOST, PORT);
 
+
```
yhirose, line 1015: Is this space something intentional?

ecleeEEL: Nope, got borked with the merge conflict

PR #836, opened 2021-01-22
Return Server& from handler setters

00ff0000red:
Server instance methods, `set_error_handler`, `set_pre_routing_handler`, and `set_post_routing_handler` all returned `void`.
This updates them to return `Server&`, allowing chaining between them.

Is there anything else that might also make sense to chain?

I think `set_expect_100_continue_handler`, `set_logger`, `set_file_request_handler`, and others should do the same.
Suggestions?

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -619,6 +619,10 @@ public:
   Server &Delete(const char *pattern, Handler handler);
   Server &Delete(const char *pattern, HandlerWithContentReader handler);
   Server &Options(const char *pattern, Handler handler);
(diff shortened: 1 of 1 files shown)


yhirose:
@00ff0000red, thanks for the pull request. There is no problem to do the same with `set_expect_100_continue_handler`, `set_logger`, `set_file_request_handler`.

00ff0000red:
If I understood that corectly, you're okay with me adding a few more; on it!

yhirose:
@00ff0000red, thanks for your fine contribution!

PR #838, opened 2021-01-22
Add void Handler overload for set_pre_routing_handler

00ff0000red:
I'm confident that most users who use the `Server::set_pre_routing_handler` will likely want to always continue with the request.

Added a void handler callback.


00ff0000red:
Oops, I did `[handler = std::move(handler)]` in the lambda closure, forgot what year this lib is from.

Issue #677, opened 2020-10-03
Segmentation Fault


vader-coder:
When I try to run this code I get a Segmentation Fault.

#include "httplib.h"
#include <iostream>
using namespace std;
int main() {
    cout << "running";
    httplib::Client cli("http://cpp-httplib-server.yhirose.repl.co");
    auto res = cli.Get("/hi");
    if (res->status == 200) {
        cout << res->body << endl;
    }
return 0;
}

yhirose:
@vader-coder, could you run unit tests on your machine? `cd test && make test`? Thanks!

yhirose:
@vader-coder, I cound't reproduce it. Here is my result.

```bash
> cat main.cpp 
#include <iostream>
#include "./httplib.h"

int main() {
  httplib::Client cli("http://cpp-httplib-server.yhirose.repl.co");
  
  auto res = cli.Get("/hi");

  if (res) {
    std::cout << res->status << std::endl;
    std::cout << res->body << std::endl;
  }
}

> g++ -std=c++17 -o main main.cpp

> ./main 
Debug: 12
Debug: text/plain
Debug: Sat, 03 Oct 2020 02:33:29 GMT
Debug: 1.1 google
Debug: close
200
Hello World!
```

Hope it helps.

PixlRainbow:
@vader-coder your code does not check if `res` is nullptr, which it will be if there is no connection to the server (e.g. offline). See yhirose's example

PR #844, opened 2021-01-27
check return value

Kexiii:
The original example can be misleading as the return value of client Get/Post method can be nullptr, we should check it before access "res", otherwise a segfault may occure #677

### DIFF
--- a/README.md
+++ b/README.md
@@ -386,7 +386,7 @@ int main(void)
   httplib::Client cli("localhost", 1234);
 
   if (auto res = cli.Get("/hi")) {
-    if (res->status == 200) {
+    if (res && res->status == 200) {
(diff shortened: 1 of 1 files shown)


yhirose:
@Kexiii, thank you for the pull request, but I don't think it's misleading at all. You can see an actual workable example [here](https://github.com/yhirose/cpp-httplib/blob/master/example/simplecli.cc).

Kexiii:
> @Kexiii, thank you for the pull request, but I don't think it's misleading at all. You can see an actual workable example [here](https://github.com/yhirose/cpp-httplib/blob/master/example/simplecli.cc).

I believe it's better to tell the user directly in the doc that the Result can be nullptr/invalid

PR #845, opened 2021-01-27
Add exception handler

Not-Nik:
This pull request adds the option to handle C++ exceptions caused by user handlers when routing. This allows users to react to exceptions globally by logging them and restarting safely. This may be used over try/catch's within the handlers to avoid repetition in error messages: an API might return the same error message for every failed interaction, so just reacting to all exception instead of reaction to individual exception saves code and reduces binary size.

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,6 +177,16 @@ svr.set_error_handler([](const auto& req, auto& res) {
 });
 ```
 
+### Exception handler
+
+```cpp
+svr.set_error_handler([](const auto& req, auto& res) {
+  res.status = 500;
+  res.set_content("<h1>Error 500</h1><p>Internal Server Error</p>",
+                  "text/html");
+});
+```
+
 ### Pre routing handler
 
 ```cpp
--- a/README.md
+++ b/README.md
@@ -178,12 +178,15 @@ svr.set_error_handler([](const auto& req, auto& res) {
 ```
 
 ### Exception handler
+The exception handler gets called if a user routing handler throws an error.
 
 ```cpp
-svr.set_error_handler([](const auto& req, auto& res) {
+svr.set_exception_handler([](const auto& req, auto& res, std::exception &e) {
   res.status = 500;
-  res.set_content("<h1>Error 500</h1><p>Internal Server Error</p>",
-                  "text/html");
+  auto fmt = "<h1>Error 500</h1><p>%s</p>";
+  char buf[BUFSIZ];
+  snprintf(buf, sizeof(buf), fmt, e.what());
+  res.set_content(buf, "text/html");
 });
 ```
 
--- a/httplib.h
+++ b/httplib.h
@@ -611,9 +611,6 @@ public:
   using HandlerWithContentReader = std::function<void(
       const Request &, Response &, const ContentReader &content_reader)>;
 
-  using ExceptionHandlerWithResponse =
-      std::function<HandlerResponse(const Request &, Response &, std::exception &e)>;
-
   using Expect100ContinueHandler =
       std::function<int(const Request &, Response &)>;
 
@@ -659,7 +656,6 @@ public:
   Server &set_error_handler(HandlerWithResponse handler);
   Server &set_error_handler(Handler handler);
   Server &set_exception_handler(ExceptionHandler handler);
-  Server &set_exception_handler(ExceptionHandlerWithResponse handler);
   Server &set_pre_routing_handler(HandlerWithResponse handler);
   Server &set_post_routing_handler(Handler handler);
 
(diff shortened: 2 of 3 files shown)


yhirose:
@Not-Nik, could you add a unit test to `test/test.cc` and provide a brief usage on README for other users, then I'll review the code. Thanks!

Not-Nik:
Alright, ready for review now :)

README.md
```diff
@@ -177,6 +177,16 @@ svr.set_error_handler([](const auto& req, auto& res) {
 });
 ```
 
+### Exception handler
+
+```cpp
+svr.set_error_handler([](const auto& req, auto& res) {
```
yhirose, line 183: missing `, std::exception &e` in the argument list?

README.md
```diff
@@ -177,6 +177,16 @@ svr.set_error_handler([](const auto& req, auto& res) {
 });
 ```
 
+### Exception handler
+
```
yhirose, line 181: Could you explain the purpose of this method in one or a few sentences? This is to intercept `std::exception &e` from users' routing handlers, but not from cpp-httplib itself, right?

test/test.cc
```diff
@@ -3348,34 +3384,6 @@ TEST(MountTest, Unmount) {
...
-
-  // Give GET time to get a few messages.
-  std::this_thread::sleep_for(std::chrono::seconds(1));
-
-  Client cli("localhost", PORT);
-
-  auto res = cli.Get("/hi");
-  ASSERT_TRUE(res);
-  EXPECT_EQ(500, res->status);
-  ASSERT_FALSE(res->has_header("EXCEPTION_WHAT"));
```
yhirose, line 3372: This test shouldn't be removed. It is for the default exception handler.

httplib.h
```diff
@@ -5064,7 +5078,17 @@ Server::process_request(Stream &strm, bool close_connection,
...
   // Rounting
-  if (routing(req, res, strm)) {
+  bool routed;
+  try {
+    routed = routing(req, res, strm);
+  } catch (std::exception & e) {
+    if (exception_handler_)
+      routed = exception_handler_(req, res, e) == HandlerResponse::Handled;
+    else
+      throw e;
```
yhirose, line 5088: `e` should never be re-thrown here. The original code (Ln 4848-4853) should be moved here instead. If a user really wants to re-throw it (it's dangerous though...), it should be re-thrown from a user's exception handler.

httplib.h
```diff
@@ -608,6 +611,9 @@ class Server {
   using HandlerWithContentReader = std::function<void(
       const Request &, Response &, const ContentReader &content_reader)>;
 
+  using ExceptionHandlerWithResponse =
+      std::function<HandlerResponse(const Request &, Response &, std::exception &e)>;
```
yhirose, line 615: I would not like to have this function which looks confusing to users.

httplib.h
```diff
@@ -652,6 +658,8 @@ class Server {
 
   Server &set_error_handler(HandlerWithResponse handler);
   Server &set_error_handler(Handler handler);
+  Server &set_exception_handler(ExceptionHandler handler);
+  Server &set_exception_handler(ExceptionHandlerWithResponse handler);
```
yhirose, line 662: I would like to remove it. It's really rare to return `Unhanded`.

httplib.h
```diff
@@ -4281,6 +4290,19 @@ inline Server &Server::set_error_handler(Handler handler) {
   return *this;
 }
 
+inline Server &Server::set_exception_handler(ExceptionHandlerWithResponse handler) {
```
yhirose, line 4293: I have explained in the above that we don't need it.

httplib.h
```diff
@@ -4281,6 +4290,19 @@ inline Server &Server::set_error_handler(Handler handler) {
...
+  exception_handler_ = std::move(handler);
+  return *this;
+}
+
+inline Server &Server::set_exception_handler(ExceptionHandler handler) {
+  exception_handler_ = [handler](const Request &req, Response &res, std::exception &e) {
+    handler(req, res, e);
+    return HandlerResponse::Handled;
```
yhirose, line 4301: The handler should be a void function.

httplib.h
```diff
@@ -5064,7 +5078,17 @@ Server::process_request(Stream &strm, bool close_connection,
   }
 
   // Rounting
-  if (routing(req, res, strm)) {
+  bool routed;
+  try {
+    routed = routing(req, res, strm);
+  } catch (std::exception & e) {
+    if (exception_handler_)
+      routed = exception_handler_(req, res, e) == HandlerResponse::Handled;
```
yhirose, line 5086: It can be just as below.

```cpp
exception_handler_(req, res, e);
routed = true;
``

yhirose (commented):
Please see my comments. I appreciate your pull request!

yhirose:
@Not-Nik, the changes look good to me. Thanks for your fine contribution!

Bala2105 (commented):
Reviewed PR #845 and it’s a very practical addition for anyone building robust APIs with cpp-httplib. By introducing a global exception handler, you've made it possible to catch crashes in user defined routes and return a clean 500 response instead of letting the server thread die. The implementation in httplib.h is clever it wraps the routing logic in a try catch block and allows users to define a custom callback to log the error or format a specific error body, which significantly reduces boilerplate code for end-users.

PR #849, opened 2021-01-28
Accept URLs with a trailing slash

isnullxbh:
According to [RFC 3986 §6.2.3](https://tools.ietf.org/html/rfc3986), we can use a trailing slash within URL that has an empty path:

> For example,
>    because the "http" scheme makes use of an authority component, has a
>    default port of "80", and defines an empty path to be equivalent to
>    "/", the following four URIs are equivalent:
> 
>       http://example.com
>       http://example.com/
>       http://example.com:/
>       http://example.com:80/

The regular expression used for URL matching doesn't accept a trailing slash. As result, we get an error like "Connection error", because our HTTP client was initialized with bad parameters.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -6983,7 +6983,7 @@ inline Client::Client(const char *scheme_host_port)
 inline Client::Client(const char *scheme_host_port,
                       const std::string &client_cert_path,
                       const std::string &client_key_path) {
-  const static std::regex re(R"(^(?:([a-z]+)://)?([^:/?#]+)(?::(\d+))?)");
+  const static std::regex re(R"(^(?:([a-z]+)://)?([^:/?#]+)(?::(\d+))?/?)");
 
   std::cmatch m;
   if (std::regex_match(scheme_host_port, m, re)) {


isnullxbh:
Failed test:

```
TEST(ConnectionErrorTest, InvalidHost2) {
  auto host = "httpbin.org/";

  SSLClient cli(host);
#else
  Client cli(host);
#endif
  cli.set_connection_timeout(2);

  auto res = cli.Get("/");
  ASSERT_TRUE(!res);
  EXPECT_EQ(Error::Connection, res.error());
}
```

I'm not sure, but I think that the rejection URLs, that contain a trailing slash, was the only reason why this test was working. Removing this restriction (to a trailing slash) results in a successful connection to httpbin.org.

yhirose:
@isnullxbh, thanks for the pull request, but the parameter `host` in the client constructor is not a URL. It's 'scheme' + 'host' + 'port'.  Hope it helps.

isnullxbh:
@yhirose, ok, thanks.

PR #861, opened 2021-02-13
Windows. Reading interrupt lock in case of http long-polling

DiaLight:
GUI application make async http request to server, but the server is not sending answer untill some server event is triggered. When user close application, the gui thread have to cancel all network threads operations.
In cpp-httplib network thread stucks on select or WSAPoll call, which cannot be interrupted by calling shutdown(sock, SD_BOTH) from the gui thread. Both threads going to lock until more data is receaved or connection is closed. 
Maybe this is my design mistake, but there is solution i made for Windows.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -1792,8 +1792,75 @@ template <typename T> inline ssize_t handle_EINTR(T fn) {
   return res;
 }
 
+#ifdef CPPHTTPLIB_USE_WSA_EVENT
+WSAEVENT g_wsa_event = WSACreateEvent();
+inline void thread_safe_interrupt_reading() {
+  WSASetEvent(g_wsa_event);
+}
+#endif
+
 inline ssize_t select_read(socket_t sock, time_t sec, time_t usec) {
-#ifdef CPPHTTPLIB_USE_POLL
+#ifdef CPPHTTPLIB_USE_WSA_EVENT
+  WSAEVENT event = WSACreateEvent();
+  if(event == WSA_INVALID_EVENT) return -1;
+
+  int status = [&] {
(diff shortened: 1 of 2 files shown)


yhirose:
@DiaLight, thanks for the pull request. Could you also add a unit test in `test/test/cc` that reveals the problem when select/poll is used? Then, I'll reproduce it on my Windows environment and start reviewing your pull request. Thanks!

DiaLight:
@yhirose Done. Here is a timing-based long-polling interrupt test. Also, the fix for Server::listen_internal. I didn't know it was using the select_read method.

yhirose:
@DiaLight, thanks for the update. When I have time, I'll review it and get back to you.

yhirose:
@DiaLight, I just found that there are a number of places where `printf` is used. Could you get rid of those calls?
Since I am still busy on other things, I'll do serious review sometime later. Thanks!

DiaLight:
@yhirose I didn't mean to make a release pull request. It's just a proof of concept. One of the ways to solve the problem in Windows environment. Sorry, didn't mention it earlier.
Every select_read call there creates and closes a WSAEVENT object. Also, g_wsa_event is global, but should be bound to the Client object respectively. My proper implementation will touch on many methods, and you probably won't like it. Therefore, it is better for you to decide how to change the project structure in order to implement interrupts.

yhirose:
@DiaLight, thanks for your reply. If the code is not on production level, I am not going to review it, since I am not too interested it yet. I'll close it for now, but I'll reopen it when other users also request the same behavior.
Thank you for your contribution anyway!

PR #865, opened 2021-02-18
Fix: using udl::operator "" _

x4t3a:
The following error started appearing `error: missing space between ‘""’ and suffix identifier -- using udl::operator ""_;`.

Compiler: `powerpc-fsl-linux-g++ (GCC) 4.9.2` (compiler for powerpc architecture)
C++ version: `--std=c++11`

Easy fix by adding spaces :)

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -2282,7 +2282,7 @@ find_content_type(const std::string &path,
   auto it = user_data.find(ext);
   if (it != user_data.end()) { return it->second.c_str(); }
 
-  using udl::operator""_;
+  using udl::operator "" _;
 
   switch (str2tag(ext)) {
   default: return nullptr;


yhirose:
@x4t3a, thanks for the pull request. The real cause of this problem is the clangformat setting `.clangformt` in this project. I often format httplib.h with it, and the current `.clangformat` removes the space... Here is the current setting:

``
BasedOnStyle: LLVM
AllowShortBlocksOnASingleLine: true
AllowShortCaseLabelsOnASingleLine: true
AllowShortIfStatementsOnASingleLine: true
Cpp11BracedListStyle: true
``

I'll find a setting that can avoid the behavior of clangformat. Thanks.

x4t3a:
@yhirose, I've tried running your clang-format config in [this online formatter](https://zed0.co.uk/clang-format-configurator/) and it seems ok. There are several versions there to try on. What version do you use?
Screenshot 2021-02-18 at 22 59 11

yhirose:
@x4t3a, mine is installed via home brew on macOS. Here is the version.

```
> clang-format --version
clang-format version 10.0.1 
```

Also the line 2271 has the same issue.

yhirose:
I confirmed the happens on the online checker as well.

image

yhirose:
@x4t3a, when you have time, could you try this solution for old gcc compilers? If this workaround works on your environment, I'll put it in httplib.h. Thanks for your help!

x4t3a:
@yhirose, alright, I'll try it. Thank you! :)

yhirose:
I'll close it for now.

liangyepianzhou:
@yhirose Thanks for your answer. I also encountered this problem. I would like to ask what is the minimum g++ standard to solve this problem?

PR #890, opened 2021-03-31
std::chrono::duration for timeout

ShirokovMikhail:
Now it is possible to use
`cli.set_read_timeout(500ms);`
if using c++14

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -4790,7 +4790,7 @@ inline bool Server::listen_internal() {
 
     while (svr_sock_ != INVALID_SOCKET) {
 #ifndef _WIN32
-      if (idle_interval_sec_ > 0 || idle_interval_usec_ > 0) {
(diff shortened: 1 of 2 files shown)


yhirose:
@ShirokovMikhail, thank you for the pull request. It looks very good. But it will break users' code, since the new API isn't compatible with the existing API... So I am not going to merge it to the master for now, but keep this pull request in mind for the future. Thank you very much for your contribution!

ShirokovMikhail:
How about keeping both API options?

yhirose:
Honestly, I am satisfied with the existing API with no problem, even if this suggestion is a bit more expressive.

yhirose:
@ShirokovMikhail, I actually implemented your suggestion to have both interfaces with minimal changes at #891. Does it satisfy your needs?

ShirokovMikhail:
@yhirose Great! Thanks a lot.

PR #901, opened 2021-04-09
Fixed minor code smells

PhilLab:


yhirose:
@PhilLab, thanks for the fine contribution!

PR #904, opened 2021-04-12
Use newer version-flexible TLS/SSL method

w3b-net-au:


yhirose:
@w3b-net-au, thanks for the fine contribution!

PR #911, opened 2021-04-23
Remove redundant call to close_socket

xorian:
The "bind_or_connect" lambda passed by "create_client_socket" calls "close_socket" on some errors.  After the lambda returns false up to the loop in "create_socket" it will also call "close_socket" which will either fail with EBADF or in a multi-threaded program may close a descriptor opened in parallel (which can cause other network sockets of files to be mysteriously closed and possibly replaced with some other descriptor).  The call to "close_socket" in the lambda is not needed, so this removes it.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -2244,7 +2244,6 @@ inline socket_t create_client_socket(
           if (is_connection_error() ||
               !wait_until_socket_is_ready(sock, connection_timeout_sec,
                                           connection_timeout_usec)) {
-            close_socket(sock);
             error = Error::Connection;
             return false;
           }


yhirose:
@xorian, thanks for finding the bug!

xorian:
You're quite welcome, thanks for the library!  A release including the fix would be most appreciated, as some multi-threaded code I work with using this library has recently been having problems with mysteriously closed descriptors.  Some code auditing turned up this issue as a possible cause.

yhirose:
v0.8.8 has just been out. Hope your project will go well with the latest version. :)

PR #912, opened 2021-04-24
reserving before encoding

developerark:
We know it will atleast be s.size(). Gives a little bit of performance boost on average cases.


yhirose:
@developerark, thanks for the improvement!

PR #921, opened 2021-05-01
Fix client.cc code, since res.error() without operator overloading…

aleroot:
I have tried the example client.cc but wasn't compiling with latest Xcode, I had to add operator overloading to make it work.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -804,7 +804,7 @@ enum class Error {
   Compression,
 };
 
-std::ostream& operator << (std::ostream& os, const Error& obj)
+inline std::ostream& operator << (std::ostream& os, const Error& obj)
 {
    os << static_cast<std::underlying_type<Error>::type>(obj);
    return os;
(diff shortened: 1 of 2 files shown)


yhirose:
@aleroot, thank you for the pull request. Could you provide a unit test for the method? Then I'll start reviewing it. Thanks!

aleroot:
I've added a simple unit test, let me know if I can help with anything else. Thanks to you for the project. This library is simple and working well.

httplib.h
```diff
@@ -804,6 +804,12 @@ enum class Error {
   Compression,
 };
 
+std::ostream& operator << (std::ostream& os, const Error& obj)
```
yhirose, line 807: Could you put `inline` keyword at the beginning of the line?

yhirose (changes requested):
Look good to me. Could you make one adjustment as below? Thanks!

yhirose (approved):

yhirose:
@aleroot, thanks for your contribution!

PR #934, opened 2021-05-14
Add option to bypass URL encode of path

bwalex:
I have a bit of a special use case on my hands that requires bypassing the normal URL encode. In particular, I need to talk to a server that expects to see literal `+` in the URL, without having been url-encoded.

I'm hoping you are OK with adding this option to the library!

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -1005,6 +1005,8 @@ public:
   void set_keep_alive(bool on);
   void set_follow_location(bool on);
 
+  void set_url_encode(bool on);
+
   void set_compress(bool on);
 
   void set_decompress(bool on);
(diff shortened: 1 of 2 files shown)


yhirose:
@bwalex, thanks for the pull request. I am OK to merge it. (It sounds a bit dangerous if a user passes a URL including space character(s) accidentally though, I guess it's a trade-off between safely and flexibility and a user's responsibility.) Could you provide a unit test in `test/test.cc` as I usually ask all the contributes. Thanks!

bwalex:
`ClientImpl::copy_settings` seems OK to me - it does copy the new `url_encode_` state.

I've added a test for the feature. It's slightly obscure in that it relies on the `decode_url` for query params to turn a `+` into a space (as otherwise you can't really tell url-encoded vs not apart with the way the server handles the path).

I agree the whole endeavour is on the dangerous side of things - but that's the sort of fun you get to deal with when talking to embedded web servers in microcontrollers.

yhirose:
@bwalex, thank you for your fine contribution!

PR #936, opened 2021-05-16
Make the code more object-oriented;

xth:
Move the common part of Request and Response to Message (new struct);
Reorder some detail functions;
Remove about 70 lines of duplicate code;

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -372,18 +372,62 @@ public:
 
 using Range = std::pair<ssize_t, ssize_t>;
 using Ranges = std::vector<Range>;
-
-struct Request {
-  std::string method;
-  std::string path;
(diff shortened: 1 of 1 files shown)


yhirose:
@xth, thanks for the pull request, but I am ok with the current code even if we see the same member functions in both Request and Response.  To me, introducing another class like Message creates more complication. Also I intentionally put most implementations of members after the line [` * implementation`](https://github.com/yhirose/cpp-httplib/blob/75fdb06696d4e27b590135c1e4a226b4e13e9bd7/httplib.h#L1448-L1450). If you follow the principle, your change won't reduce 70 lines anymore. But I'll keep your fine suggestion!

PR #938, opened 2021-05-17
fix ssesvr use of deleted function

bftjoe:
Examples fail to compile with gcc version 10.2.0 on Ubuntu. 

ssesvr.cc:42:20: error: use of deleted function ‘std::atomic<int>::atomic(const std::atomic<int>&)’
   42 |   atomic_int id_ = 0;

This pull request fixes it.


yhirose:
@bftjoe, thank you for your contribution!

PR #941, opened 2021-05-19
Explicit cast from size_t to uInt

qt1:
strm_.avail_out is defined uInt
it is assigned from buff.size() which is size_t

This fails compilation in the project I integrated this into (possible loss of accuracy warning treated as error).


yhirose (changes requested):
Could you use `static_cast` instead of C-style old cast?

qt1:
Cast changed to ````static_cast````

yhirose:
Thanks for the contribution!

PR #942, opened 2021-05-20
Fix virtual call in ClientImpl::~ClientImpl()

Synss:
This fixes a warning in clang tidy:

> Call to virtual method 'ClientImpl::shutdown_ssl' during
> destruction bypasses virtual dispatch

`ClientImpl::~ClientImpl()` calls `lock_socket_and_shutdown_and_close()`
that itself calls `shutdown_ssl()`.  However, `shutdown_ssl()` is virtual
and C++ does not perform virtual dispatch in destructors, which results
in the wrong overload being called.

This change adds a non-virtual `shutdown_ssl_impl()` function that is
called from `~SSLClient()`.  We also inline `sock_socket_and_shutdown_and_close()`
and removes the virtual call in `~ClientImpl()`.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -1413,6 +1413,7 @@ public:
 private:
   bool create_and_connect_socket(Socket &socket, Error &error) override;
   void shutdown_ssl(Socket &socket, bool shutdown_gracefully) override;
+  void shutdown_ssl_impl(Socket &socket, bool shutdown_socket);
 
   bool process_socket(const Socket &socket,
                       std::function<bool(Stream &strm)> callback) override;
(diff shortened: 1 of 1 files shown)


Synss:
With this patch, you might as well inline and remove `lock_socket_and_shutdown_and_close()`, which now has but a single caller. I can do that myself if you like.

yhirose:
@Synss, could you go ahead to inline and remove `lock_socket_and_shutdown_and_close`?

Synss:
> @Synss, could you go ahead to inline and remove `lock_socket_and_shutdown_and_close`?

Done. I have added eef33857f88553520797110a55fce4599da2d49d to this PR.

yhirose:
@Synss, thanks for the fine contribution!

Issue #930, opened 2021-05-12
Server: get notified if transfer is aborted ?


ottigeda:
In some code using a httplib::Server() I would like to learn when a client aborts the transfer,
or if the file was successfully and completely delivered to the client.
I did think Server::set_error_handler would do the job, but for me the error handler never gets called.
Is this a bug, or did I not fully understand what is the purpose of the error handler ?

In my test I use wget to download a bigger file from a httplib::Server(),
which serves the contents using a content provider.
When I abort the wget I expected that the error handler gets called,
but it does not do that for me.

yhirose:
@ottigeda, `Server:: set_error_handler` isn't for that purpose. It is called when the path doesn't match any of registered request handlers. You can see some examples in `test/test.cc`. There is no way to detect such an abort event on client for now.

PixlRainbow:
> In my test I use wget to download a bigger file from a httplib::Server(),
which serves the contents using a content provider.

You can write your content provider to set a global flag to indicate that a transfer is complete if the last batch of data has been written to sink. There is still a possibility that an abort will not be caught if a client aborts halfway through the transfer of the last batch, but at least you will be catching aborts between any other batch.

PR #946, opened 2021-05-26
content provider: add result callback to content provider

ottigeda:
to learn if the server did successfully complete a client-request or not.

As a follow up to my question in #930, I tried to come up with a solution for it.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -446,11 +446,13 @@ struct Response {
 
   void set_content_provider(
       size_t length, const char *content_type, ContentProvider provider,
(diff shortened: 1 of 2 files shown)


yhirose:
@ottigeda, thanks for the pull request, but unfortunately the code looks problematic... You have to care for all the methods (not only `write_response_with_content`, but `write_response`) which end up calling `write_content_with_provider`.

Anyway, I understand fully what you really wanted to do in #930. I'll ponder over how I can implement it. Thanks for your attempt!

Issue #375, opened 2020-03-10
Basic/digest authentication support on server


pampo:
Hello from Italy and many thanks for this library!

Is it possible to authenticate a remote browser (with basic or digest mode), i.e add authentication support on server?

Regards,

Matteo

yhirose:
@pampo, thank you for the feedback. But unfortunately, authentication support on server is out of scope of this library. (In fact, the authentication support on client is much easier, and it has been already implemented in cpp-httplib. This is because it doesn't require any another resource.)

In order to support the feature on server, there are a number of different ways to do depending on user's requirements and user's environments. For example, how to store and secure user names and passwords? on database, or encrypted hidden, or just plain text file? Supporting digest authentication requires more such as how to manage values such as `nonce` and `cnonce` on server.

So I decided keep cpp-httplib simple by not providing any particular implementation for authentication support on server. It seems like other light-weight HTTP libraries such as Flask are taking the same way. Flask itself doesn't provide authentication, but we can add the feature with an external plugin like Flask-HTTPAuth. But Flask-HTTPAuth may not fit in all the situations. In that case, you have to find another plugin.

You actually receive all you need via HTTP headers from client to handle authentication. It's difficult to handle the digest authentication by yourself, but it's not too difficult to handle the basic authentication on your own. Because the basic authentication header is pretty simple for extracting username and password.

Hope it helps!

pampo:
Many thanks for your kind and exhaustive reply!

Regards, Matteo

PR #952, opened 2021-06-02
Server http digest authentication support

Norman44:
Added HTTP digest authentication feature to Server. The feature is able to authenticate the HTTP digest responses in "MD5" or "MD5-sess" algorithm mode. For now only the "auth" authentication mode is supported. The issue was already discussed but never implemented here: https://github.com/yhirose/cpp-httplib/issues/375

I took nonce calculation as:
H(time-stamp ":" realm <":" private-key>)

Authentication is available for only one username and password.


yhirose:
@Norman44, thank you for the pull request. But I am not going to accept it because of the reason I mentioned [here](https://github.com/yhirose/cpp-httplib/issues/375#issuecomment-597319746). Thanks for your understanding!

Norman44:
Thank you for reviewing. :)

PR #462, opened 2020-05-07
Add meson subproject support

smarttowel:
This PR add meson subproject support.
F.ex., place next wrap in your subprojects directory:
```
[wrap-git]
revision = head
url = https://github.com/smarttowel/cpp-httplib.git
```

In your `meson.build` add next lines:
```
httplib_proj = subproject('httplib')
httplib_dep = httplib_proj.get_variable('httplib_dep')
```

PR #955, opened 2021-06-05
Add header-only Meson support

Tachi107:
This allows users to call `dependency('httplib')` and have the include directory automatically configured. 

This doesn't add any advanced functionality (as you can see by the diff) so there's nothing to maintain (unless you move `httplib.h` in a separate directory; in that case `include_directories('.')` should be changed to `include_directories('newdir')`). In fact, it only makes users' life simpler :)

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,0 +1,7 @@
+project('httplib', 'cpp', license: 'MIT')
+
+httplib_dep = declare_dependency(include_directories: include_directories('.'))
+
+if meson.version().version_compare('>=0.54.0')
+  meson.override_dependency('httplib', httplib_dep)
+endif
--- a/meson.build
+++ b/meson.build
@@ -1,7 +1,7 @@
-project('httplib', 'cpp', license: 'MIT')
+project('cpp-httplib', 'cpp', license: 'MIT')
 
-httplib_dep = declare_dependency(include_directories: include_directories('.'))
+cpp_httplib_dep = declare_dependency(include_directories: include_directories('.'))
 
 if meson.version().version_compare('>=0.54.0')
-  meson.override_dependency('httplib', httplib_dep)
+  meson.override_dependency('cpp-httplib', cpp_httplib_dep)
 endif


yhirose:
@Tachi107, thank you for your pull request, but I am probably not taking it... Please see [this comment](https://github.com/yhirose/cpp-httplib/pull/462#issuecomment-625379529).

Tachi107:
Yep I saw it, but this is to allow Meson users to use the project more easily, it doesn't change the repo's workflow

Tachi107:
It works like this: users can add this repo as a git submodule in a folder called `subprojects`, and when they want to use this project they can simply add `dependency('httplib')` to the list of dependencies of their executable/build target. 

```meson
executable(
  'my-program',
  sources: 'main.cpp',
  dependencies: dependency('httplib')
)
```

yhirose:
@Tachi107, thanks for the comment. Please see the [former discussion](https://github.com/yhirose/cpp-httplib/pull/470#issuecomment-630562373) about the current CMake support, so that you would understand what I was trying to say.

As for CMake, I don't touch it at all, but @sum01 has been kindly and faithfully maintaining it for me up till now when any requests or problem reports come. Likewise, if you would like to accept the same responsibility regarding the Meson support, I am glad to merge it. Please let me know your answer.

Thanks for your understanding!

Tachi107:
I implemented only the bare minimum, so it wouldn't be hard to maintain; happy to accept the responsibility :)

Before merging I have a question though, how should the project be called? "cpp-httplib" or "httplib"?

yhirose:
@Tachi107, great! I chose "cpp-????" as cmake project names in other cpp projects (cpp-peglib, cpp-fstlib and so on). But the current CMakeLists.txt maintained by @sum01 sets "httplib" to the project name. You can pick either one, since I don't have a strong opinion about it. :)
Please let me know when you think the pr is ready to merge.

Tachi107:
I prefer "cpp-httplib" as it is the same name as the repo, and makes Meson's automatic fallback work without issues. 

Ready to merge!

yhirose:
I have just merged it. Thanks for the fine contribution!

Issue #650, opened 2020-09-11
Client Got 401 error code


Zhehua-Hu:
Hi, first of all , goodjob for your library!

I have a server api need basic auth with blank user name and effective password, but `cli.set_basic_auth("", "12345"); `does not achieve goal. I use postman to prove that api works and need blank username.

Any idea about how to solve it?

PixlRainbow:
@yhirose I've identified the offending section: https://github.com/yhirose/cpp-httplib/blob/e9575bcb783d9b731c133be47c50b03130d9ef8d/httplib.h#L4874-L4883

Interestingly, although [RFC 7613](https://tools.ietf.org/html/rfc7613) states that "_A username MUST NOT be zero bytes in length._" this does not appear to be enforced in practice, as [Chrome supports sending zero length usernames](https://stackoverflow.com/a/27044051) and various servers ([example](https://skovpen.org/tst/)) support blank usernames.

yhirose:
@PixlRainbow, thanks for the research. I am tending to lift the restriction due to the fact that a number of servers accept blank username. I'll ponder over it for a while.

@Zhehua-Hu, could you try the following change at line 4874 in httplib.h to see your server now responds to the request?

```cpp
if (!basic_auth_password_.empty()) { 
```

Thanks!

Zhehua-Hu:
Thank you all! It is the correct code that solved my problem! 
Goodjob , guys!

yhirose:
@Zhehua-Hu, thanks for the confirmation. I just merged the change to the master branch.

PR #958, opened 2021-06-10
Fix code err code 401 when the password is empty in base_auth.

CncGpp:
Fix: When setting the base_auth with a non-empty username and an empty password  no auth is insert in the header.
The problem is similar to #650 but for the password.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -5696,7 +5696,7 @@ inline bool ClientImpl::write_request(Stream &strm, Request &req,
     }
   }
 
-  if (!basic_auth_password_.empty()) {
+  if (!basic_auth_password_.empty() || !basic_auth_username_.empty()) {
     req.headers.insert(make_basic_authentication_header(
         basic_auth_username_, basic_auth_password_, false));
   }


yhirose:
@CncGpp, thanks for the pull request, but it looks dangerous to accept blank password... RFC 7613 4.1 says:

> A password MUST NOT be zero bytes in length.  This rule is to be
>   enforced after any normalization and mapping of code points.

As you mentioned in #650, I made a change to allow username to be blank, since it looks a pretty common practice by popular servers. But I am not sure that the same thing can apply to password. Please give me more information why you really need it, or what servers are really allowing it. Thanks!

CncGpp:
I'm  using cpp-httplib to implement a basic client for ArangoDB (HTTP API), the DB password can be empty.

yhirose:
@PixlRainbow, what do you think about this pull request?

PixlRainbow:
I've looked into it; this should handle the case where a password needs to be sent that is blank. However, I found that certain servers may need the password field to be not present (not sent at all). But if your use case only requires the first case, then OK.

I would recommend however to use XOR instead of OR since we might not want to send basic auth header if both username and password are unset.

yhirose:
@CncGpp, could you respond the comment from @PixlRainbow?

CncGpp:
Sorry for the delay, wouldn't using XOR break the case where both userand pass are set? I think the OR is fine, when they are both empty it does not enter the if.

PixlRainbow:
ahh my bad, I had mistaken the conditions backwards :D

yhirose:
Thanks for the replies. I just merged it. Thanks for the contribution!

PR #961, opened 2021-06-16
Fix mulitple content length headers err

umbrellary:


yhirose:
@umbrellary, could you provide a unit test case and explain the situation where this problem happens? Then, I'll review the pull request. Thanks!

umbrellary:
@yhirose when Digest authentication is enabled, the problem in the picture below will occur when requesting POST.
Thanks!
image

yhirose:
@umbrellary, thank you for the update. Could you also provide lines in httplib.h where the first 'Connection: clone' and 'Content-Length: 3' are set? Thanks!

umbrellary:
My guess is that the client has already set it once when the server returns 401 for the first request, and it has been set again when the second request is made.
first：
image
second：
image

yhirose:
@umbrellary, sorry but I still don't understand your comment... Could provide the except lines in httplib.h where the first 'Connection: close' and 'Content-Length: 3' are set as I requested in my previous comment? That information will tell me more insight about the problem.

umbrellary:
Sorry, I didn't answer clearly. The two positions are highlighted in the picture below.

# 'Connection: close'

# 'Content-Length: 3'

yhirose:
@umbrellary, thanks for your contribution!

Issue #899, opened 2021-04-06
Does space matter? using udl::operator""_;


MugiwaraJAK:
Is this correct having definition with a space and directive without? 

I don't know if this is a problem with syntax or just an issue with eclipse. 

```
namespace udl {

inline constexpr unsigned int operator"" _(const char *s, size_t l) {
  return str2tag_core(s, l, 0);
}

} // namespace udl
```
`
  using udl::operator""_;
`

yhirose:
@MugiwaraJAK, please see #865. clangformat doesn't put a spece between `""` and `_`. It's totally legal according to C++ spec, but some older compilers incorrectly report it as an error. Hope it helps.

PR #963, opened 2021-06-17
fix compiling on some old Arm compilers

folibis:
That should fix compiling on some old Arm compilers. For example it crashed for me on gcc version 4.8.3 20140320 for Arm with error: `missing space between '""' and suffix identifier`


yhirose:
Same as #899.

PR #964, opened 2021-06-17
wxwidgets compatibility

simonedlund:
operator""_ replaced by operator""_t

My compiler (cl.exe 19.28.29334, VS2019, with /std:c++17) does not like the ""_ , using ""_t works better

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -2334,7 +2334,7 @@ inline unsigned int str2tag(const std::string &s) {
 
 namespace udl {
 
-inline constexpr unsigned int operator"" _(const char *s, size_t l) {
+inline constexpr unsigned int operator"" _t(const char *s, size_t l) {
   return str2tag_core(s, l, 0);
 }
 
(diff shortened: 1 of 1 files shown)


simonedlund:
Further investigation revealed that in wxwidgets, someone decided to steal the underscore:

 #define _(s)                               wxGetTranslation((s))

Which destroys the operator""_ in httplib. So I shall not blame the compiler.

Anyway, the proposed change makes cpphttplib compatible with wxwidgets, take it or leave it :-)

yhirose:
@simonedlund, thanks for the suggestion. In fact, I already use the udl in my other projects, and others might use it in their case as well... So I'll keep it as it is. Thanks for your understanding!

https://github.com/yhirose/pl0-jit-compiler/blob/c90945a724c757d93577f41e18d45f7753f53a88/pl0.cc#L139-L151

yhirose:
@simonedlund, thanks for the suggestion. I'll thing about it.

PR #972, opened 2021-06-26
content type not correctly identified

voiser:
Ran into this bug in OS X with Clang


yhirose:
@voiser, thanks, but it's correct. If you still think your change is correct, could you add a unit test to validate it? Thanks!

voiser:
You're absolutely right, my brain seems not to be working today.

Sorry for this, and thank you for the fantastic job. I hope I can contribute to this project with something that is not a fiasco.

PR #980, opened 2021-06-30
Ignore gcc `maybe-uninitialized` warning.

mgautierfr:
Recent version of gcc (fedora 34, gcc 11.1.1) warn about the use `buf` where it may be
uninitialized.
But we are using `buf` to initialize it so it should be ok to use it.

I'm not sure this is the right way to fix it.
The other solution I see would be to initialize the buffer with 0 which would be somehow counterproductive here.
I let you decide :)

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -4098,11 +4098,14 @@ inline ssize_t Stream::write_format(const char *fmt, const Args &...args) {
   const auto bufsiz = 2048;
   std::array<char, bufsiz> buf;
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
 #if defined(_MSC_VER) && _MSC_VER < 1900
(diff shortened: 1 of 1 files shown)


yhirose:
@mgautierfr, thank you for the pull request. This warning isn't correct, or too strict, and I don't want to add `#pragma` which is only for suppressing the warning. For now, could you just put the pragma in your application's source code, or build system file like Makefile or CMakeLists.txt as bellow? Sorry that I can't give you more convenient solution.

```
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
#include <httplib.h>
#pragma GCC diagnostic pop
```

mgautierfr:
I agree this is not the best way to fix it (and my fix was broken with clang where it complains about the unknown `-Wmaybe-uninitialized`.)

I will handle this issue on my side for now. But I suppose you will have to correct this somehow in the future. Now you know :)

Issue #978, opened 2021-06-30
Provide a mapping function from Error to String


gjasny:
Hello,

It would be nice if you could provide a `to_string` function for the `httplib::Error` class so that callers could log a human-readable string.

Thanks,
Gregor

yhirose:
@gjasny, thanks for the feedback. I'll take a look at it when I have time. Thanks.

PR #999, opened 2021-07-17
Add httplib::Error to std::string function

gjasny:
Fixes: #978

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -798,8 +798,31 @@ enum class Error {
   Compression,
 };
 
+inline std::string to_string(const Error error) {
+  switch (error) {
+  case Error::Success: return "Success";
+  case Error::Connection: return "Connection";
+  case Error::BindIPAddress: return "BindIPAddress";
+  case Error::Read: return "Read";
+  case Error::Write: return "Write";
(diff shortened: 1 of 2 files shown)


httplib.h
```diff
@@ -811,8 +811,30 @@ enum class Error {
...
+  case Error::Write: return "Write";
+  case Error::ExceedRedirectCount: return "ExceedRedirectCount";
+  case Error::Canceled: return "Canceled";
+  case Error::SSLConnection: return "SSLConnection";
+  case Error::SSLLoadingCerts: return "SSLLoadingCerts";
+  case Error::SSLServerVerification: return "SSLServerVerification";
+  case Error::UnsupportedMultipartBoundaryChars:
+    return "UnsupportedMultipartBoundaryChars";
+  case Error::Compression: return "Compression";
+  case Error::Unknown: break;
```
yhirose, line 829: Can you make it like this?

```cpp
  case Error::Unknown:   return "Unknown";
  default: break;
  }

  return "Invalid";
}
```

yhirose:
@gjasny, please see my review. Also it seems like the unit tests on GitHub Actions failed...

yhirose:
Thanks for the corrections!

PR #1000, opened 2021-07-17
Use googletest 1.11.0

gjasny:


yhirose:
@gjasny, great! Thanks for the fine contribution!

PR #1002, opened 2021-07-19
Fix gzip compression/decompression over 4 GiB data size

yosh-matsuda:
I found gzip compression feature does not work correctly for large data (over 4 GiB).  
It seems that compression should repeat with the correct chunk size until the end of data.
I have checked this fixes the issue with the large-data post request. 

The same issue can occur with gzip decompression and has been fixed, but I haven't tested it yet.   
Could you review this PR?

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -2547,28 +2547,40 @@ public:
                 Callback callback) override {
     assert(is_valid_);
 
-    auto flush = last ? Z_FINISH : Z_NO_FLUSH;
-
-    strm_.avail_in = static_cast<decltype(strm_.avail_in)>(data_length);
-    strm_.next_in = const_cast<Bytef *>(reinterpret_cast<const Bytef *>(data));
-
-    int ret = Z_OK;
-
-    std::array<char, CPPHTTPLIB_COMPRESSION_BUFSIZ> buff{};
     do {
-      strm_.avail_out = static_cast<uInt>(buff.size());
-      strm_.next_out = reinterpret_cast<Bytef *>(buff.data());
+      constexpr size_t max_avail_in =
+          std::numeric_limits<decltype(strm_.avail_in)>::max();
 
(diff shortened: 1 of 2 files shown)


yhirose:
@yosh-matsuda, thanks for the pull request. Could you add at least one unit test to `test/test.cc` to validate your fix before I start reviewing?

yosh-matsuda:
@yhirose I added the gzip compression and decompression test for large random data. This test fails on the master branch as follows and I checked it will be fixed. Please note that the test requieres about 13GB memory.

```
Note: Google Test filter = GzipDecompressor.LargeRandomData
[==========] Running 1 test from 1 test case.
[----------] 1 test from GzipDecompressor
[ RUN      ] GzipDecompressor.LargeRandomData
test.cc:2913: Failure
Value of: decompressed_data.size()
  Actual: 134217728
Expected: data_size
Which is: 4429185024
[  FAILED  ] GzipDecompressor.LargeRandomData (28526 ms)
[----------] 1 test from GzipDecompressor (28526 ms total)
```

yhirose:
@yosh-matsuda, thanks for adding the unit test. I have reviewed the code and looks very good! I personally feel that [the zlib usage example page](https://zlib.net/zlib_how.html) should use this code instead. :)

The only problem is that the unit tests run on 'ubuntu-latest'in GitHub Actions is always killed (and probably on macOS as well)... I am not sure of the exact reason, but it seems like it's due to timeout. (On my iMac with 3.8 GHz Quad-Core Intel i5 and 24GB Memory, this test takes about 3 minutes. So the total time of tests is about 5 minutes.)
image

Fortunately, the unit tests run on Windows works with no problem. So I surround the test with `#ifdef _WIN32 ... #endif` at ccbddd88423d9ce364bfbaada130c36fa2fc1bd6, so that we at least run this test on Windows.

Anyway, thanks for the excellent contribution!

Issue #1003, opened 2021-07-20
https redirect error


xxrlzzz:
hey guys, I'm using this library to do little project, and with a problem. 

### Problem Description:

When I want send http request with `Client`,  it may fail to get response when the server require redirect from http request to https request. 

### My way to resolve

I found I need to call `set_ca_cert_path`, but it lost during redirect, `copy_settings` don't copy ca cert path, and I found `client_cert_path_` and `client_key_path_`, but no else place was using this.   
`SSLClient` actually using `ca_cert_file_path_` and `ca_cert_dir_path_`. I don't know relation between them.  
Currently I'm setting `ca_cert_file_path_` and `ca_cert_dir_path_` in `copy_settings`. It works for me. It may be a bug or just my wrong usage. looking for your reply.

yhirose:
@xxrlzzz, thanks for the feedback. It sounds like a bug. Could you send a pull request with your fix and a unit test if possible? Then, I'll review it soon. Thanks!

yhirose:
Resolved by #1004.

PR #1004, opened 2021-07-21
[Fix] ca_cert_path/ce_cert_store lose

xxrlzzz:
When redirect from http to https, user setting for ca_cert will lose

issue: #1003


yhirose:
@xxrlzzz, the code looks excellent. Thanks for your contribution!

PR #1011, opened 2021-07-30
Improve split script

jrosdahl:
- Added shebang and made the script executable.
- Added help text.
- Added -o/--out argument for specifying output directory.
- Added -e/--extension argument for specifying file extension of the
  implementation file.
- Made the script find httplib.h next to split.py instead of the current
  working directory. This makes it possible to call the script from
  another directory.
- Simplified code structure slightly.
- Improved variable naming to follow Python conventions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -749,10 +749,20 @@ res->body; // Compressed data
 Split httplib.h into .h and .cc
 -------------------------------
 
-```bash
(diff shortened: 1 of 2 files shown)


yhirose:
Thanks for the improvement!

PR #1012, opened 2021-07-30
Avoid hardcoded ports in RedirectToDifferentPort.Redirect test

jrosdahl:
The RedirectToDifferentPort.Redirect test assumes that port 8080 and
8081 are available on localhost. They aren’t on my system so the test
fails. Improve this by binding to available ports instead of hardcoded
ones.


yhirose:
Yes, this is better!

PR #1013, opened 2021-07-30
Add missing template method implementations

### LINKED ISSUE #1008: Implementations of templated methods are missing when splitting into .h and .cc
Thanks for writing cpp-httplib!

I made an attempt to use the non-header-only version of httplib (in [ccache](https://ccache.dev)) to reduce the time to recompile code that uses the library. However, the templated implementation of e.g. `set_connection_timeout` ends up in httplib.cc and therefore results in a linker error since the needed template has not been instantiated.

It looks like these methods are affected:

```c++
template <typename T>
T Request::get_header_value(const char *key, size_t id = 0) const;

template <typename T>
T Response::get_header_value(const char *key, size_t id = 0) const;

template <typename... Args>
ssize_t Stream::write_format(const char *fmt, const Args &...args);

template <class Rep, class Period>
Server &Server::set_read_timeout(const std::chrono::duration<Rep, Period> &duration);

template <class Rep, class Period>
Server &set_write_timeout(const std::chrono::duration<Rep, Period> &duration);

template <class Rep, class Period>
Server &set_idle_interval(const std::chrono::duration<Rep, Period> &duration);

template <typename T>
T Result::get_request_header_value(const char *key, size_t id = 0) const;

template <class Rep, class Period>
void ClientImpl::set_connection_timeout(const std::chrono::duration<Rep, Period> &duration);

template <class Rep, class Period>
void ClientImpl::set_read_timeout(const std::chrono::duration<Rep, Period> &duration);

template <class Rep, class Period>
void ClientImpl::set_write_timeout(const std::chrono::duration<Rep, Period> &duration);

template <class Rep, class Period>
void Client::set_connection_timeout(const std::chrono::duration<Rep, Period> &duration);

template <class Rep, class Period>
void Client::set_read_timeout(const std::chrono::duration<Rep, Period> &duration);

template <class Rep, class Period>
void Client::set_write_timeout(const std::chrono::duration<Rep, Period> &duration);
```

Would you welcome a pull request that moves some or all of these into the header file part? If not, perhaps some caveat text could be added to the [Split httplib.h into .h and .cc](https://github.com/yhirose/cpp-httplib#split-httplibh-into-h-and-cc) section?

Related: I tried to modify the test suite to run against a split version to make it possible to verify that the split version works, but I realized it will be impossible to do without moving the tests that depend on `details` into a separate test file. Is that something that would be of interest?

jrosdahl:
When using the split version of httplib.h the templated implementation of e.g. `Client::set_connection_timeout` ends up in httplib.cc and therefore results in a linker error since the needed template specialization has not been instantiated. Fix this by moving the implementation of template methods into the part that ends up in httplib.h after the split.

Fixes #1008.

I have tested this with "Approach A" described in #1008. The tests don't have full coverage, but it gives some confidence at least. I'll submit a follow-up PR with my suggested test suite changes if this PR, #1011 and #1012 are accepted.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -257,7 +257,7 @@ namespace detail {
 
 template <class T, class... Args>
 typename std::enable_if<!std::is_array<T>::value, std::unique_ptr<T>>::type
-make_unique(Args &&...args) {
+make_unique(Args &&... args) {
   return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
 }
 
@@ -490,7 +490,7 @@ public:
   virtual socket_t socket() const = 0;
 
   template <typename... Args>
-  ssize_t write_format(const char *fmt, const Args &...args);
+  ssize_t write_format(const char *fmt, const Args &... args);
   ssize_t write(const char *ptr);
   ssize_t write(const std::string &s);
 };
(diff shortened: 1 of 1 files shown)


yhirose:
The change looks very good to me. Please go ahead to submit a follow-up PR. Thanks a lot!

PR #1015, opened 2021-07-31
Add test of httplib.h split into .h + .cc

jrosdahl:
In order to test the split version (.h + .cc via split.py):

- Added a test_split program in the test directory whose main purpose is to verify that it works to compile and link the test case code against the split httplib.h version.
- Moved types needed for test cases to the “header part” of httplib.h. Also added forward declarations of functions needed by test cases.

The changes to httplib.h just move code around (or add forward declarations), with one exception: `detail::split` and `detail::process_client_socket` have been converted to non-template functions (taking an std::function instead of using a type parameter for the function) and forward-declared instead. This avoids having to move the templates to the “header part”.

As discussed in #1008.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -1613,10 +1613,191 @@ Client::set_write_timeout(const std::chrono::duration<Rep, Period> &duration) {
   cli_->set_write_timeout(duration);
 }
 
+/*
+ * Forward declarations and types that will be part of the .h file if split into
+ * .h + .cc.
+ */
+
+std::pair<std::string, std::string> make_range_header(Ranges ranges);
+
+std::pair<std::string, std::string>
+make_basic_authentication_header(const std::string &username,
+                                 const std::string &password,
+                                 bool is_proxy = false);
+
+namespace detail {
+
+std::string encode_query_param(const std::string &value);
+
+void read_file(const std::string &path, std::string &out);
+
+std::string trim_copy(const std::string &s);
+
+void split(const char *b, const char *e, char d,
+           std::function<void(const char *, const char *)> fn);
+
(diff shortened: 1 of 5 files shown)


httplib.h
```diff
@@ -1914,81 +2096,70 @@ template <class Fn> void split(const char *b, const char *e, char d, Fn fn) {
...
 }
 
-// NOTE: until the read size reaches `fixed_buffer_size`, use `fixed_buffer`
-// to store data. The call can set memory on stack for performance.
-class stream_line_reader {
-public:
-  stream_line_reader(Stream &strm, char *fixed_buffer, size_t fixed_buffer_size)
-      : strm_(strm), fixed_buffer_(fixed_buffer),
-        fixed_buffer_size_(fixed_buffer_size) {}
+stream_line_reader::stream_line_reader(Stream &strm, char *fixed_buffer,
```
yhirose, line 2099: Need `inline`

yhirose:
@jrosdahl, thanks for the changes. They look good to me. I found a number of implementations of functions (like `type Class::method(...) {`) are missing `inline` keyword. Could you makes sure to put it to them?

httplib.h
jrosdahl: Sure thing. I'll make up for the mistake by adding a test that catches such errors.

jrosdahl:
> Could you makes sure to put it to them?

Done.

yhirose:
Everything now looks good. Thanks for your fine contribution!

yhirose:
I particularly appreciated the idea to verify any inline keyword is missing with `test/include_httplib.cc`. :)

PR #1017, opened 2021-08-02
Clarify that content_reader is blocking and improve example

gamelaster:
In the **Receive content with a content receiver** part, the example misses response content when multipart form data is sent.

Also I think it's good to clarify that `content-reader` is blocking until every form data is uploaded.

### DIFF
--- a/README.md
+++ b/README.md
@@ -252,6 +252,8 @@ svr.Post("/multipart", [&](const auto& req, auto& res) {
 
 ### Receive content with a content receiver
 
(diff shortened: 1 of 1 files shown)


gamelaster:
Also, why is header and receiver functions returns true? What happens if return false is returned? (I tried to trace it in source code but I wasn't able to trace it :D )

yhirose:
@gamelaster, thanks for the suggestion!

PR #1021, opened 2021-08-10
Don't define INVALID_SOCKET if it has been defined already

TobaYn3:
If including pcap.h from libpcap before httplib.h, INVALID_SOCKET has already been defined.
Therefore, I added an ifndef.


yhirose:
@TobaYn3, thank you for the improvement!

PR #1026, opened 2021-08-21
Support offline unit tests only

pemensik:
Some environments provide no network connectivity. Allow simple way to
skip tests requiring working internet connection. Just defining
-DOFFLINE would omit tests failing without connection.

### DIFF
--- a/test/test.cc
+++ b/test/test.cc
@@ -18,6 +18,11 @@
 #define CLIENT_CERT_FILE "./client.cert.pem"
 #define CLIENT_PRIVATE_KEY_FILE "./client.key.pem"
 
+/* Skip rest of the test if connection failed */
+#define TEST_ONLINE(res) \
+  if (!res && res.error() == Error::Connection) \
+    { GTEST_SKIP(); return; }
+
 using namespace std;
 using namespace httplib;
 
@@ -368,6 +373,7 @@ TEST(ChunkedEncodingTest, FromHTTPWatch) {
 
   auto res =
       cli.Get("/httpgallery/chunked/chunkedimage.aspx?0.4153841143030137");
+  TEST_ONLINE(res);
   ASSERT_TRUE(res);
 
   std::string out;
@@ -396,6 +402,7 @@ TEST(ChunkedEncodingTest, WithContentReceiver) {
                 body.append(data, data_length);
                 return true;
               });
+  TEST_ONLINE(res);
   ASSERT_TRUE(res);
 
   std::string out;
@@ -428,6 +435,7 @@ TEST(ChunkedEncodingTest, WithResponseHandlerAndContentReceiver) {
         body.append(data, data_length);
         return true;
       });
+  TEST_ONLINE(res);
   ASSERT_TRUE(res);
 
   std::string out;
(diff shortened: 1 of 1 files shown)


pemensik:
I have created [Fedora RPM package](https://src.fedoraproject.org/rpms/cpp-httplib) for your library. I wanted to use tests during builds, but Fedora packages are built inside mock environment, where connection to outside world is blocked on purpose. I found no existing way to skip those tests. Hence I am proposing quick and dirty way to do that.

Better fix would be ability to use localhost server for tests, for example nginx with test configuration.

yhirose:
@pemensik, thanks for your interest in this library. I took a quick look at it, but it drops a lot of `ifndef OFFLINE` into test.cc. Also when adding new test cases accessing other sites like httplib.org in the future, it's pretty easy to forget putting `#ifndef OFFLINE`. Is there any better way to handle this situation on your environment without touching cpp-httplib project? Thanks for your consideration!

pemensik:
There is no way to provide connection for testing, I would have to disable all tests during builds.  One of reasons I wanted the package with tests enabled is testing on all platforms. Main reason is to let our builder test it on all its platforms, which I have no simple access to otherwise. Also it makes test results more stable, because it would not depend on third party services to be running and responding expected way. Also a privacy one, it does not report each test online.

It would help if tests requiring external connectivity could skip test if they could not resolve hostnames for example. Ie reporting not FAILED, but SKIPPED on tests connecting yahoo.com and similar. I am not sure existing error reporting allows simple way to implement it.

Alternative might be own provision of http server on localhost and redirecting names to it. Unfortunately our builders on Fedora do not allow running as root, leaving only non-privileged options in place. I cannot find similar easy enough approach.

Last alternative might be renaming just failing tests with DISABLED_ prefix, as suggested in https://stackoverflow.com/questions/7208070/googletest-how-to-skip-a-test. I guess long sed recipe could be used, but I would like some solution working also for other people.

pemensik:
I made alternative change, perhaps more clear and simple to understand. Each test reaching online checks first time result. If it is connection failure, just skip the test. Advantage is clear message those tests did not fail or pass, just were skipped.

In order to stay in shape, it might help include network-less testing. Here is what I am using to test in on GNU/Linux:

```
sudo ip netns add nonet
sudo ip -netns nonet link set lo up
sudo ip netns nonet exec ./test
```

Are you using some kind of continuous integration, where this test could be added? Perhaps at least make it optional target from make?

yhirose:
@pemensik, thanks for the update, I like it better than the previous one. I have actually another idea. How about to just adding suffix string `_Online` to such test cases like `FromHTTPWatch_Online`. Then we can skip those tests with `--gtest_filter=-*_Online`. The advantage of this approach is that we don't have to add any additional lines, also it will completely skip those tests even without checking online connectivity.

pemensik:
Ah, didn't know such filter is possible. Common prefix or suffix would work fine if that is what you prefer.

yhirose:
I added `_Online` to all the test cases that access external servers. Could you check to see if `--gtest_filter=-*_Online` on your build system. After you confirm it, I'll bump up the version to v0.9.3. Thanks!

pemensik:
My filter has to look a bit different (``./test --gtest_filter='-*.*_Online'``), but it passes in offline mode. Thanks!

My pull request is therefore not necessary, closing it.

PR #1033, opened 2021-09-02
Full Meson support

Tachi107:
cpp-httplib can be now built with Meson even in compiled library mode.

The library is built with LTO, supports OpenSSL, zlib and Brotli, and the build system also generates a pkg-config file when needed.

Compared to the CMake file this one is quite small (more than five times smaller!), and maintaining it won't be an issue :) (and I'll keep maintaining it as I promised)

Edit: Note that I set the project version to 0.9.4 even if the released version is 0.9.3 so that when you'll release again (hopefully soon, so that people can use the improved Meson support) you'll don't have to update that yourself

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -5,7 +5,6 @@
 project(
   'cpp-httplib',
   'cpp',
-  version: '0.9.4',
   license: 'MIT',
   default_options: [
     'buildtype=release',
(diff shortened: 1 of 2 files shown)


yhirose:
@Tachi107, thanks for the contribution, but I can't accept this pull request because the file embeds the version number string. Could you take look at #505 where @sum01 did an excellent job to support automatic versioning to CMake?

Tachi107:
Hi, I can't implement the same thing with Meson, as it requires that `project()` is called before any other function. The only way to set a project version from an external file is to have a plaintext file containing just the version number in the repository, but that would also require to be manually updated. 

If you prefer I can remove the version number from `project()` and use a similar approach to the one used in CMake, but having the version directly in project is more convenient (it automatically sets library version / soname, pkg-config version, and allows users using cpp-httplib as a subproject to look only for a specific version of the library).

If you decide to keep it, I will watch the releases and every time a new release is made I can submit a PR updating the version number in advance

Tachi107:
Hey @yhirose, edited as you requested :)

Ready to merge!

yhirose:
@Tachi107, great. Thanks for your fine contribution!

PR #1036, opened 2021-09-05
Set error variable for failed write_data

gjasny:
Hello,

when looking for the origin of a `Unknown` error I noticed this `write_data` caller does not set the error variable.

Thanks,
Gregor


yhirose:
@gjasny, thanks for the pull request, and looks good! I'll merge it right away.

PR #1044, opened 2021-09-10
meson: add tests

Tachi107:
This integrates the "main" test suite (`test/test.cc`) in Meson.

This allows to run the tests in the CI with the Meson-built version of the library to ensure that nothing breaks unexpectedly.

It also simplifies life of downstream packagers, that do not have to write a custom build script to split the library and run tests but can instead just let Meson do that for them.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -7,6 +7,7 @@ project(
   'cpp',
   license: 'MIT',
   default_options: [
+    'cpp_std=c++11',
     'buildtype=release',
     'b_ndebug=if-release',
     'b_lto=true',
@@ -71,7 +72,7 @@ if brotli_found_all
   args += '-DCPPHTTPLIB_BROTLI_SUPPORT'
 endif
 
-cpp_httplib_dep = dependency('', required : false)
+cpp_httplib_dep = dependency('', required: false)
 
 if get_option('cpp-httplib_compile')
   httplib_ch = custom_target(
@@ -105,4 +106,8 @@ endif
 
 if meson.version().version_compare('>=0.54.0')
   meson.override_dependency('cpp-httplib', cpp_httplib_dep)
+endif
+
+if get_option('cpp-httplib_test')
+  subdir('test')
 endif
--- a/test/www/dir/meson.build
+++ b/test/www/dir/meson.build
@@ -1,0 +1,7 @@
+# SPDX-FileCopyrightText: 2021 Andrea Pappacoda
+#
+# SPDX-License-Identifier: MIT
+
+configure_file(input: 'index.html', output: 'index.html', copy: true)
+configure_file(input: 'test.abcde', output: 'test.abcde', copy: true)
+configure_file(input: 'test.html',  output: 'test.html',  copy: true)
--- a/test/www2/dir/meson.build
+++ b/test/www2/dir/meson.build
@@ -1,0 +1,6 @@
+# SPDX-FileCopyrightText: 2021 Andrea Pappacoda
+#
+# SPDX-License-Identifier: MIT
+
+configure_file(input: 'index.html', output: 'index.html', copy: true)
+configure_file(input: 'test.html',  output: 'test.html',  copy: true)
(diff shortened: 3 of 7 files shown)


Tachi107:
Tests failed because `ServerTest.SlowPostFail` fails sporadically on macOS, but you can see that they all passed in my repo, [here](https://github.com/Tachi107/cpp-httplib/actions/runs/1222881106).

meson.build
```diff
@@ -7,6 +7,7 @@ project(
   'cpp',
   license: 'MIT',
   default_options: [
+    'cpp_std=c++14',
```
yhirose, line 10: This library supports C++11. Could change it to 'cpp_std=c++11'?

test/www/dir/meson.build
```diff
@@ -0,0 +1,7 @@
+# SPDX-FileCopyrightText: 2021 Andrea Pappacoda
```
yhirose, line 1: Could you remove this file from 'test/www/dir' directory?

test/www2/dir/meson.build
```diff
@@ -0,0 +1,6 @@
+# SPDX-FileCopyrightText: 2021 Andrea Pappacoda
```
yhirose, line 1: Could you remove this file from 'test/www2/dir' directory?

test/www3/dir/meson.build
```diff
@@ -0,0 +1,6 @@
+# SPDX-FileCopyrightText: 2021 Andrea Pappacoda
```
yhirose, line 1: Could you remove this file from 'test/www3/dir' directory?

yhirose:
@Tachi107, thank you for the pull request. I don't know about meson build system though, I reviewed the code. Could you follow my comments? Thanks!

test/www/dir/meson.build
Tachi107: Unfortunately I can't.
The "old" tests are run in-tree, and require `ca-bundle.crt`, `image.jpg` `www/dir*/*` to be ran. 
In-tree builds are disallowed in Meson (and discouraged in CMake), so it is necessary to copy the required files in the build directory to recreate the same directory structure as the one found in the source tree (that is, `test/ca-bundle.crt` → `build/test/ca-bundle.crt`, etc). Copying files to the build dir is accomplished by the `configure_file(input: 'filename', output: 'filename', copy: true)` function, where the input is a file name relative to the same directory as the current `meson.build` file. Copying a file from a subdirectory is not allowed, as Meson guaranties that a file in a certain file in the build dir (like `build/test/www/dir/index.html`) has to come from the same path of the source tree (`test/www/dir/index.html`).
This makes debugging easier, but obviously has the downside of having to create a `meson.build` file in the same dir as the files you want to copy. So I can't remove these three files :/

yhirose: 'www', 'www2' and 'www3' are just htdocs folders, and I may add more such folders like 'www4' in the future. I would not like to remember adding 'meson.build' file to each new data directory, and it puts an extra burden on me. I personally don't care for testing on Meson build system, because I only use just plain 'Make' files. (I also don't use CMake file in this project personally, and I think it is not used for testing as well.) Thanks for your understanding!

Tachi107: I understand your concern, but don't worry, the worst that could happen if you don't add the meson.build file is a failed test, better than no tests at all :)

I've committed to maintain the Meson part of this repo, so if something Meson-related has issues I'll fix it as soon as possible (the same way @sum01 is the "maintainer" of the CMake portion of cpp-httplib).

If for whatever reason I stop maintaining something that needs to be updated (like the tests) you're always free to remove it (like you did with CMake some time ago).

Having tests that ensure that Meson users do not use a broken build of the library (when not using the header-only version) are important to ensure that cpp-httplib will always work fine for everyone.

yhirose: Glad to see your willingness! I'll merge it.

PR #1051, opened 2021-09-16
Make Client move-constructible

### LINKED ISSUE #1050: Add move constructor for Client
Since `Client` already uses the PIMPL pattern, it seems reasonably to provide a default move constructor for it. This enables classes that have a `Client` member variable to be movable.

If using C++11 or later, we only need to add
```c++
Client(Client &&) = default;
```
unless we want to support older standards.

ZizhengTai:
Resolves #1050

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -1161,6 +1161,8 @@ public:
                   const std::string &client_cert_path,
                   const std::string &client_key_path);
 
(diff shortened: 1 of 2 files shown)


yhirose:
@ZizhengTai, thanks for the pull request and looks very good to me!

PR #1054, opened 2021-09-16
test: avoid infinite loop when IPV6 is unsupported

### LINKED ISSUE #1053: `RedirectFromPageWithContentIP6` test locks if CONFIG_IPV6 is disabled in the Linux kernel
When the Linux kernel is built without IPV6 support, the `RedirectFromPageWithContentIP6` test never completes, as it doesn't go past `while(!svr.is_running())` (line 974).

https://github.com/yhirose/cpp-httplib/blob/3c522386e961d61768ea527d04713b5402356dd4/test/test.cc#L955-L1013

I believe that this happens because when IPV6 support isn't available `svr.listen("::1", 1234)` never actually starts anything, so the condition `!svr.is_running()` will always remain true.

A possible solution would be to only wait for a few seconds, and if svr does not start abort the test

Edit: Since `Server::listen()` returns `false` on failure, it may be more appropriate to do something like `ASSERT_TRUE(svr.listen("::1", 1234))`

Edit2: `ASSERT_TRUE` in the other thread doesn't work

Tachi107:
Fixes #1053

### DIFF
--- a/test/test.cc
+++ b/test/test.cc
@@ -971,8 +971,15 @@ TEST(RedirectFromPageWithContentIP6, Redirect) {
 
   auto th = std::thread([&]() { svr.listen("::1", 1234); });
 
-  while (!svr.is_running()) {
+  // When IPV6 support isn't available svr.listen("::1", 1234) never
+  // actually starts anything, so the condition !svr.is_running() will
+  // always remain true, and the loop never stops.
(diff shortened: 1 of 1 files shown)


yhirose:
@Tachi107, thanks for the change. I'll merge it right away!

PR #1057, opened 2021-09-25
Avoid min/max macro expansion on Windows

estshorter:
Due to min max macros in `Windows.h`,  building the following code fails on Windows.
This PR fix the build failure.

```cpp
#define WIN32_LEAN_AND_MEAN
//#define NOMINMAX

#define CPPHTTPLIB_ZLIB_SUPPORT

#include <Windows.h>
#include "httplib.h"

int main(void) {
}
```


yhirose:
@estshorter, thanks for your contribution!

PR #1060, opened 2021-09-28
Split the header only if needed

theShmoo:
We include this project with cmake into ours. On every cmake configuration it splits the header  into its 2 parts. This change invalidstes a lot of targets resulting in unecessary rebuilds.
This pull request changes the split.py to only split if the header is newer than the split files.


yhirose:
@theShmoo, thanks for the improvement!

theShmoo:
Thanks for accepting it so fast!

PR #1067, opened 2021-10-07
Allow to specify server IP address

zhenyolka:
cpp-httplib uses getaddrinfo to get server IP address to connect to. However sometimes this function might not work. For example when staticially linked with musl binary running on system. that don't have /etc/resolv.conf, android for example. Also plain DNS requests may be spoofed by ISP to block access to sites. Some Russian ISPs do that and with implementing by government National System of Domain Names this problem will grow. Also it affects privacy.
One of solution is to allow projects that use this library specify IP address of server when do HTTP/HTTPS request. So we can resolve IP over DoH/DoT or any other secure way and then pass it to cpp-httplib Client class constructor.
This pull request implements this feature.

### DIFF
--- a/test/test.cc
+++ b/test/test.cc
@@ -765,7 +765,7 @@ TEST(SpecifyServerIPAddressTest, RealHostname) {
 
   cli.set_hostname_addr_map({{host, wrong_ip}});
   auto res = cli.Get("/");
-  ASSERT_FALSE(!res);
+  ASSERT_TRUE(!res);
   EXPECT_EQ(Error::Connection, res.error());
 }
 


zhenyolka:
@yhirose take a look. This PR for my issue

yhirose:
@PixlRainbow, thanks for the excellent suggestion!
@zhenyolka, `set_hostname_addr_map(const std::map<std::string, std::string> addr_map)` looks very easy for me to understand what it tries to do instead of adding new constructors with ip. Could you adjust your code accordingly? Thanks a lot!

zhenyolka:
@yhirose done. Could you check?

yhirose:
@zhenyolka, thanks for the excellent contribution!

PR #1073, opened 2021-10-14
support custom ssl ctx configuration for SSLServer

CarlosLeeGit:
In some conditions, users need to set custom options for ssl_ctx to meet security requirements. Therefore, a constructor is added to SSLServer to support users to customize ssl_ctx during SSLServer construction.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -1364,6 +1364,9 @@ public:
   SSLServer(X509 *cert, EVP_PKEY *private_key,
             X509_STORE *client_ca_cert_store = nullptr);
 
+  SSLServer(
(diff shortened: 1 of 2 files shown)


CarlosLeeGit:
@yhirose sorry, i did not find the reviews since you mentioned first. am i missing something?

yhirose:
@CarlosLeeGit, sorry that I didn't submit my review... Now you see them. :)

CarlosLeeGit:
@yhirose i will use cpp-httlib in our project. so i want to known the release plan for next version.  could i be informed. :)

PR #1082, opened 2021-11-02
Get request with only params

hctym1995:
Some methods have been added to support the client to send GET requests with only params

### DIFF
--- a/test/test.cc
+++ b/test/test.cc
@@ -2881,6 +2881,17 @@ TEST_F(ServerTest, GetStreamedChunkedWithGzip) {
   EXPECT_EQ(std::string("123456789"), res->body);
 }
 
+TEST_F(ServerTest, GetMethodWithParams) {
+  Params params;
+  params.emplace("test1", "param1");
+
+  auto res = cli_.Get("/getmethodwithparams", params);
+  ASSERT_TRUE(res);
+  EXPECT_EQ(200, res->status);
+  EXPECT_EQ(std::string("123456789"), res->body);
+}
+
+
 TEST_F(ServerTest, GetStreamedChunkedWithGzip2) {
   Headers headers;
   headers.emplace("Accept-Encoding", "gzip, deflate");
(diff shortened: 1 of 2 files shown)


yhirose:
@hctym1995, thanks for the pull request. Could you add unit test cases for the additions?

hctym1995:
Can you tell me where I should add it in test.cc?

hctym1995:
Or some keywords

yhirose:
@hctym1995, you could add a unit test like `TEST_F(ServerTest, GetMethodWithParams) { ... }`.

hctym1995:
@yhirose  I have already added unit test function, please check if it is correct

test/test.cc
```diff
@@ -2881,6 +2881,17 @@ TEST_F(ServerTest, GetStreamedChunkedWithGzip) {
   EXPECT_EQ(std::string("123456789"), res->body);
 }
 
+TEST_F(ServerTest, GetMethodWithParams) {
+  Params params;
+  params.emplace("test1", "param1");
+
+  auto res = cli_.Get("/getmethodwithparams", params);
```
yhirose, line 2888: I don't see any corresponding `/getmethodwithparams` in `ServerTest::SetUp` where you have to test the given parameters.

You could refer to the following:
https://github.com/yhirose/cpp-httplib/blob/301faa074c4a0fa1dbe470dfb4f77912caa1c57f/test/test.cc#L1448-L1456

test/test.cc
```diff
@@ -2881,6 +2881,17 @@ TEST_F(ServerTest, GetStreamedChunkedWithGzip) {
...
+  ASSERT_TRUE(res);
+  EXPECT_EQ(200, res->status);
+  EXPECT_EQ(std::string("123456789"), res->body);
```
yhirose, line 2891: This doesn't actually test the parameters at all...

yhirose:
@hctym1995, the unit test you added doesn't really test the given parameters. Could you follow my comments in the review?

Supsource (approved):

yhirose:
@hctym1995, I just compiled your changes with `test/test.cc`, and got a number of compile errors. It seems like the changes are not really tested... I appreciated your efforts though, I rejected the change for this reason. Thank you for your understanding.

image

PR #1085, opened 2021-11-03
Do not use `buf.size()` on a uninitialized array.

mgautierfr:
Recent version of gcc raise a warning `maybe-uninitialized` when using
`size()` method on a uninitialized array.
This prevent us to use httplib in a project compiled with `-Werror`.

This fixes the same problem than #980 but in a better way.
I concentrate myself on the place where gcc raised a warning, not everywhere in the code where we could replace `buf.size()` by a constant.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -1490,9 +1490,9 @@ inline ssize_t Stream::write_format(const char *fmt, const Args &...args) {
   std::array<char, bufsiz> buf;
 
 #if defined(_MSC_VER) && _MSC_VER < 1900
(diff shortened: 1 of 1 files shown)


yhirose:
@mgautierfr, thanks for the feedback, but I don't agree with the warning from gcc. It looks like a false negative to me. I'll google it, and get back to you.

yhirose:
@mgautierfr, I guess i fixed this issue in a better way. Could you check with the latest httplib.h on your machine when you have a chance? Thanks!

mgautierfr:
It works on my computer yes. Thanks for the fix.

PR #1092, opened 2021-11-15
build(meson): feature args in pkg-config file

Tachi107:
Follow-up for #1090. The args are now also added to the pkg-config file.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -96,6 +96,7 @@ if get_option('cpp-httplib_compile')
   import('pkgconfig').generate(
     lib,
     description: 'A C++ HTTP/HTTPS server and client library',
+    extra_cflags: args,
     url: 'https://github.com/yhirose/cpp-httplib',
     version: version
   )


yhirose:
@Tachi107, sorry that I just made the pull request #1093 to remove meson-build tests from GitHub Actions. I have been actually thinking of it for quite a while, because these tests don't help me, and costs extra 10 minutes to finish the workflow. The regular tests are enough for me to fully test all the functionalities in cpp-httplib quickly. Also I have been struggled with the failure of 'test/meson-build (windows)' very often, and somethings 'test/meson-build (ubuntu)'... Of course, I leave `meson.build` and related files. Thanks for your understanding!

Tachi107:
@yhirose don't worry, the project is yours after all. Anyway, the Meson tests will be run by the Debian CI, as Meson is used to build the package, so Meson will always be tested downstream :)

Issue #1102, opened 2021-11-23
Remove CMake Warning of "failed to find the latest Git tag"


theShmoo:
If you pull the source code as an archive (e.g.: https://github.com/yhirose/cpp-httplib/archive/refs/tags/v0.9.7.tar.gz) you get a CMake warning at:

```
CMake Warning at cmake-build-debug/_deps/cpphttplib-src/CMakeLists.txt:78 (message):
  cpp-httplib failed to find the latest Git tag, falling back to using user
  agent as the version.
```

This is not pretty in our build pipeline. Could you remove this warning?

yhirose:
@sum01, do you have any thought?

sum01:
If you remove it then people won't know why the version is then missing the patch ver.

I feel there should be some recognition of when this happens, instead of silently doing things in the background. But if you don't think so then remove it.

Could also make it a dev warning and then you'll be free to suppres dev warnings if you want to.

yhirose:
@sum01, thanks for the reply.
@theShmoo, what do you think?

theShmoo:
see the PR

sum01:
Is the goal just to silence the warning, or do you actually want to pass custom versions at build time?

theShmoo:
Both. Silence the warning and pass the full version at build time.

KTGH ***@***.***> schrieb am Mi., 24. Nov. 2021, 23:02:

sum01:
It seems fine then.

I guess you could use a proper variable name though instead of co-opting the old (hidden) var.
Like `HTTPLIB_FORCE_SET_VERSION` or something to that effect (don't forget to document it at the top of the file). Then have it skip all the git & string matching stuff, and just `set(_httplib_version "${HTTPLIB_FORCE_SET_VERSION}")`

theShmoo:
Thanks for the suggestions.
I updated the PR #1104

yhirose:
@sum01, @theShmoo, is it better to have the patch version as well in the user agent string? If do so, we can get the full version number from httplib.h itself and don't need to have the git dependency in CMakeLists.txt any more. Only downside is that I need to make sure of updating the version string even time I release another patch version. But I am ok with it.
https://github.com/yhirose/cpp-httplib/blob/master/httplib.h#L5981

sum01:
I think that would be ideal (short of putting it into the Cmake file directly).

Then it would just always work & be the full version. Up to you if you want to maintain that or not though.

[...]

PR #1104, opened 2021-11-24
if _raw_version_string is defined skip detection

theShmoo:
issue #1102


yhirose:
This is handled with #1108.

PR #1148, opened 2021-12-29
Work around silly Win defines to support BoringSSL

witaly-iwanow:


httplib.h
```diff
@@ -213,6 +213,12 @@ using socket_t = int;
 #include <thread>
 
 #ifdef CPPHTTPLIB_OPENSSL_SUPPORT
+// these are defined in wincrypt.h and it breaks compilation if BoringSSL is used
```
yhirose, line 216: Should it be done only for `_WIN32` since it relates to the WinSock2 `wincrypt.h` header file?

yhirose (commented):
Please see my comment above.

httplib.h
witaly-iwanow: Sure, good call!

yhirose:
@witaly-iwanow, thanks for your contribution!

PR #1171, opened 2022-01-25
Report connection timeout as separate event

gjasny:


yhirose:
@gjasny, thanks for the pull request. Could you take a look at the problem on GitHub Actions workflow?
https://github.com/yhirose/cpp-httplib/actions/runs/1745000655

test/test.cc
```diff
@@ -4302,10 +4307,11 @@ TEST(SSLClientTest, ServerCertificateVerification4) {
   cli.set_connection_timeout(30);
 
   auto res = cli.Get("/test");
-  ASSERT_TRUE(res);
-  ASSERT_EQ(200, res->status);
 
   t.join();
+
```
yhirose, line 4312: @gjasny, I don't think this change is needed. (Also I see the same change a number of times in this file.) Could you reverted these?

yhirose (changes requested):
Could please revert back the changes of moving the following lines in test/test.cc?

```
ASSERT_TRUE(res);
ASSERT_EQ(200, res->status);
```

test/test.cc
gjasny: It is needed that the test does not crash if the assertion is not fulfilled. In contrast to `EXPECT` the `ASSERT` macros return on error. If a `std::thread` is destructed and it is neither detached nor joined it will throw an assertion and the test will break.

gjasny: If you'd like I could move them into a different PR. But they are already different commits.

yhirose: Yes, could you take the unrelated changes to a separated pull request, so that this pull request becomes simpler and clearer?

yhirose:
According to the code added in the unit test, does this new error type occur only with AF_INET? If so, it will actually create more confusion for users using AF_UNSPEC, because they won't receive the ConnectionTimeout even if it's declared in httplib.h...

```cpp
cli.set_address_family(AF_INET);
```

gjasny:
> According to the code added in the unit test, does this new error type occur only with AF_INET? If so, it will actually create more confusion for users using AF_UNSPEC, because they won't receive the ConnectionTimeout even if it's declared in httplib.h...
> 
> ```c++
> cli.set_address_family(AF_INET);
> ```

It took me a while why the test was failing on Linux. What happens is:

`google.com` gets resolved to both: an IPv4 and an IPv6 address. The client code loops over both and for IPv4 sets the error code to "ConnectionTimeout". Then it tries IPv6 but that's not working because the test host only has an IPv4 address. Thus the error code gets overwritten and set to "Connection".

By explicitly requesting an IPv4 lookup the test avoids the second IPv6 loop and the overwriting of the error code.

If the host would have IPv6 connectivity it would also return "ConnectionTimeout" for IPv6.

yhirose:
@gjasny, thanks for the clear explanation. I'll merge it to the master branch soon. Thanks for your contribution!

yhirose (approved):

PR #1173, opened 2022-01-26
Fix #1172

### LINKED ISSUE #1172: Error::Read when attempting to read from an NVR API
Hi,

I'm building an application for pulling data from an API of a NVR we use. During development, I encountered no issues in pulling the data I needed. However, as I am starting to move the code into production, I have found that there are a couple of NVRs that cause the library to throw a Error::Read. I attempted to step through the code as far as possible, but was only able to find that the strm_.read() function during stream_line_reader::getline() was returning 0, and the i variable in the loop was also 0, so the function would therefore return false. I am not sure why this is returning the way this.

I have not been able to replicate this issue via the browser, curl, or Postman, and am therefore confused as to what is going on here. What is odd about this particular issue is that only this type of NVR seems to have the problem; yet if it the NVR was the source of the issue, I would think that I should be able to replicate the problem outside of this library. Do you have any ideas for assisting me in determining if this is a library issue?

yhirose:
This change is based on RFC7230, § 3.5 'Message Parsing Robustness': "Although the line terminator for the start-line and header fields is the sequence CRLF, a recipient MAY recognize a single LF as a line terminator and ignore any preceding CR."

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -3217,17 +3217,26 @@ inline bool read_headers(Stream &strm, Headers &headers) {
     if (!line_reader.getline()) { return false; }
 
     // Check if the line ends with CRLF.
+    auto line_terminator_len = 2;
     if (line_reader.end_with_crlf()) {
       // Blank line indicates end of headers.
       if (line_reader.size() == 2) { break; }
+#ifdef CPPHTTPLIB_ALLOW_LF_AS_LINE_TERMINATOR
+    } else {
+      // Blank line indicates end of headers.
+      if (line_reader.size() == 1) { break; }
+      line_terminator_len = 1;
+    }
+#else
     } else {
       continue; // Skip invalid line.
     }
+#endif
 
     if (line_reader.size() > CPPHTTPLIB_HEADER_MAX_LENGTH) { return false; }
 
-    // Exclude CRLF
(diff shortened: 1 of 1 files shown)


ASBaj:
Works. Has a couple of syntax errors though. Here is the fixed for loop in the read_headers function:

            for (;;) {
                if (!line_reader.getline()) { return false; }

                // Check if the line ends with CRLF.
                auto line_terminator_len = 2;
                if (line_reader.end_with_crlf()) {
                    // Blank line indicates end of headers.
                    if (line_reader.size() == 2) { break; }
                }
#ifdef CPPHTTPLIB_ALLOW_LF_AS_LINE_TERMINATOR
                else {
                    // Blank line indicates end of headers.
                    if (line_reader.size() == 1) { break; }
                    line_terminator_len = 1;
                }
#else
                else {
                    continue; // Skip invalid line.
                }
#endif

Not sure if you want this as a full pull request, or if pasting the function here works. Let me know.

yhirose:
@ASBaj, thanks for the good catch! I corrected it according to your report. Could you verify the latest one in this pull request? If it works, I'll merge it right away.

ASBaj:
Yep, it's working. Thanks for that!

PR #1176, opened 2022-01-27
Fix test fixture crashes

gjasny:


yhirose:
@gjasny, could you explain more about it, because I have never experienced the crashes on my local machines and GitHub Actions workflow. Thanks.

Tachi107:
I think that this PR could fix some flaky behaviour of the test suite. I've actually encountered some test failures of `ServerCertificateVerification4` and `ClientCertPresent` a few times, but they were hard to reproduce, as they failed at random intervals.

I've been able to observe this flaky behaviour especially well in Debian's build network; the cpp-httplib package gets built multiple times, for multiple architectures, and every once in a while the test suite reports an error - as an example, the latest "bookworm" build reported a `ClientCertPresent` error, while the second attempt in a row of the "unstable" build reported an issue within `ServerCertificateVerification4`. ([example](https://tests.reproducible-builds.org/debian/rb-pkg/bookworm/amd64/cpp-httplib.html))

These sporadic errors made me think of some sort of thread race, and in fact this PR moves the asserts after the call to `thread::join()`, so that the main test suite thread doesn't check for results before the spawned thread has finished doing its stuff. This should, in theory, avoid data races and fix sporadic failures.

gjasny:
I saw those crashes when the pem files were not properly generated. Just delete any of those files and the test will crash due to the "unexpected" thread joining.
As you experienced: if everything  is setup as expected the tests pass without problems. It only starts breaking once something is not right. In that case the test itself should survive.

yhirose:
@gjasny, thanks for the update. If it happens only when "the pem files were not properly generated", I don't think this pull request is necessary. So I'll close it for now, but thanks for your contribution.

Tachi107:
@yhirose please look at the changes, it also fixes the race condition I described in my previous comment.

PR #1181, opened 2022-01-28
User-Agent update cpp-httplib/0.10.2

Rockybilly:


yhirose:
@Rockybilly, oops. Thanks for catching the problem!

PR #1182, opened 2022-01-29
build(meson): always install a pkg-config file

Tachi107:
A pkg-config file was previously installed only if cpp-httplib was being built as a compiled library.

Since architecture-independent .pc files can exist in /usr/share/pkgconfig, it can be useful to install one even when installing the header-only version (for example, it could be used by third party projects to easily find out if cpp-httplib is installed and its version, using something like Meson's `dependency()` or CMake's `pkg_check_modules()`).

This is accomplished using the `dataonly` kwarg of Meson's pkg-config generator, introduced in version 0.54.0. This means that I had to bump the minimum required version, but since that has been released in March 2020 and is available in Debian 11 I believe it's fine to require it.

The change makes the Meson build behave a bit more like the CMake one, as it also always installs a CMake Config file, but here the pkg-config file gets installed to the correct architecture-independent directory (`datadir` represents /usr/share on Linux and simiar systems).

Lastly, I made some minor cleanups, partially thanks to the version bump

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -19,12 +19,13 @@ project(
 # Check just in case downstream decides to edit the source
 # and add a project version
 version = meson.project_version()
-python = import('python').find_installation('python3')
+python3 = find_program('python3')
 if version == 'undefined'
   # Meson doesn't have regular expressions, but since it is implemented
   # in python we can be sure we can use it to parse the file manually
   version = run_command(
-      python, '-c', 'import re; raw_version = re.search("User\-Agent.*cpp\-httplib/([0-9]+\.?)+", open("httplib.h").read()).group(0); print(re.search("([0-9]+\\.?)+", raw_version).group(0))'
+      python3, '-c', 'import re; raw_version = re.search("User\-Agent.*cpp\-httplib/([0-9]+\.?)+", open("httplib.h").read()).group(0); print(re.search("([0-9]+\\.?)+", raw_version).group(0))',
+      check: true
   ).stdout().strip()
 endif
 
@@ -68,7 +69,7 @@ if get_option('cpp-httplib_compile')
     'split',
     input: 'httplib.h',
     output: ['httplib.cc', 'httplib.h'],
-    command: [python, files('split.py'), '--out', meson.current_build_dir()],
+    command: [python3, files('split.py'), '--out', meson.current_build_dir()],
     install: true,
     install_dir: [false, get_option('includedir')]
   )
@@ -92,6 +93,14 @@ if get_option('cpp-httplib_compile')
 else
   install_headers('httplib.h')
   cpp_httplib_dep = declare_dependency(compile_args: args, dependencies: deps, include_directories: include_directories('.'))
+
+  import('pkgconfig').generate(
+    name: 'cpp-httplib',
+    description: 'A C++ HTTP/HTTPS server and client library',
+    install_dir: join_paths(get_option('datadir'), 'pkgconfig'),
+    url: 'https://github.com/yhirose/cpp-httplib',
+    version: version
+  )
 endif
 
 if meson.version().version_compare('>=0.54.0')
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -5,5 +5,5 @@
 option('cpp-httplib_openssl', type: 'feature', value: 'auto', description: 'Enable OpenSSL support')
 option('cpp-httplib_zlib',    type: 'feature', value: 'auto', description: 'Enable zlib support')
 option('cpp-httplib_brotli',  type: 'feature', value: 'auto', description: 'Enable Brotli support')
-option('cpp-httplib_compile', type: 'boolean', value: false,  description: 'Split the header into a compilable header & source file (requires Python 3)')
+option('cpp-httplib_compile', type: 'boolean', value: false,  description: 'Split the header into a compilable header & source file')
 option('cpp-httplib_test',    type: 'boolean', value: false,  description: 'Build tests')


eli-schwartz:
dataonly is incorrect, it denotes pkg-config files which are used for storing information not relevant to a C compiler. Headers are not architecture-independent (even if in many cases they are in practice, this does not generalize).

More pertinent, Meson's dataonly kwarg suppresses the generation of `Cflags:` and `includedir=`, which is... kinda important for finding a header-only library.

Tachi107:
I noticed that, and I somehow thought that it wasn't a problem because the header got installed in the default include dir and that is always searched by default, but this is obviously wrong because if you install the header somewhere else and look for the .pc file you would get a positive result when running Meson/CMake but compilation would fail :/

Thanks for pointing this out :)

Tachi107:
Uhm... Not enabling `dataonly` makes the .pc file usable, but the `libdir` variable gets created and set to the host architecture value, even if header-only libraries have little to do with the library dir.

Apart from the fact that this is incorrect in the context of files installed in `/usr/share`, I think this could create issues with reproducibility and in systems like Debian's Multiarch - a pkg-config file of an header only library generated on x86_64 should be bit-by-bit the same as one built on aarch64.

eli-schwartz:
> Apart from the fact that this is incorrect in the context of files installed in `/usr/share`,

Again, references to /usr/include should not as a matter of course be installed in the arch-independent datadir, but in the arch-dependent libdir. This is, in the general case, a feature, not a bug.

> I think this could create issues with reproducibility and in systems like Debian's Multiarch - a pkg-config file of an header only library generated on x86_64 should be bit-by-bit the same as one built on aarch64.

It doesn't create an issue, because of two reasons:
- they will just flag it as not-multiarch
- header-only libraries are a bit of an anti-feature, fortunately since this project makes it optional (that is the entire point of this PR), Debian has the option to configure with `-Dcpp-httplib_compile=true` and benefit from better security tracking and cross-package caching of compiler outputs (a.k.a. compiled libraries)... and indeed Debian does precisely that (you are the Debian maintainer who implemented this)... which means that ***it is not arch-independent, it refers to `/usr/lib/x86_64-linux-gnu/libcpp-httplib.so`***.


> Edit: I may be looking at this issue from the wrong perspective, but header files are the same plain text files regardless of the architecture

Or at least, this project installs header files that happen to be the same plain text files regardless of the architecture.

Again:

> Headers are not architecture-independent (even if in many cases they are in practice, this does not generalize).

So the general meson behavior of `dataonly: false` is correct IMO. That being said, if you would very much like to submit a PR to meson so that it doesn't generate an empty `Libs:` line and doesn't write out `libdir=` if it is not referenced? I guess I would not be strongly opposed.

Tachi107:
> references to /usr/include should not as a matter of course be installed in the arch-independent datadir, but in the arch-dependent libdir. This is, in the general case, a feature, not a bug.

I don't understand why... If the header is the same between all architectures and there's no compiled library, why should I keep my .pc file in $libdir? If I move it to $datadir, I don't have to install the same stuff _n_ times but just once, for all architectures.

> * they will just flag it as not-multiarch

Yeah thinking about this more it is not really an issue for multiarch, only for reproducibility between different build machines.

> * header-only libraries are a bit of an anti-feature, fortunately since this project makes it optional (that is the entire point of this PR), Debian has the option to configure with `-Dcpp-httplib_compile=true` and benefit from better security tracking and cross-package caching of compiler outputs (a.k.a. compiled libraries)... and indeed Debian does precisely that (you are the Debian maintainer who implemented this)... which means that _**it is not arch-independent, it refers to `/usr/lib/x86_64-linux-gnu/libcpp-httplib.so`**_.

I don't love header-only libraries either, but I think that you got a bit confused here. This project is actually both header-only and not header-only. The project is developed and maintained as a header-only file, but it can be splitted in a .h and a .cpp file using a Python script, and the .pc file generated when `cpp-httplib_compile` is `true` is perfectly fine. Coming back to this PR, the pkg-config file we are talking about is the one that gets generated when Meson is configured to not split the header, so when cpp-httplib is header-only. Since it is a .pc file that only refers to an architecture-independent file stored `/usr/include/httplib.h`, it shouldn't contain any reference to the libdir; no references to the libdir means that the file will be identical across architectures, and that means that is should be stored in `/usr/share/pkgconfig`.

> Or at least, this project installs header files that happen to be the same plain text files regardless of the architecture.
> > Headers are not architecture-independent (even if in many cases they are in practice, this does not generalize).

But I don't understand why they aren't. I see it as one header is not always arch-dependent or arch-independent, but well it... depends :) Because if I'm talking about a header file that has to be used in conjunction with a compiled library then I agree that that header is not truly arch-independent, because while its text might be the same between different machines, it can't exist on its own, and is part of an arch-dependent bundle; but if we look at a header only library, then the whole bundle would consist of the same content and would be used the same way between every different architecture, making it truly arch-independent.

In short, not all headers are made equal.

> So the general meson behavior of `dataonly: false` is correct IMO.

Yes, `dataonly` is fine.

> That being said, if you would very much like to submit a PR to meson so that it doesn't generate an empty `Libs:` line and doesn't write out `libdir=` if it is not referenced? I guess I would not be strongly opposed.

I'm looking into it, but I make no promises. My Python skills not that great :)

eli-schwartz:
> Coming back to this PR, the pkg-config file we are talking about is the one that gets generated when Meson is configured to not split the header

Yeah, I just mean: Debian is not affected either way. :)


> In short, not all headers are made equal.

Indeed they aren't. Some of them are generated at build time and tuned to the CPU architecture they are being installed on. Some are part of a split source tree that has per-architecture headers. Most are boring, and the same on all architectures.

Again, it's not a property of "being a header" that makes it architecture-independent... it is a property of the project choices when designing the header. Actually taking advantage of the ability to be arch-dependent, is probably fairly niche.

This probably doesn't matter for cpp-httplib, but it *is* something that changes to meson's upstream pkgconfig module will need to take into consideration. Anyway, we can discuss that on the Meson issue tracker.

> I'm looking into it, but I make no promises. My Python skills not that great :)

I am available to provide guidance in the #mesonbuild IRC support channel. :)

Tachi107:
Hey @yhirose, sorry for the noise here. This PR is good as is, and you can merge it whenever you like. The issues me and Eli were talking about will be fixed in Meson itself

yhirose:
@Tachi107, thanks for the update. I'll merge it soon.

PR #1183, opened 2022-01-29
test: add missing "_Online" suffix

Tachi107:
This test fails reproducibly in a Debian build chroot, and they generally don't have internet access


yhirose:
Thanks!

PR #1187, opened 2022-02-06
Fix typo in Dockerfile

hyperupcall:
Like `WORKDIR`, `RUN`, `COPY`, etc. `FROM` should be capitalized ^w^


yhirose:
@hyperupcall, thank you for your contribution!

PR #1201, opened 2022-02-22
CPPHTTPLIB_NO_DEFAULT_USER_AGENT skips default user agent

au-ee:
For bandwidth critical environments, it can be desirable to send no User-Agent header.
This modification causes cpp-httplib not to write the default User-Agent if CPPHTTPLIB_NO_DEFAULT_USER_AGENT is defined.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -6153,9 +6153,11 @@ inline bool ClientImpl::write_request(Stream &strm, Request &req,
 
   if (!req.has_header("Accept")) { req.headers.emplace("Accept", "*/*"); }
 
+#ifndef CPPHTTPLIB_NO_DEFAULT_USER_AGENT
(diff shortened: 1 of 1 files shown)


yhirose:
@au-ee, thank you for the feed. Could you explain your motivation with the pull request?
https://stackoverflow.com/questions/24274789/is-lack-of-user-agent-in-http-request-valid

au-ee:
Sure. Here is our use case: We work in an environment where lots of iot devices running on low volume mobile network plans regularly contact our web servers. If we do not transfer the User-Agent header, this saves us hundreds of megabytes of mobile network traffic per month in our global mobile network traffic pool.

yhirose:
@au-ee, thanks for the explanation. I just merged it. Thanks for your contribution!

PR #1205, opened 2022-02-23
Add optional private key password to SSLServer ctor

sebastienblanchet:
This parameter bypasses the prompt at runtime when using an encrypted private key:
```
Enter PEM pass phrase:
```

Let me know if you think it might be worth adding this and I can add required unit tests.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -1384,7 +1384,8 @@ class SSLServer : public Server {
 public:
   SSLServer(const char *cert_path, const char *private_key_path,
             const char *client_ca_cert_file_path = nullptr,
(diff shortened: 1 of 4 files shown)


yhirose:
@sebastienblanchet, could you add a unit test in `test/test.cc`? Then, I'll take a look at it. Thanks!

sebastienblanchet:
@yhirose, awesome thanks, `test/test.cc` added!

Can you trigger your CI job to see if passes against your full test suite ?

yhirose:
@sebastienblanchet, looks good to me. Thanks for your fine contribution!

PR #1209, opened 2022-03-03
Add soversion

abouvier:
Add version and soversion when compiling the shared library.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -172,6 +172,11 @@ if(HTTPLIB_COMPILE)
 			$<BUILD_INTERFACE:${_httplib_build_includedir}/httplib.h>
 			$<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}/httplib.h>
 	)
+	set_target_properties(${PROJECT_NAME}
+		PROPERTIES
+			VERSION ${${PROJECT_NAME}_VERSION}
+			SOVERSION ${${PROJECT_NAME}_VERSION_MAJOR}
+	)
 else()
 	# This is for header-only.
 	set(_INTERFACE_OR_PUBLIC INTERFACE)


yhirose:
@sum01, could you take a look at the change? If you think it's ok, I'll merge it. Thanks!

sum01:
@abouvier Seems good, but if the build is static does this error out? I haven't tested personally, so I don't know how Cmake reacts to that.

yhirose:
@abouvier, could you respond to the @sum01's comment, since I am not familiar to the CMakeLists.txt file?

abouvier:
There is no errors because these properties are ignored when building a static library.

sum01:
If so, then it seems fine @yhirose

yhirose:
Just merged it. Thanks for the fine contribution!

Tachi107:
The Meson build currently does the same thing (i.e. setting the soversion to the major version number), but I'm not sure it is the right thing to do. The soname should be the same across releases only if the ABI is kept backwards compatible, but as this is (mainly) a header-only library I'm not sure this is guaranteed. Should it instead be set to major.minor?

yhirose:
@abouvier, could you respond to @Tachi107's comment?

abouvier:
No, I can't :p
You're the one who knows if the shared library is ABI compatible between minor versions.

Tachi107:
What I can say is that since I started packaging cpp-httplib for Debian (version 0.9.7) ABI compat has never been broken. But in my opinion the soname should be changed in a predictable manner, and not simply bumped when breakages occur (either way is fine, but the former is easier to manage).

@yhirose, if you don't want to bother with ever providing a stable ABI we could set the soversion to major.minor.patch, so that you can break ABI anytime (even in patch releases). Not ideal for distros, but as you mainly provide this library as header-only I would completely understand your position :)

Tachi107:
As the latest release broke API and ABI compatibily, I believe it is appropriate to set the SONAME to version_major.version_minor.

I don't see anywhere in the documentation promises about API or ABI stability, but looking at the release history I think that major.minor is accurate enough.

I'll submit a PR to update both the Meson and the CMake build systems as soon as possible.

Issue #1203, opened 2022-02-23
Avoid having request headers in a separate package


au-ee:
Hi @yhirose ,

thanks for this awesome package and thanks for the quick merge, yesterday!

While evaluating cpp-httplib for use in our application, we did a bit of network tracing and discovered that the header of the http request is always sent as a separate package due to the `write_data` in line 6227. This causes a small tcp frame to be sent immediately with the PSH flag set.

The result is that each http request gets split up into two tcp frames at minimum. Which causes a bit of overhead which is usually not critical, but again, can sum up to gigabytes of cellular data volume per month in an iot application.

You know the library and its surroundings better: Do you see any chance to combine sending the header together with the body without having to copy too much data around or do you see a chance to make the stream/socket wait for more data before it is flushed to the network?

Thanks,

Markus

PixlRainbow:
It's been a while since I dug in depth into the implementation of httplib, but as far as I can remember, httplib depends on the system's socket library implementation to handle buffering, and does not by default flush the buffer until the content body is reached.
If the headers are being sent as a separate frame, it suggests that your system's network buffer is disabled, either due to nodelay option or other reasons.
It may still be possible to rewrite httplib to use the existing `BufferStream` instead of depending on the system network buffer, but I am not sure.

yhirose:
@au-ee, as @PixlRainbow mentioned, cpp-httplib relies on the system's socket library implementation. If you would like to control data buffering, you can make your own stream calls by inheriting from `Stream` interface and replace `SocketStream` with yours. Hope it helps!

https://github.com/yhirose/cpp-httplib/blob/d73395e1dc652465fa9524266cd26ad57365491f/httplib.h#L4651-L4660

au-ee:
@PixlRainbow @yhirose - Thanks for your thoughts!

Issue #1144, opened 2021-12-25
How to restrict only one process listening on the given port?


mcordova1967:
Hello, I observed that multiple copies of my program can attach to port 443, only one will receive requests, but the svr.listen() method is always returning true.

If I use the socket option flag svr.listen("0.0.0.0", 443, SO_EXCLUSIVEADDRUSE) the listen() always returns false and no process can listen to the port, not even the first one.

Is there a portable (windows/linux) way of setting a flag so that only one program listens to the given port at the same time?

Thank you,
Martin

yhirose:
@mcordova1967, the current cpp-httplib behavior is based on the following code.

https://github.com/yhirose/cpp-httplib/blob/b324921c1aeff2976544128e4bb2a0979a4aa595/httplib.h#L1531-L1547

You can completely change the socket option setting with the `set_socket_options `.
https://github.com/yhirose/cpp-httplib/blob/b324921c1aeff2976544128e4bb2a0979a4aa595/httplib.h#L652

As you may know the actual effect of each option isn't consistent among operating systems unfortunately. So I am not in good position to answer your questions...

Hope is helps.

PR #1210, opened 2022-03-03
Feature/tcp cork switch on linux

au-ee:
We added the following for the client: if CPPHTTPLIB_LINUX_TCP_CORK is defined on Linux, we add the TCP_CORK flag to the socket. 
This is useful in bandwidth critical situations as the http header can be combined with the first chunk of payload in one tcp frame (see issue https://github.com/yhirose/cpp-httplib/issues/1203).

The TCP_CORK flag causes the tcp frame to be sent when the packet reaches the maximum length. Otherwise the packet will be sent after a timeout of 200ms. For this reason we have not added a manual mechanism to send the packet after the payload is finished.

We wanted to contribute this back to you and hope you like it.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -6153,9 +6153,11 @@ inline bool ClientImpl::write_request(Stream &strm, Request &req,
 
   if (!req.has_header("Accept")) { req.headers.emplace("Accept", "*/*"); }
 
+#ifndef CPPHTTPLIB_NO_DEFAULT_USER_AGENT
   if (!req.has_header("User-Agent")) {
     req.headers.emplace("User-Agent", "cpp-httplib/0.10.2");
   }
+#endif
 
   if (req.body.empty()) {
     if (req.content_provider_) {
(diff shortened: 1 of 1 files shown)


yhirose:
@au-ee, thanks for the pull request. Since TCP_CORK is only available in Linux, I would not prefer to including it in httplib.h.
Is it enough for you to use `set_socket_option`?

Here is an example for server:

```cpp
svr.set_socket_options([] (auto sock) {
    int state = 1;
    setsockopt(sock, IPPROTO_TCP, TCP_CORK, reinterpret_cast<char *>(&state), sizeof(state));
});
```

You can do the same for client as well.

https://github.com/yhirose/cpp-httplib/issues/1144#issuecomment-1001079769

au-ee:
Hi @yhirose , thanks for the hint. We understand that you want to keep the library as platform independent as possible. 
This works for us. I will now close the merge request.

PR #1218, opened 2022-03-18
fix: update user agent

Tachi107:


yhirose:
Oops... Thanks for catching it.

PR #1228, opened 2022-03-24
Give DataSink a "sendfile" method which  performs zero-copy on Linux

ArnaudBienner:
On Linux, this performs zero-copy to send a file from disk to socket.
On other platforms, this is simply a helper function to send a file.

More information about sendfile:
* manpage: https://man7.org/linux/man-pages/man2/sendfile.2.html
* Nice article: https://medium.com/swlh/linux-zero-copy-using-sendfile-75d2eb56b39b
From the benchmark I did on my side, I noticed the same improvements as the one described in this article

I wish I could implement this for the static file server too, since this is where I assume this will be used the most, but this is more complicated to do.

But having it in content provider sounds like a good first step.

If the change look good to you, I will update the README.md to add examples on how to use this (based on the unit tests).

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -173,7 +173,9 @@ using socket_t = SOCKET;
 #include <netdb.h>
 #include <netinet/in.h>
 #ifdef __linux__
+#include <fcntl.h>
 #include <resolv.h>
+#include <sys/sendfile.h>
 #endif
 #include <netinet/tcp.h>
 #ifdef CPPHTTPLIB_USE_POLL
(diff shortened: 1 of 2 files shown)


mcordova1967:
A very nice improvement, I also think it would be nice to have it implemented in the static file server. For Windows I think this function does the equivalent: 
https://docs.microsoft.com/en-us/windows/win32/api/mswsock/nf-mswsock-transmitfile

For MingW-64 (g++ v11.2) it is declared in winsock.h

ArnaudBienner:
Not sure why CI failed :( I tested locally on Ubuntu LTS (with gcc and clang) and macOS and it works there.
Will try to install Ubuntu 21.10 to give it a try to reproduce when I have a minute.

ArnaudBienner:
I realized github actions' "ubuntu-latest" is the LTS version: the one I was using.
Also, the test is not failing in the `sendfile` but on the client `Get`.

I still have no success reproducing this locally, as it always works for me using both g++ or clang++

I'm going to leave this open for now, in case someone has an idea, but unfortunately for now I'm not sure what else I can do :(

yhirose:
@ArnaudBienner, thanks for the pull request. I'll leave it here for a while.

yhirose:
I close it for now.

PR #1242, opened 2022-04-11
fix BindIPAddress error

### LINKED ISSUE #1235: BindIPAddress when force address family to IPV6 and set outbound interface
```
//
// Created by Kotarou on 2022/4/6.
//
#include <httplib.h>

std::unique_ptr<httplib::Client>
get_http_client(std::string_view url, int family = AF_UNSPEC, const char *nif_name = nullptr) {
    auto client = std::make_unique<httplib::Client>(url.data());

    // set outbound interface
    if (nif_name != nullptr) {
        client->set_interface(nif_name);
    }

    // set address family
    client->set_address_family(family);
    client->set_connection_timeout(5);
    client->set_read_timeout(5, 1000);
    client->set_follow_location(true);
    client->set_default_headers({{"User-Agent", "Mozilla"}});

    return client;
}

void print_response(const httplib::Result &result) {
    if (result) {
        std::cout << result->body << std::endl;
    } else {
        std::cerr << "Error: " << httplib::to_string(result.error()) << std::endl;
    }
}

int main(int argc, char *argv[]) {
    std::cout << "AF_UNSPEC, no nif" << std::endl;
    auto client = get_http_client("https://api.ip.sb:443");
    auto response = client->Get("/ip");
    print_response(response);

    std::cout << "AF_UNSPEC, en0" << std::endl;
    client = get_http_client("https://api.ip.sb:443", AF_UNSPEC, "en0");
    response = client->Get("/ip");
    print_response(response);

    std::cout << "AF_INET, no nif" << std::endl;
    client = get_http_client("https://api.ip.sb:443", AF_INET);
    response = client->Get("/ip");
    print_response(response);

    std::cout << "AF_INET, en0" << std::endl;
    client = get_http_client("https://api.ip.sb:443", AF_INET, "en0");
    response = client->Get("/ip");
    print_response(response);

    std::cout << "AF_INET6, no nif" << std::endl;
    client = get_http_client("https://api.ip.sb:443", AF_INET6);
    response = client->Get("/ip");
    print_response(response);

    std::cout << "AF_INET6, en0" << std::endl;
    client = get_http_client("https://api.ip.sb:443", AF_INET6, "en0");
    response = client->Get("/ip");
    print_response(response);
    
    return 0;
}
```

Output:
```
AF_UNSPEC, no nif
2405:aacf:acdc:6dc:33:cda2:153e:745

AF_UNSPEC, en0
38.13.192.76

AF_INET, no nif
38.13.192.76

AF_INET, en0
38.13.192.76

AF_INET6, no nif
2405:aacf:acdc:6dc:33:cda2:153e:745

AF_INET6, en0
Error: BindIPAddress
```

not sure what cause this, but when set address family to AF_INET6 and also force to use en0 as outbound interface, library returned a bind ip error.

CyberKoo:
fix AF_UNSPEC cause BindIPAddress
fix #1235

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -2653,9 +2653,10 @@ inline bool bind_ip_address(socket_t sock, const char *host) {
 inline std::string if2ip(int address_family, const std::string &ifn) {
   struct ifaddrs *ifap;
   getifaddrs(&ifap);
+  std::string addr_candidate;
   for (auto ifa = ifap; ifa; ifa = ifa->ifa_next) {
-    if (ifa->ifa_addr && ifa->ifa_addr->sa_family == address_family &&
-        ifn == ifa->ifa_name) {
+    if (ifa->ifa_addr && ifn == ifa->ifa_name && (AF_UNSPEC == address_family ||
+        ifa->ifa_addr->sa_family == address_family)) {
       if (ifa->ifa_addr->sa_family == AF_INET) {
         auto sa = reinterpret_cast<struct sockaddr_in *>(ifa->ifa_addr);
         char buf[INET_ADDRSTRLEN];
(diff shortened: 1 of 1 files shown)


httplib.h
```diff
@@ -2665,10 +2665,14 @@ inline std::string if2ip(int address_family, const std::string &ifn) {
         }
       } else if (ifa->ifa_addr->sa_family == AF_INET6) {
         auto sa = reinterpret_cast<struct sockaddr_in6 *>(ifa->ifa_addr);
-        char buf[INET6_ADDRSTRLEN];
-        if (inet_ntop(AF_INET6, &sa->sin6_addr, buf, INET6_ADDRSTRLEN)) {
-          freeifaddrs(ifap);
-          return std::string(buf, INET6_ADDRSTRLEN);
+        auto is_global_addr = ((((const uint32_t *) (&sa->sin6_addr))[0]
+                                & htonl(0x70000000)) == htonl (0x20000000));
```
yhirose, line 2669: @CyberKoo, could you explain what `is_global_addr` tries to do?

CyberKoo: this was taken from
https://github.com/pupnp/pupnp/blob/c61460be44645f5c9040c1cf4cf7ac5efbe6b01d/upnp/src/api/upnpapi.cpp#L96
to exclude all reserved IPv6 addresses, more detail
https://en.wikipedia.org/wiki/Reserved_IP_addresses#IPv6
if return local link address (fe80::/64) will cause connection error, since the local link address is only designed for used within LAN.

CyberKoo: ULA is normally used within a private network, which means this address is normally not accessible from internet (not routable). 
In some special case (NAT66 used), the ULA address will be translated on the router site, so the address become useable.
How about we return ULA if only if there is GUA, otherwise return GUA. 
Does that sound good to you?

yhirose: Do you mean "How about we return ULA if only if there is **not** GUA, otherwise return GUA."?

CyberKoo: yes

yhirose: Your suggestion sounds good to me.

PR #1243, opened 2022-04-12
Fix #1235

### LINKED ISSUE #1235: BindIPAddress when force address family to IPV6 and set outbound interface

yhirose:


yhirose:
@CyberKoo could you do the final test with your test code? If it works, I'll merge it and bump the version. Thanks!

yhirose:
Somehow, `SSLClientTest.SetInterfaceWithINET6` test fails...

CyberKoo:
it works on mac, but not on linux, I'll post more deteil later.

CyberKoo:
When using SLAAC to obtain IPv6 address will result in the interface having more than 1 address buy only one is useable and others are deprecated.
fig1

When I run the test program, the IP address returned from if2ip is the first IP, which is deprecated and causes a connection error. You can see from the example, the preferred ipv6 address should end with 15a3, but our if2ip returns 5ffd which corresponding to the first ipv6 address.

fig2

To address this problem, the deprecated address also needs to be excluded. But there is no POSIX way to do this, this has to be done in platform-dependent way.

For Linux use the constant IFA_F_DEPRECATED defined in rtnetlink 
https://man7.org/linux/man-pages/man7/rtnetlink.7.html

For *BSD (MacOS) use macro IFA6_IS_DEPRECATED defined in in6.h (Line 448)
https://opensource.apple.com/source/xnu/xnu-1699.24.8/bsd/netinet6/in6.h

CyberKoo:
By the way, not sure why, but after removing the if_nametoindex assignment my test code seems working now, even with a deprecated IPv6 address.

yhirose:
> Sorry, this is incorrect should be
> 
> ```
> return !addr_candidate.empty() ? addr_candidate : std::string();
> ```
> 
> or just simply return addr_candidate;

I made the change.

> By the way, not sure why, but after removing the if_nametoindex assignment my test code seems working now, even with a > deprecated IPv6 address.

What does it mean? Don't we need the following code any more?

image

CyberKoo:
> > Sorry, this is incorrect should be
> > ```
> > return !addr_candidate.empty() ? addr_candidate : std::string();
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > or just simply return addr_candidate;
> 
> I made the change.
> 
> > By the way, not sure why, but after removing the if_nametoindex assignment my test code seems working now, even with a > deprecated IPv6 address.
> 
> What does it mean? Don't we need the following code any more?
> 
> image
> 
> Please send another pull request if you find any things to be adjusted. Thanks.

Yes, maybe remove it just for now, I'll do more research about it.

yhirose:
> Yes, maybe remove it just for now, I'll do more research about it.
I removed it.

PR #1253, opened 2022-04-19
meson: fix regression that broke extracting version

eli-schwartz:
In commit 33f67386fec8040a36a26b2038b39dd8dcf2814d the code that heuristically parsed the version broke due to the version being moved around into a more easily accessible define.

Also fix using this as a subproject since commit 8ecdb1197967dea050fd38a8e9b5020e02320b31.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -21,12 +21,11 @@ project(
 version = meson.project_version()
 python3 = find_program('python3')
 if version == 'undefined'
-  # Meson doesn't have regular expressions, but since it is implemented
-  # in python we can be sure we can use it to parse the file manually
-  version3 = run_command(
-      python3, '-c', 'import re, sys; raw_version = re.search("#define CPPHTTPLIB_VERSION \"([0-9]+\.?)+\"", open(sys.argv[1]).read()).group(0); print(re.search("([0-9]+\\.?)+", raw_version).group(0))', files('httplib.h'),
-      check: true
-  ).stdout().strip()
+  cxx = meson.get_compiler('cpp')
+  version = cxx.get_define('CPPHTTPLIB_VERSION',
+    prefix: '#include <httplib.h>',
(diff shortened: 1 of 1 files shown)


eli-schwartz:
Incidentally, if it's fine to bump the meson minimum required version to 0.57 (I know you don't really care about build systems much anyway) you could use the version in `project(..., version: files('VERSION.txt'))`, and just maintain VERSION.txt as a text file with the project version in it. It's not so bad to keep version information in two places, one of them a fully technology-agnostic location.

Tachi107:
Thanks, this check is much better than the previous one

meson.build
```diff
@@ -21,12 +21,11 @@ project(
...
 python3 = find_program('python3')
 if version == 'undefined'
-  # Meson doesn't have regular expressions, but since it is implemented
-  # in python we can be sure we can use it to parse the file manually
-  version = run_command(
-      python3, '-c', 'import re; raw_version = re.search("User\-Agent.*cpp\-httplib/([0-9]+\.?)+", open("httplib.h").read()).group(0); print(re.search("([0-9]+\\.?)+", raw_version).group(0))',
-      check: true
-  ).stdout().strip()
+  cxx = meson.get_compiler('cpp')
+  version = cxx.get_define('CPPHTTPLIB_VERSION',
```
Tachi107, line 25: `cxx.get_define()` now prints `Fetching value of define "CPPHTTPLIB_VERSION" : "0.10.6"`, so you could also remove the `message('cpp-httplib version ' + version)` at line 31

meson.build
```diff
@@ -21,12 +21,11 @@ project(
 version = meson.project_version()
 python3 = find_program('python3')
```
Tachi107, line 22: As Python isn't used anymore here the call to `find_program('python3')` could be moved in the `if get_option('cpp-httplib_compile')` section

PR #1254, opened 2022-04-19
Windows Crypto libs linking

mylogin:


yhirose:
@mylogin, the changes in Makefile are ok, but I don't think the changes in httplib.h are not necessary. Please let me know why you modified httplib.h.

mylogin:
if CPPHTTPLIB_OPENSSL_SUPPORT is not defined, then it is not necessary to link these libraries (crypt32.lib, cryptui.lib)

yhirose:
@mylogin, thanks for the answer. Another thing that I notice is that `LIBS` includes not only liner options, but also compiler related options.

mylogin:
git revert performed or better to do a reset last commits?

yhirose:
@mylogin, you are right. Actually reset is better for keeping the git history clean. Thanks!

yhirose:

PR #1256, opened 2022-04-20
build(meson): allow using OpenSSL 3.0

Tachi107:
Following 0857eba17b9d3ef90d45950f9853b7579d6a7f29 cpp-httplib is fully compatible with OpenSSL versions newer than 1.1.1


yhirose:
Thanks!

PR #1261, opened 2022-04-26
Accept large data transfer over SSL

yosh-matsuda:
I found that the HTTP client cannot send the large request body when SSL is enabled.

In `SSL_write` and `SSL_read` the `size_t` type is cast to `int` hence the threshold is 2GiB.
This pull request fixes this problem by replacing them with `SSL_*_ex` functions and accepts large data transfers over SSL.

The changes are as follows:

- [x] Replace `SSL_write` with `SSL_write_ex`
- [x] Replace `SSL_read` with `SSL_read_ex`
- [x] Add a test for large data transfer over SSL

@yhirose Could you review this PR?

### DIFF
--- a/test/test.cc
+++ b/test/test.cc
@@ -4661,7 +4661,8 @@ TEST(SSLClientServerTest, CustomizeServerSSLCtx) {
   t.join();
 }
 
-TEST(SSLClientServerTest, LargeDataTransfer) {
+// Disabled due to the out-of-memory problem on GitHub Actions Workflows
+TEST(SSLClientServerTest, DISABLED_LargeDataTransfer) {
 
   // prepare large data
   std::random_device seed_gen;
--- a/test/test.cc
+++ b/test/test.cc
@@ -4676,8 +4676,7 @@ TEST(SSLClientServerTest, LargeDataTransfer) {
 
   svr.Post("/binary", [&](const Request &req, Response &res) {
     EXPECT_EQ(large_size_byte, req.body.size());
-    EXPECT_TRUE(std::memcmp(binary.data(), req.body.data(), large_size_byte) ==
-                0);
+    EXPECT_EQ(0, std::memcmp(binary.data(), req.body.data(), large_size_byte));
     res.set_content(req.body, "application/octet-stream");
   });
 
@@ -4697,8 +4696,7 @@ TEST(SSLClientServerTest, LargeDataTransfer) {
   // compare
   EXPECT_EQ(200, res->status);
   EXPECT_EQ(large_size_byte, res->body.size());
-  EXPECT_TRUE(std::memcmp(binary.data(), res->body.data(), large_size_byte) ==
-              0);
+  EXPECT_EQ(0, std::memcmp(binary.data(), res->body.data(), large_size_byte));
 
   // cleanup
   svr.stop();
--- a/test/test.cc
+++ b/test/test.cc
@@ -4660,6 +4660,48 @@ TEST(SSLClientServerTest, CustomizeServerSSLCtx) {
 
   t.join();
 }
+
+TEST(SSLClientServerTest, LargeDataTransfer) {
+
+  // prepare large data
+  std::random_device seed_gen;
+  std::mt19937 random(seed_gen());
+  constexpr auto large_size_byte = 2147483648UL + 1048576UL; // 2GiB + 1MiB
+  std::vector<std::uint32_t> binary(large_size_byte / sizeof(std::uint32_t));
+  std::generate(binary.begin(), binary.end(), [&random]() { return random(); });
+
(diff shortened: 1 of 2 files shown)


test/test.cc
```diff
@@ -4660,6 +4660,51 @@ TEST(SSLClientServerTest, CustomizeServerSSLCtx) {
...
+  std::generate(binary.begin(), binary.end(), [&random]() { return random(); });
+
+  // server
+  SSLServer svr(SERVER_CERT_FILE, SERVER_PRIVATE_KEY_FILE);
+  ASSERT_TRUE(svr.is_valid());
+
+  svr.Post("/binary", [&](const Request &req, Response &res) {
+    EXPECT_EQ(large_size_byte, req.body.size());
+    EXPECT_TRUE(std::memcmp(binary.data(), req.body.data(), large_size_byte) ==
+                0);
```
Tachi107, line 4680: Why not `EXPECT_EQ(std::memcmp(binary.data(), req.body.data(), large_size_byte), 0)`?

test/test.cc
```diff
@@ -4660,6 +4660,51 @@ TEST(SSLClientServerTest, CustomizeServerSSLCtx) {
...
+  cli.enable_server_certificate_verification(false);
+  cli.set_read_timeout(std::chrono::seconds(100));
+  cli.set_write_timeout(std::chrono::seconds(100));
+  auto res = cli.Post("/binary", reinterpret_cast<char *>(binary.data()),
+                      large_size_byte, "application/octet-stream");
+
+  // compare
+  EXPECT_EQ(200, res->status);
+  EXPECT_EQ(large_size_byte, res->body.size());
+  EXPECT_TRUE(std::memcmp(binary.data(), res->body.data(), large_size_byte) ==
```
Tachi107, line 4700: Same as above

yhirose:
@yosh-matsuda, thank you for the pull request, but Github Actions Workflows failed at `ServerTest.GetStreamedEndless` on `macOS-latest` and `ubuntu-latest` with this change as you can see the results of the above checks. Could you take a look at the issue and fix it? Then, I'll start reviewing. Thanks a lot!

yosh-matsuda:
@yhirose The fix is committed including refactoring. Some paths had different return values than previous implementations.   
The added test `SSLClientServerTest.LargeDataTransfer` seems to have been killed probably due to huge memory usage. I confirmed that the killed test passed on the local machine.

yhirose:
@yosh-matsuda, thanks for fixing the problem that I reported. If `SSLClientServerTest.LargeDataTransfer` causes the out-of-memory problem on GitHub Actions Workflows, could you simply comment it out and add some note please?

yhirose:
@yosh-matsuda, thanks for fixing the problem that I reported. If `SSLClientServerTest.LargeDataTransfer` causes the out-of-memory problem on GitHub Actions Workflows, could you simply comment it out and add some note please?

Tachi107:
Instead of commenting it out, its name can be prefixed with `DISABLED_` so that GoogleTest knows that it shouldn't run it. This permits users who can or want run this particular test to run it without editing the source code. For more info, take a look at their docs: https://google.github.io/googletest/advanced.html#temporarily-disabling-tests

yosh-matsuda:
I have disabled the test and all checks are passed.

yhirose:
Everything now looks good to me. Thanks for your find contribution!

yosh-matsuda:
@yhirose @Tachi107 
Thanks to the review and merging!

artpaul:
I found that the commit broke compatibility with boringssl because it does not have SSL_write_ex / SSL_read_ex.
Any suggestions how to fix it?

yhirose:
@yosh-matsuda, any response for @artpaul's comment?

yhirose:
@yosh-matsuda, thanks for the comment.
@artpaul, please see the above comment from @yosh-matsuda. I am not planning to support any other SSL libraries since I am satisfied with OpenSSL and don't have time to support others. If you can do, could you make necessary changes for boringssl and send a pull request? Thanks!

artpaul:
@yhirose Imho, the main cause of the issue was the fact that the client code can handle partial read but cannot handle partial write. So, I have restored the implementation of SSLSocketStream::read and implemented SSLSocketStream::write with the loop:
https://github.com/yhirose/cpp-httplib/pull/1312

The test SSLClientServerTest::LargeDataTransfer has completed successfully on my desktop with Ubuntu.

yhirose:
@yosh-matsuda, do you have any response on the above comment and the pull request (#1312)?
Thanks for your attention.

yosh-matsuda:
@yhirose @artpaul Thanks for the PR. I used the ex functions without deep consideration, but your changes look better to me.

PR #1267, opened 2022-04-30
build(meson): don't require python3

Tachi107:
Thanks to abf3a67dd070e138c0f1a20b913abf003193cb79 the use of python3 isn't required anymore to configure the build, so I moved the `find_program('python3')` inside the "if compile" block.

This makes it possible to configure cpp-httplib on systems where python isn't available with tools like [muon](https://sr.ht/~lattis/muon/)


yhirose:
Thanks!

PR #1272, opened 2022-05-14
Fix Compiling with Clang for CMake

JonathanCline:
I noticed that compiling the library with clang and cmake simply failed due to clang not providing the C++ compile features whatsoever at least with the version of clang I am using. I threw together a quick fix described below:

Added a CMake macro to determine the name of the C++ compiler executable. This is then used to determine if clang (clang++) is being used.

If clang is being used then the target compile features call is skipped.

The justification is that if clang is being used then
using `target_compile_features` would fail as clang doesn't provide this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -187,20 +187,35 @@ endif()
 # Only useful if building in-tree, versus using it from an installation.
 add_library(${PROJECT_NAME}::${PROJECT_NAME} ALIAS ${PROJECT_NAME})
 
-# Might be missing some, but this list is somewhat comprehensive
-target_compile_features(${PROJECT_NAME} ${_INTERFACE_OR_PUBLIC}
-	cxx_std_11
-	cxx_nullptr
-	cxx_lambdas
-	cxx_override
-	cxx_defaulted_functions
-	cxx_attribute_deprecated
-	cxx_auto_type
-	cxx_decltype
-	cxx_deleted_functions
-	cxx_range_for
-	cxx_sizeof_member
-)
+# Determines the C++ compiler executable name
+macro (DETERMINE_CXX_COMPILER_EXECUTABLE_NAME out_Name)
+	set(_outStr )
+	string(REGEX MATCH "[^/\\]*$" _outStr ${CMAKE_CXX_COMPILER})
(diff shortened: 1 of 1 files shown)


JonathanCline:
Might be a reasonably idea to throw in a check to see if `CMAKE_CXX_COMPILE_FEATURES` is not empty so that if/when clang has this available it will automatically be used.

yhirose:
@sum01, could you review this pull request, since I don't have enough knowledge about CMake. Thanks!

CMakeLists.txt
```diff
@@ -187,20 +187,35 @@ endif()
...
-	cxx_range_for
-	cxx_sizeof_member
-)
+# Determines the C++ compiler executable name
+macro (DETERMINE_CXX_COMPILER_EXECUTABLE_NAME out_Name)
+	set(_outStr )
+	string(REGEX MATCH "[^/\\]*$" _outStr ${CMAKE_CXX_COMPILER})
+	string(REGEX MATCH "[^\.]+" _outStr ${_outStr})
+	set(${out_Name} ${_outStr})
+endmacro()
```
Tachi107, line 196: Why is this needed? Can't you just do

```cmake
if (NOT CMAKE_CXX_COMPILER MATCHES "clang_regex")
    target_compile_features(...)
endif()
```

sum01:
Yeah I wasn't aware it was not recommended. It could instead be edited so only `cxx_std_11` is used, as mentioned in the docs.

Regardless, @Tachi107 is right that trying to detect Clang and disable feature settings if it's used isn't a good idea.

Also, I don't have issues using Clang++ on v13.0.1 either, for both header-only and compiled versions of the build. Not sure how @JonathanCline was calling the build function, but using this works just fine on my end.
```cmake
cd build
# Adding -DHTTPLIB_COMPILE=ON also runs as expected.
cmake -DCMAKE_CXX_COMPILER="clang++" ..
cmake --build .
```

If needed I can submit a small patch to delete those few unnecessary feature lines, or anyone else can feel free to as well.

Tachi107:
@JonathanCline this was closed as I believe that your issues was caused by the low-level compile features. Could you please check if your issue has been actually solved?

PR #1274, opened 2022-05-16
build(cmake): minor tweaks

Tachi107:
- Enable [THREADS_PREFER_PTHREAD_FLAG](https://cmake.org/cmake/help/latest/module/FindThreads.html#variable:THREADS_PREFER_PTHREAD_FLAG) to use -pthread where supported
- Remove low-level compile features (closes #1272)
- Remove unneeded DESTINATION options where possible

CC: @sum01

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -100,6 +100,7 @@ if (BUILD_SHARED_LIBS AND WIN32 AND HTTPLIB_COMPILE)
 endif()
 
 # Threads needed for <thread> on some systems, and for <pthread.h> on Linux
+set(THREADS_PREFER_PTHREAD_FLAG true)
 find_package(Threads REQUIRED)
 # Since Cmake v3.11, Crypto & SSL became optional when not specified as COMPONENTS.
 if(HTTPLIB_REQUIRE_OPENSSL)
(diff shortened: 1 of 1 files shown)


sum01:
Looks good.

I don't understand the point in changing DESTINATION, but it's not important I guess. I just say this because `TYPE INCLUDE` seems to evaluate to read from `CMAKE_INSTALL_INCLUDEDIR` anyways (if I'm reading the docs right).

As for the prefer pthread flag, is there a reason for that? I'm just curious.
I know a long time ago I couldn't get things to compile without it (on Linux) but it seems to have since been a non-issue.

Tachi107:
> I don't understand the point in changing DESTINATION, but it's not important I guess. I just say this because `TYPE INCLUDE` seems to evaluate to read from `CMAKE_INSTALL_INCLUDEDIR` anyways

Yes, you're right, but it makes the code a bit cleaner (quite subjective but whatever, I can revert it if you prefer)



> As for the prefer pthread flag, is there a reason for that? I'm just curious.

Yes. The first is that the docs "highly recommend" its usage where possible (i.e. In all projects targeting CMake >= 3.1), the second is that it's also the recommended way of enabling pthreads on compiler that support the flag, like GCC (according to their docs). Simply linking to -lpthread isn't always enough as the pthreads headers could require some macro like REENTRANT (or similar) to be defined, and a linker flag of course doesn't define that. So -pthread takes care of defining the necessary macros and linking to the necessary libraries.

sum01:
Thanks for the response. You don't need to change the `TYPE INCLUDE`, it's fine.

yhirose:
@Tachi107, is everything ready for merge?

Tachi107:
Yes

yhirose:
Thanks for your help!

Issue #1134, opened 2021-12-15
POSTing large Text


npritchard:
Hi thanks for the fantastic library - a lot of effort is greatly appreciated.

I'm posting large volumes of data to a cpp-httplib server and getting error 413 in response. It's returned as a result of this section around line 5004 in the latest version:

```
      const auto& content_type = req.get_header_value("Content-Type");
      if (!content_type.find("application/x-www-form-urlencoded")) {
        if (req.body.size() > CPPHTTPLIB_REQUEST_URI_MAX_LENGTH) {
          res.status = 413; // NOTE: should be 414?
          return false;
        }
        detail::parse_query_text(req.body, req.params);
      }
```

So what I understand is that, unless there's a Content-Type header 'application/x-www-form-urlencoded', this makes sure that the body of the data cannot be larger than CPPHTTPLIB_REQUEST_URI_MAX_LENGTH. 

Is this the correct logic?

Even so, I'm not sure what this line is doing:

`      if (!content_type.find("application/x-www-form-urlencoded")) {`

Should that not be:
`      if (content_type.find("application/x-www-form-urlencoded") == std::string::npos) {`

Thanks

yhirose:
That's opposite. If there's a Content-Type header 'application/x-www-form-urlencoded', then do the CPPHTTPLIB_REQUEST_URI_MAX_LENGTH check. If the test succeeds, then parse the form text and construct req.params.
Hope the explanation helps!

PR #1289, opened 2022-05-25
fix incorrect checking of Content-Type in Server::read_content

### LINKED ISSUE #1288: Large URI payload handled incorrectly
httplib denies any URI with length > CPPHTTPLIB_REQUEST_URI_MAX_LENGTH, even if the header `Content-Type` is set to `application/x-www-form-urlencoded`.

This can be reproduced by sending a form of size > 8192 to an httplib server. I believe this is due to a typo. PR coming up.

lionkor:
fixes #1288.

This was obviously just an oversight and wasn't supposed to act like that.

`std::string::find` will return the *position* (index) of the value, if found, which will be `0` (implicit false) if the Content-Type is exactly "application/x-www-form-urlencoded". This PR fixes this by correctly checking for `std::string::npos`, which is the "not found" of `std::string::find()`.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -5188,7 +5188,7 @@ inline bool Server::read_content(Stream &strm, Request &req, Response &res) {
             return true;
           })) {
     const auto &content_type = req.get_header_value("Content-Type");
-    if (!content_type.find("application/x-www-form-urlencoded")) {
+    if (content_type.find("application/x-www-form-urlencoded") == std::string::npos) {
       if (req.body.size() > CPPHTTPLIB_REQUEST_URI_MAX_LENGTH) {
         res.status = 413; // NOTE: should be 414?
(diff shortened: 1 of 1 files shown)


yhirose:
@lionkor, thank you for your pull request, but it breaks a number of unit test cases in `test/test.cc`. (You can see with `cd test && make`... Please see #1134.

By the way, `!find("...")` or `find("...") == 0` represents 'starts with'. (From C++20, `std::starts_with` is available.)
You can see more information from [this SO post](https://stackoverflow.com/questions/8095088/how-to-check-string-start-in-c) or [this one](https://stackoverflow.com/questions/1878001/how-do-i-check-if-a-c-stdstring-starts-with-a-certain-string-and-convert-a).

Hope it helps!

Issue #1285, opened 2022-05-24
Unable to upload large data via post-request


chrxh:
Hello,

I have a problem with sending larger data. It seems to be related to https://github.com/yhirose/cpp-httplib/issues/1211.
Here is a minimal example for my case:

```
httplib::SSLClient client(_address);
client.set_ca_cert_path("./resources/ca-bundle.crt");
client.enable_server_certificate_verification(true);

std::string test(1000000, '0');
httplib::MultipartFormDataItems items = {{"data", test, "", "application/octet-stream"}};
auto result = client.Post("/upload.php", items);

```
I obtain a `Write` (5) error. If the test string is smaller (~10 KB) then everything works.
Did I forget to configure something or is it because of the server (it's a hosted web server)?
(On the other hand, I had already managed to send larger data to this server via post using Qt and it worked.)

Thanks for the help in advance

yhirose:
@chrxh, thank you for the report. Your code looks ok, and I honestly don't know why it returns with the Write error. Could you run the following unit test which uploads 2MB data to a server to find more information?

https://github.com/yhirose/cpp-httplib/blob/47044c05a8587dff86ab90526daabfef61079490/test/test.cc#L4881

Here is my result on my macOS.

```
[master] ~/cpp-httplib/test$ make test && ./test --gtest_filter="MultipartFormDataTest.LargeData"
make: `test' is up to date.
Running main() from gtest/gtest_main.cc
Note: Google Test filter = MultipartFormDataTest.LargeData
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from MultipartFormDataTest
[ RUN      ] MultipartFormDataTest.LargeData
[       OK ] MultipartFormDataTest.LargeData (1100 ms)
[----------] 1 test from MultipartFormDataTest (1100 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (1101 ms total)
[  PASSED  ] 1 test.
```

yhirose:
I'll leave it open for a while. Please let me know if you find more information. Thanks!

chrxh:
I've now found out that I was using an outdated version of cpp-httplib via vcpkg (namely 0.9.7). In the latest version my problem does not occur at all. Has something been fixed in this context?
In any case my I can close my pull request.

PR #1291, opened 2022-05-26
writing data in chunks

chrxh:
Hi,
this PR is not really meant for merging yet. I found out that if one writes the data in chunks in `write_data`, my problem in https://github.com/yhirose/cpp-httplib/issues/1285 doesn't occur anymore.
Unfortunately, I do not know the cause. But maybe this PR could help other users experiencing a similar problem.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -109,6 +109,10 @@
 #define CPPHTTPLIB_LISTEN_BACKLOG 5
 #endif
 
+#ifndef CPPHTTPLIB_WRITE_CHUNK_SIZE
+#define CPPHTTPLIB_WRITE_CHUNK_SIZE 0x4000
+#endif
+
 /*
  * Headers
  */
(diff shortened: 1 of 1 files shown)


yhirose:
@chrxh, thank you for the pull request. I don't prefer introducing another buffer layer though, I am OK to accept it if this is only way to fix your problem described in #1285. How did you decide the chunk buffer size as 16K (0x4000)? Is there any particular reason?

chrxh:
Yes I understand the objection. The pull request was not intended for merging in this form either.
The chunk buffer size is just a guess. I found out that it works for me up to 81920 bytes and after that I got the write error. Sending large POST request with curl is working. That's all I have found out so far.

chrxh:
My pull request should now be obsolete (see comment in https://github.com/yhirose/cpp-httplib/issues/1285).

PR #1294, opened 2022-05-27
resolve http server can't send file large than 2GB (Fix #1290)

### LINKED ISSUE #1290: http server can't provide zip file large than 2GB with project both defined _WIN32 and _WIN64
if project both defined _WIN32 and _WIN64, http server can't provide zip file large than 2GB

in function "SocketStream::write", change:

#ifdef _WIN32
  size =
      (std::min)(size, static_cast<size_t>((std::numeric_limits<int>::max)()));
#endif

to:

#if defined(_WIN32) && !defined(_WIN64)
  size =
      (std::min)(size, static_cast<size_t>((std::numeric_limits<int>::max)()));
#endif

resolve this problem.

conghuawang:
resolve problem: http server can't send file large than 2GB. 
add unit test for http server send file large than 2GB.  
add /bigobj compile option to msvc x64.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -4691,7 +4691,7 @@ inline ssize_t SocketStream::read(char *ptr, size_t size) {
 inline ssize_t SocketStream::write(const char *ptr, size_t size) {
   if (!is_writable()) { return -1; }
 
-#ifdef _WIN32
+#if defined(_WIN32) && !defined(_WIN64)
   size =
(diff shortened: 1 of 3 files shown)


yhirose:
@conghuawang, thanks for the pull request! Could you take a look at the test failure on windows-latest in GitHub Actions workflow?

yhirose:
Thanks for your conribution!

PR #1297, opened 2022-05-30
clang-tidy suggested fixes

GiuseppeCesarano:
This PR applies neat corrections at source and small hand-made solutions. Some of these changes are great for performance, such as avoiding unnecessary copies, others are just for code readability, but each of these changes have been manually revised. I also confirm that the changes pass the test suite

Main changes:
- Multiple parameters declared as const references in function signatures to avoid unnecessary copies
- Where possible member functions have been made static
- Avoid implicit conversions
- Removed default destructors where possible

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -722,7 +722,7 @@ private:
       std::vector<std::pair<std::regex, HandlerWithContentReader>>;
 
   socket_t create_server_socket(const char *host, int port, int socket_flags,
-                                const SocketOptions& socket_options) const;
+                                const SocketOptions &socket_options) const;
   int bind_internal(const char *host, int port, int socket_flags);
   bool listen_internal();
 
@@ -752,7 +752,7 @@ private:
   bool read_content(Stream &strm, Request &req, Response &res);
   bool
   read_content_with_content_receiver(Stream &strm, Request &req, Response &res,
-                                     const ContentReceiver& receiver,
+                                     const ContentReceiver &receiver,
                                      MultipartContentHeader multipart_header,
                                      ContentReceiver multipart_receiver);
   bool read_content_core(Stream &strm, Request &req, Response &res,
@@ -868,20 +868,20 @@ public:
   Result Get(const char *path, const Headers &headers);
   Result Get(const char *path, Progress progress);
   Result Get(const char *path, const Headers &headers, Progress progress);
-  Result Get(const char *path, const ContentReceiver& content_receiver);
+  Result Get(const char *path, const ContentReceiver &content_receiver);
   Result Get(const char *path, const Headers &headers,
-             const ContentReceiver& content_receiver);
-  Result Get(const char *path, const ContentReceiver& content_receiver,
+             const ContentReceiver &content_receiver);
+  Result Get(const char *path, const ContentReceiver &content_receiver,
              Progress progress);
   Result Get(const char *path, const Headers &headers,
-             const ContentReceiver& content_receiver, Progress progress);
+             const ContentReceiver &content_receiver, Progress progress);
   Result Get(const char *path, ResponseHandler response_handler,
-             const ContentReceiver& content_receiver);
+             const ContentReceiver &content_receiver);
   Result Get(const char *path, const Headers &headers,
              ResponseHandler response_handler,
-             const ContentReceiver& content_receiver);
+             const ContentReceiver &content_receiver);
   Result Get(const char *path, ResponseHandler response_handler,
-             const ContentReceiver& content_receiver, Progress progress);
+             const ContentReceiver &content_receiver, Progress progress);
   Result Get(const char *path, const Headers &headers,
              ResponseHandler response_handler,
              const ContentReceiver &content_receiver, Progress progress);
@@ -889,7 +889,8 @@ public:
   Result Get(const char *path, const Params &params, const Headers &headers,
              Progress progress = nullptr);
   Result Get(const char *path, const Params &params, const Headers &headers,
-             const ContentReceiver& content_receiver, const Progress& progress = nullptr);
+             const ContentReceiver &content_receiver,
+             const Progress &progress = nullptr);
   Result Get(const char *path, const Params &params, const Headers &headers,
              const ResponseHandler &response_handler,
              const ContentReceiver &content_receiver,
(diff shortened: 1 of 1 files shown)


yhirose:
@GiuseppeCesarano, thanks for the pull request, but there are so many thing is one pull request and a bit hard to me to review... Could you break it into smaller pull requests based on objectives? Thanks a lot!

httplib.h
```diff
@@ -1865,7 +1869,7 @@ class stream_line_reader {
 namespace detail {
 
 inline bool is_hex(char c, int &v) {
-  if (0x20 <= c && isdigit(c)) {
+  if (0x20 <= c && (isdigit(c) != 0)) {
```
Tachi107, line 1872: Checking that the return value of `isdigit()` is not zero decreases readability - the fact that the function returns `int` is almost irrelevant, as the function returns "Non-zero if the character is a numeric character, zero otherwise", i.e. it returns a boolean. The return type is `int` probably because when `isdigit()` was added to the C standard library `_Bool` wasn't even a thing.
By the way, you might want to change this to [`std::isdigit()`](https://en.cppreference.com/w/cpp/string/byte/isdigit)

httplib.h
```diff
@@ -2024,7 +2028,7 @@ inline std::string encode_query_param(const std::string &value) {
   escaped << std::hex;
 
   for (auto c : value) {
-    if (std::isalnum(static_cast<uint8_t>(c)) || c == '-' || c == '_' ||
+    if ((std::isalnum(static_cast<uint8_t>(c)) != 0) || c == '-' || c == '_' ||
```
Tachi107, line 2031: Same as `std::isdigit()`.

httplib.h
```diff
@@ -2121,7 +2125,7 @@ inline std::string file_extension(const std::string &path) {
   std::smatch m;
   static auto re = std::regex("\\.([a-zA-Z0-9]+)$");
   if (std::regex_search(path, m, re)) { return m[1].str(); }
-  return std::string();
+  return {};
```
Tachi107, line 2128: What clang-tidy diagnostic did you address here exactly (just curious, never saw this before)

httplib.h
```diff
@@ -2193,7 +2197,7 @@ inline bool stream_line_reader::getline() {
   glowable_buffer_.clear();
 
   for (size_t i = 0;; i++) {
-    char byte;
+    char byte = 0;
```
Tachi107, line 2200: Why is this initialization needed?

httplib.h
```diff
@@ -2376,7 +2374,7 @@ inline Error wait_until_socket_is_ready(socket_t sock, time_t sec,
     socklen_t len = sizeof(error);
     auto res = getsockopt(sock, SOL_SOCKET, SO_ERROR,
                           reinterpret_cast<char *>(&error), &len);
-    auto successful = res >= 0 && !error;
+    auto successful = res >= 0 && (error == 0);
```
Tachi107, line 2377: I find `!error` ("not error") more readable than `error == 0` ("error is zero")

httplib.h
```diff
@@ -2521,12 +2519,12 @@ inline int shutdown_socket(socket_t sock) {
...
 socket_t create_socket(const char *host, const char *ip, int port,
                        int address_family, int socket_flags, bool tcp_nodelay,
-                       SocketOptions socket_options,
+                       const SocketOptions &socket_options,
                        BindOrConnect bind_or_connect) {
   // Get address info
   const char *node = nullptr;
-  struct addrinfo hints;
-  struct addrinfo *result;
+  struct addrinfo hints {};
```
Tachi107, line 2526: Initializing `hints` here is useless, as `std::memset` is called below to zero-initialize the struct as well. You either initialize it with `{}` or `memset`

httplib.h
```diff
@@ -2631,20 +2629,20 @@ inline bool is_connection_error() {
 }
 
 inline bool bind_ip_address(socket_t sock, const char *host) {
-  struct addrinfo hints;
-  struct addrinfo *result;
+  struct addrinfo hints {};
```
Tachi107, line 2632: Same as above

httplib.h
```diff
@@ -2660,25 +2658,27 @@ inline bool bind_ip_address(socket_t sock, const char *host) {
 
 #ifdef USE_IF2IP
 inline std::string if2ip(int address_family, const std::string &ifn) {
-  struct ifaddrs *ifap;
+  struct ifaddrs *ifap = nullptr;
```
Tachi107, line 2661: `ifap` is initialized with `getifaddrs(&ifap)` below, setting it to `nullptr` is not needed.

httplib.h
```diff
@@ -2799,11 +2797,11 @@ inline bool get_remote_ip_and_port(const struct sockaddr_storage &addr,
 }
 
 inline void get_remote_ip_and_port(socket_t sock, std::string &ip, int &port) {
-  struct sockaddr_storage addr;
+  struct sockaddr_storage addr {};
```
Tachi107, line 2800: `addr` is initialized below with `getpeername()`

httplib.h
```diff
@@ -3186,9 +3187,9 @@ inline bool brotli_decompressor::decompress(const char *data,
     return 0;
   }
 
-  const uint8_t *next_in = (const uint8_t *)data;
+  const auto *next_in = (const uint8_t *)data;
   size_t avail_in = data_length;
-  size_t total_out;
+  size_t total_out = 0;
```
Tachi107, line 3192: `total_out` is initialized by `BrotliDecoderDecompressStream`

httplib.h
```diff
@@ -3371,28 +3374,30 @@ inline bool read_content_chunked(Stream &strm,
 
     if (!line_reader.getline()) { return false; }
 
-    if (strcmp(line_reader.ptr(), "\r\n")) { break; }
+    if (static_cast<int>(strcmp(line_reader.ptr(), "\r\n") != 0) != 0) { break; }
```
Tachi107, line 3377: What is this `static_cast` doing here?

httplib.h
```diff
@@ -5964,7 +5970,7 @@ inline bool ClientImpl::send(Request &req, Response &res, Error &error) {
 #ifdef CPPHTTPLIB_OPENSSL_SUPPORT
       // TODO: refactoring
       if (is_ssl()) {
-        auto &scli = static_cast<SSLClient &>(*this);
+        auto &scli = dynamic_cast<SSLClient &>(*this);
```
Tachi107, line 5973: Why `dynamic_cast`?

httplib.h
```diff
@@ -6706,9 +6715,8 @@ inline Result ClientImpl::Post(const char *path, const Headers &headers,
 inline Result ClientImpl::Post(const char *path, const Headers &headers,
                                const MultipartFormDataItems &items,
                                const std::string &boundary) {
-  for (size_t i = 0; i < boundary.size(); i++) {
-    char c = boundary[i];
-    if (!std::isalnum(c) && c != '-' && c != '_') {
+  for (char c : boundary) {
+    if ((std::isalnum(c) == 0) && c != '-' && c != '_') {
```
Tachi107, line 6719: same as before

Tachi107 (changes requested):
Overall this doesn't look bad, but a lot of changes are questionable. I personally like to follow clang-tidy's advices too, but only when it makes sense to do so. For example, there are a lot of useless (pessimizing?) initializations; I know that you should generally initialize your variables, but not when its initialization is handled by a function that takes it by non-const-reference- it makes the zero-initialization 100% useless.

One other thing that you should consider is adding a `.clang-tidy` file to the repo with the specific  checks that you followed, so that everybody knows what guidelines to follow and this PR doesn't end up in cleaning some code now but without preventing mistakes in the future.

Thanks for this PR though! The C++ community really needs to embrace clang-tidy a bit more.

httplib.h
```diff
@@ -823,7 +824,7 @@ class Result {
       : res_(std::move(res)), err_(err),
         request_headers_(std::move(request_headers)) {}
   // Response
-  operator bool() const { return res_ != nullptr; }
+  explicit operator bool() const { return res_ != nullptr; }
```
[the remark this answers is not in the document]
GiuseppeCesarano: This could definitely break user's code. I'll remove the keyword explicit since not everyone have implicit conversions disabled

httplib.h
```diff
@@ -1778,20 +1782,20 @@ class nocompressor : public compressor {
...
-  ~gzip_compressor();
+  ~gzip_compressor() override;
 
   bool compress(const char *data, size_t data_length, bool last,
                 Callback callback) override;
 
 private:
   bool is_valid_ = false;
-  z_stream strm_;
+  z_stream strm_{};
```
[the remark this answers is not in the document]
GiuseppeCesarano: It is not strictly necessary but it does not become runtime code and can avoid the attempt to access an uninitialized variable

GiuseppeCesarano:
Thanks for the suggestions, I have implemented most of them. Let me know if other changes are needed and if the PR still needs to be broken down into smaller ones anyway.
  
My clang-tidy setup is as follows:
```
---
Checks: "*,
         -abseil- *,
         -altera- *,
         -android- *,
         -fuchsia- *,
         -Google-*,
         -llvm *,
         -modernize-use-trailing-return-type,
         -zircon- *,
         -readability-else-after-return,
         -readability-static-accessed-through-instance,
         -readability-avoid-const-params-in-decls,
         -cppcoreguidelines-non-private-member-variables-in-classes,
         -misc-non-private-member-variables-in-classes,
"
WarningsAsErrors: ''
HeaderFilterRegex: ''
FormatStyle: none
```  
Many warnings like "No magic numbers" have been ignored to avoid too invasive changes to the code base, but probably using constants for http states is a good idea.  

The file has not been added to the PR because I believe rules should be decided by someone who works actively on the project but still I think these settings are a good place to start

Tachi107:
> Let me know if other changes are needed and if the PR still needs to be broken down into smaller ones anyway.

Could you please clean up the commit history? And instead of splitting this into different PRs you could simply split your changes in smaller commits (e.g. one for pass-by-copy -> pass-by-reference, one for explicit checks for `== 0`, etc).

Oh and by the way, my feedback regarding `std::isnum()` and co. only applies to the various `std::is*` functions, as they really return only a bool. `std::strcmp()` does not return a bool, but three different values (`<1`, `0`, `>1`), so checking that the return value is `== 0` is fine IMO.

httplib.h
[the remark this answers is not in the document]
GiuseppeCesarano: If the compiler has visibility on the reassign (like in this case) even at -O1 the initialization will not take place if the value is overwritten without being read, but there are some cases that where changed where the initialization could cost a move like for the `ifap` initialization, i will look in each added initialization

yhirose:
@GiuseppeCesarano, @Tachi107, the outputs from clang-tidy look pretty good overall, but I don't agree upon all of them. Some of them are obviously too strict to me and sacrificing code readability. That's why I requested @GiuseppeCesarano to break it into smaller pull requests based on categories, so that I can decide whether or not to adopt each category. Thanks for your understanding!

GiuseppeCesarano:
I will close this PR and break the changes into smaller PRs

PR #1302, opened 2022-06-04
Avoid copies where possible

GiuseppeCesarano:


ylz-at:
Well, this is good for performance.

yhirose:
@GiuseppeCesarano, thank you for the pull request. I took a look at it today, and I don't agree with you on some changes.

For instance, on my use cases, most of handlers are defined as lambda functions which are 'rvalue's.

```cpp
svr.Get("/hi", [](const Request& req, Response& res) {
});

std::string body;
auto res = cli.Get("/large-data",
  [&](const char *data, size_t data_length) {
    body.append(data, data_length);
    return true;
  });
```

In those cases, I think the current signature (passing by value) is faster than the const reference one (passing by reference) because unnecessary copy won't happen.

But suggested changes in other areas in this pull request look ok to me.

Could you take a look at the following so post, and let me know your thought?
https://stackoverflow.com/questions/10231349/are-the-days-of-passing-const-stdstring-as-a-parameter-over

Thanks!

GiuseppeCesarano:
I thought about the exact scenario you gave me when I made the changes to the handlers; I know that in this case the new function signature would be less efficient since it would cause the pointer to be deferenced before making the function call, but on the other hand if you pass a `std::functional` it is probably because the function is large in size and the intention is to reuse it, a single copy of an object containing a large function could cost much more than a few pointer deferensiations. Obviously you know better the use cases of your library so if you think it's better to pass by value anyway, I'll bring it back to its original state.

As for the post on stackoverflow, passing by value most of the time results in an unnecessary copy, rather than saying "the days of passing by const reference `std::string` are over" I would encourage to pay more attention to the use of string, because if in any case a copy is to be made then it is definitely better to pass by value in the hope that a move can take place and in fact, I removed the cost reference in the constructors where it was guaranteed that a copy would take place in any case, in this way we enable the possibility of a move to take place.

If you read the accepted answer, you realize that the problem is that function C is lying. In its signature it says that it doesn't need a copy but then makes one, which could very well be affecting the way B was written as well.

yhirose:
I'll close it for now.

PR #1319, opened 2022-07-08
Skip preamble and epilogue in multipart/form-data (Fix #1317)

### LINKED ISSUE #1317: class MultipartFormDataParser cannot handle content starts with 2 CRLF
**Description:**

when handling multipart/form-data content, function `httplib::detail::MultipartFormDataParser::parse()` will fail contents that does not start with the boundary delimiter (line #3798).

https://github.com/yhirose/cpp-httplib/blob/caa31aafda38af208ba7a4790d9ab88f741a9687/httplib.h#L3796-L3798

example multipart/form-data content:
```
\r\n\r\n------53014704754052338\r\nContent-Disposition: form-data; name=\"file\"; filename=\"346ba067e51efc5f.jpg\"\r\nContent-Type: image/jpeg\r\n\r\n\377\330\377............
```

Above form-data work perfectly with other websites.

However, I think [RFC 2046](https://www.rfc-editor.org/rfc/rfc2046.txt) acquiesce CRLF preceding the boundary delimiter line
> The CRLF preceding the boundary delimiter line is conceptually
   attached to the boundary

Is this a bug? Or I misunderstand how boundary delimiter works?

**Possible fix:**

maybe left trim all the CRLFs before parsing form-data?

Gavin1937:


yhirose:
@Gavin1937, thanks for the pull request. It works for 'preamble', but doesn't handle 'epilogue'. https://github.com/yhirose/cpp-httplib/issues/1317#issuecomment-1178539082
I'll take a look at it.

yhirose:
Moved to #1320

Gavin1937:
I actually tried to handle 'epilogue' but not sure whether it is appropriate or not. 
Here is my solution:
https://github.com/yhirose/cpp-httplib/blob/caa31aafda38af208ba7a4790d9ab88f741a9687/httplib.h#L3897-L3900

change above block to:
```cpp
      case 5: { // Done
        buf_epos_ = buf_spos_;
        return false;
      }
```

In current `MultipartFormDataParser::parse()` function, we will enter above switch block after satisfying 2 conditions:
1. `buf_start_with(dash_ + crlf_) == true`, or `buf_spos_` is at a close-delimiter ([this if statement](https://github.com/yhirose/cpp-httplib/blob/caa31aafda38af208ba7a4790d9ab88f741a9687/httplib.h#L3887) is true)
2. `buf_size() > 0` after satisfy above condition, which means there are extra bytes after close-delimiter

In my solution above, by setting `buf_epos_ = buf_spos_;`, we can right trim all the extra bytes after close-delimiter.

However, I read RFC 2046 again and find out there can be a transport-padding in-between the close-delimiter and [CRLF epilogue]

>      multipart-body := [preamble CRLF]
>                     dash-boundary transport-padding CRLF
>                     body-part *encapsulation
>                     close-delimiter transport-padding
>                     [CRLF epilogue]
>      transport-padding := *LWSP-char
>                        ; Composers MUST NOT generate
>                        ; non-zero length transport
>                        ; padding, but receivers MUST
>                        ; be able to handle padding
>                        ; added by message transports.

I'm not sure how to handle this transport-padding. Can we ignore it? Or do we need to do something extra for it?
And do you have any example request content with transport-padding?

Please correct me if I have any misunderstanding.

PR #1327, opened 2022-07-15
Enable specification of interface address for clients on Windows and Android too

Hual:


yhirose:
@Hual, thanks for the contribution. Could you explain more about the motivation of this pull request?

yhirose:
I am not comfortable with returning `std::string();` from `if2ip`. I'll close it for now unless there is a valid reason to do it.

Issue #1322, opened 2022-07-12
Question/Request: Why does Client::Get not accept std::string as well?


n00b42:
Is there a particular reason why `Server::Get` accepts std::string, but `Client::Get` does not?

yhirose:
@n00b42, thanks for the question. There is not a particular reason. I may change both Client and Serve to use `std::string` for consistency.

Tachi107:
> https://stackoverflow.com/questions/57951871/is-there-sense-in-using-const-stdstring-arguments-in-c17

I usually work around the NULL-termination issue by not using `std::string_view` directly, but by instead using a C++20 concept that requires that the `std::string_view` is either convertible to a `const char*` or a `std::string`, thus NULL-terminated. This way the user can pass into functions using this concept either a `const char*` or a `std::string`, but not a `std::string_view` that's not NULL-terminated, without having to do runtime checks.

```cpp
// NULL-terminated std::string_view
template<typename T>
concept Text = std::convertible_to<T, std::string_view> && (std::convertible_to<T, std::string> || std::convertible_to<T, const char*>);
```

This of course requires C++20, and cpp-httplib only uses C++11, but when the day will come you could look into using this little trick :)

PR #1331, opened 2022-07-22
Changed to use `const std::string &` as much as possible instead of `const char *`

yhirose:


Tachi107:
Just curious, what's the rationale behind this change?

yhirose:
According to this suggestion a while back.
https://github.com/yhirose/cpp-httplib/issues/1322

Tachi107:
Ok, thanks for the answer!

Issue #1333, opened 2022-07-26
Enabling SSL causes `max` definition conflict


oculusdrifts:
As per the title, enabling SSL causes a `max` definition conflict when `NOMINMAX` **isn't** defined with Visual Studio 2017 on Windows.

`WIN32_LEAN_AND_MEAN` is defined. `NOMINMAX` is not defined (unable to disable it for legacy reasons).

The compilation error:

`1>httplib.h(7265): error C2589: '(': illegal token on right side of '::' (compiling source file application.cpp)
1>httplib.h(7265): error C2062: type 'unknown-type' unexpected (compiling source file application.cpp)
1>httplib.h(7265): error C2144: syntax error: 'unknown-type' should be preceded by '(' (compiling source file application.cpp)
1>httplib.h(7265): error C2059: syntax error: ')' (compiling source file application.cpp)
1>httplib.h(7267): error C3536: 'handle_size': cannot be used before it is initialized (compiling source file application.cpp)
`

It specifically references this line within `SSLSocketStream::write`:

`auto handle_size = static_cast<int>(
        std::min<size_t>(size, std::numeric_limits<int>::max()));`

To resolve this, simply wrap `std::numeric_limits<int>::max` in parenthesis:

`auto handle_size = static_cast<int>(
        std::min<size_t>(size, (std::numeric_limits<int>::max)()));`

yhirose:
@oculusdrifts, thanks for the report. Could you send me a pull request?

yhirose:
Handled by #1334.

PR #1334, opened 2022-07-27
With SSL enabled and NOMINMAX not defined, there is a conflict with '…

oculusdrifts:
…max', which this fixes

See issue #1333


yhirose:
@oculusdrifts, thanks for the fine contribution!

PR #1335, opened 2022-07-27
Fix: openssl memory leak issue.

pymumu:


yhirose:
@pymumu, thank you for the pull request, but according to the OpenSSL manual, we are discouraged from calling OPENSSL_cleanup, and supposed to rely on the auto-deinitialisation unless there is a valid reason. So I'll close it.
Please let me know if I am missing something regarding this matter. Thanks!

> The OPENSSL_cleanup() function deinitialises OpenSSL (both libcrypto and libssl). All resources allocated by OpenSSL are freed. Typically there should be no need to call this function directly as it is initiated automatically on application exit. This is done via the standard C library atexit() function. 

> Developers of libraries using OpenSSL are discouraged from calling this function and should instead, typically, rely on auto-deinitialisation. This is to avoid error conditions where both an application and a library it depends on both use OpenSSL, and the library deinitialises it before the application has finished using it.

https://www.openssl.org/docs/man1.1.1/man3/OPENSSL_cleanup.html

pymumu:
Understand, I found the problem when I used valgrind to detect memory leak.

Issue #1336, opened 2022-07-28
Bad/Conflicting ssize_t definition on Windows


oculusdrifts:
I've just added a separate library into one of my applications and noticed there is a conflict happening with the definition of ssize_t:

1>x.h(43,14): error C2371: 'ssize_t': redefinition; different basic types
1>httplib.h(135): message : see declaration of 'ssize_t'

x.h defines ssize_t on Windows as:

#if !defined(_POSIX_C_SOURCE) && defined(_MSC_VER)
// 64-bit Windows is the only mainstream platform
// where sizeof(long) != sizeof(void*)
#ifdef _WIN64
typedef long long ssize_t;
#else
typedef long ssize_t;
#endif
#endif

httplib.h defines ssize_t on Windows as:

#ifdef _WIN64
using ssize_t = __int64;
#else
using ssize_t = int;
#endif
#endif // _MSC_VER

Windows has a definition of SSIZE_T (note the capilisation) in basetsd.h (VS 20190) as: typedef LONG_PTR SSIZE_T, *PSSIZE_T;

Changing the httplib ssize_t definition to long resolves the problem. Are we incorrectly defining it?

yhirose:
@oculusdrifts, thanks for the report. I am fine with `using ssize_t = long;` as you suggests according to the Microsoft's documentation. Could you confirm on your machine and send me a pull request? Thanks!

<img width="687" alt="image" src="https://user-images.githubusercontent.com/357397/181564615-419a7826-58f6-491e-a198-eb3338dd9e18.png">
https://docs.microsoft.com/en-us/windows/win32/winprog/windows-data-types

yhirose:
Moved to #1337.

PR #1337, opened 2022-07-28
ssize_t redefinition on Windows - int/int64 vs long/long long

oculusdrifts:
Per issue #1336

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -130,9 +130,9 @@
 #pragma comment(lib, "ws2_32.lib")
 
 #ifdef _WIN64
-using ssize_t = __int64;
+using ssize_t = long long;
 #else
-using ssize_t = int;
+using ssize_t = long;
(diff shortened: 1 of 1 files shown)


httplib.h
```diff
@@ -130,9 +130,9 @@
 #pragma comment(lib, "ws2_32.lib")
 
 #ifdef _WIN64
-using ssize_t = __int64;
+using ssize_t = long long;
```
yhirose, line 133: @oculusdrifts, I would like to leave it as `__int64`, so that it matches the exactly same as what [the Microsoft documentation describes](https://docs.microsoft.com/en-us/windows/win32/winprog/windows-data-types#size_t). Could you try `__int64` on your machine if it works?

image

image

yhirose:
@oculusdrifts, thanks for your fine contribution!

PR #1342, opened 2022-07-31
Fix tests build warning

Rockybilly:


yhirose:
@Rockybilly, it seems like this pull request also include commits in your previous pull request. Could you make it clean?

Issue #721, opened 2020-10-26
Support for unix sockets


nqle:
Hey,
I didn't know where else to ask.
I was curious about whether there are future plans to support unix domain sockets for local interprocess communication or is this project rather focused on web applications and tcp sockets?

yhirose:
@nqle, thanks for the question. I am using this library for HTTP support in my projects. I don't have any plan to support Unix socket because the projects don't need it. Hope it helps!

okirmis:
@nqle If you still need it, I hacked something together (quick & dirty) for my own spare time project (communicating with docker sockets):

In `create_socket` I added the following code before the `  auto service = std::to_string(port);` line:

```c++
  std::string hostName = host;

  int sock = INVALID_SOCKET;
  if (hostName.size() > 5 && hostName.substr(0, 5) == "unix:") {
      sock = socket(AF_UNIX, SOCK_STREAM, 0);

      sockaddr_un addr{};
      addr.sun_family = AF_UNIX;
      strncpy(
          addr.sun_path,
          hostName.substr(5, hostName.size() - 5).c_str(),
          std::min(sizeof(addr.sun_path) - 1, hostName.size() - 5)
      );

      addrinfo info{};
      info.ai_addr = (sockaddr *)&addr;
      info.ai_addrlen = sizeof(addr);
      info.ai_next = nullptr;
      info.ai_socktype = SOCK_STREAM;
      info.ai_family = AF_UNIX;
      info.ai_canonname = nullptr;
      info.ai_protocol = 0;

      if (bind_or_connect(sock, info)) {
          return sock;
      }

      return INVALID_SOCKET;
  }
```
Also, include `#include <sys/un.h>`.

Then you can run
```c++
httplib::Client client("unix:/var/run/docker.sock");
client.set_default_headers({ { "Host", "localhost" } });
auto response = client.Get("/v1.24/containers/json");
```

for example to talk to the local docker socket.

Maybe it helps!

nqle:
Thanks for your effort! I might use it at some point.

PR #1346, opened 2022-08-01
UNIX domain socket support

gh4ck3r:
Recently, I saw https://github.com/yhirose/cpp-httplib/issues/721, and aware @yhirose doesn't have any plan about it. But, @AhnLab-OSSG uses cpp-httplib on UNIX domain socket so far. So, I'd like to share the functionality to support UNIX domain socket. Hope this useful for somebody like me.


yhirose:
@gh4ck3r, looks very good. Thanks for the fine contribution!

PR #1351, opened 2022-08-04
add multipart formdata for PUT requests.

Gopinath-Kalaiyarasan:
* httplib.h

  add multipart formdata for PUT in addition to POST as some REST
  APIs use that.

  Factor the boundary checking code into a helper and use it from
  both Post() and Put().

* test/test.cc

  add test cases for the above.


yhirose:
@Gopinath-Kalaiyarasan, thanks for your contribution!

PR #1357, opened 2022-08-12
build: set soversion to major.minor

Tachi107:
Release 0.11 broke backwards compatibility, meaning that different cpp-httplib versions are compatible with each other only if the major and minor version numbers are the same.

This patch reflects this in the build systems.

See #1209 for some more context.

@yhirose, is it correct that cpp-httplib only maintains backwards compatibility only if both the major and the minor version components are the same? (i.e. `0.10.3` and `0.10.9` are compatible, but `0.10.9` and `0.11.0` are not)

Cc: @sum01

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -176,7 +176,7 @@ if(HTTPLIB_COMPILE)
 	set_target_properties(${PROJECT_NAME}
 		PROPERTIES
 			VERSION ${${PROJECT_NAME}_VERSION}
-			SOVERSION ${${PROJECT_NAME}_VERSION_MAJOR}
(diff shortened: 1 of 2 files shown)


yhirose:
@Tachi107, thank you for the change. Yes, if only the patch version changes, it guarantees the backward compatibility.

abouvier:
The comment is no more correct, it should be:
```
# Example: if you find_package(httplib 0.5.4)
# then anything >= 0.5.4 and < 0.6 is accepted
```

Tachi107:
@abouvier what are you referring to?

yhirose:
@abouvier thanks for the report. Since I don't maintain CMake related files, could you send a pull request as @Tachi107 mentioned? Thanks!

Issue #1370, opened 2022-08-26
Detect client disconnection on Server-Sent-Event Server


gh4ck3r:
I'm making server functionality of Server-Sent-Event on cpp-httplib for my project.
I aware there's example code (ssesvr.cc) in the repository and that's where I began for my own implementation.

In my requirement for the functionality, SSE server should keep client as long as possible even there's nothing to send. For that my implementation dedicate a thread per client and keeps it as long as connected. Problem is server can't aware unexpected disconnection of client without trying to send some data. If that kind of connection is increased, it'll consume entire threadpool after all. I know SSE comment can be used for it. But, I don't want to use it to check connectivity because it makes bunch of useless data to be transferred.

So, I made my own patch to tackle the problem and run existing unittest on Linux. But, it makes failure on ServerTest.ClientStop testcase at following.
https://github.com/yhirose/cpp-httplib/blob/a9cf09795170a72372cbeb3feef3851b859c0e33/test/test.cc#L1646
I think the `EXPECT_TRUE` above should be reconsidered. Because, a sink passed into content provider could be unwritable for some reason if it's based on SocketStream.
I modified `detail::SocketStream::is_writable()` in the patch to check socket via `is_socket_alive()`. I think, the patch should be applied to `SSLSocketStream` as well. I want to contribute the patch with relative unittest and patch if you're OK with changing `EXPECT_TRUE` above and behavior of `SocketStream` and `SSLSocketStream`.

PixlRainbow:
It must be noted that TCP connections have two streams, an upload stream and a download stream, which can be independently disconnected without affecting each other. Clients can disconnect their upload streams and only keep their download streams alive or vice-versa, so it is possible for a socket to be "half-alive".

gh4ck3r:
@PixlRainbow

gh4ck3r:
@PixlRainbow Thanks for the notes. Those are good point to discuss.
I agree with your first note about "half-alive". Originally, I made the patch with `send(sock_, nullptr, 0, 0)` instead of `is_socket_alive()`. And it works fine on Linux. What I'm not sure is that whether it works on other OS too. That's why I chose `is_socket_alive()` which is exists already and make the test pass. Still, I think, `send(sock_, nullptr, 0, 0)` is better for writable case if it works on other OS too.

[...]

PR #1373, opened 2022-08-29
Detecting client disconnection

gh4ck3r:
As described in #1370, Server requires to detect client disconnection if it offers Server-Sent-Event. Unless, it possibly stall the server in certain circumstance. Because if "chunked content provider" which is a component to offer SSE wants to keep connection without sending data, it'll hold given thread as long as given `DataSink` is "writable". The `DataSink`, wrapper of `SocketStream`, uses underlying `is_writable()` to check it's "writable" of not. But, `SocketStream::is_writable()` returns true even after disconnection of peer. That means "chunked content provider" can't aware disconnection of client without writing anything.

In this PR, `is_writable()` of `SocketStream` and `SSLSocketStream` returns false if socket is disconnected.

I'll be better to detect half-close status. But, it's not covered here.

This is worked from @AhnLab-OSSG

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -4722,7 +4722,8 @@ inline bool SocketStream::is_readable() const {
 }
 
 inline bool SocketStream::is_writable() const {
-  return select_write(sock_, write_timeout_sec_, write_timeout_usec_) > 0;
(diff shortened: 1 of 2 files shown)


yhirose:
@gh4ck3r, thanks for the fine contribution!

PR #1402, opened 2022-10-14
Support compilation on AIX

andreaskem:
Disable the ifaddrs.h include statement and the USE_IF2IP flag on AIX.
AIX is a Unix OS but does not offer the ifaddrs.h header.


yhirose:
Thanks for your contribution!

PR #1411, opened 2022-10-28
Feature: support custom handle_file_request.

eaglexmw-gmail:
Feature: support custom handle_file_request.
Used for other issues: memory file system/Zip file system etc.


yhirose:
@eaglexmw-gmail, thanks for the pull request. But I don't accept it, since `set_pre_routing_handler` doesn't the same thing.

PR #1412, opened 2022-10-28
Feature: use pre_routing_handler_ for memory file system, reuse base_…

eaglexmw-gmail:
Feature: use pre_routing_handler_ for memory file system, reuse base_dirs_/set_base_dir/set_mount_point/remove_mount_point etc.
need change base_dirs_ to protected for derived class.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -718,6 +718,13 @@ protected:
   time_t idle_interval_usec_ = CPPHTTPLIB_IDLE_INTERVAL_USECOND;
   size_t payload_max_length_ = CPPHTTPLIB_PAYLOAD_MAX_LENGTH;
 
+  struct MountPointEntry {
+    std::string mount_point;
+    std::string base_dir;
+    Headers headers;
+  };
+  std::vector<MountPointEntry> base_dirs_;
+
 private:
   using Handlers = std::vector<std::pair<std::regex, Handler>>;
   using HandlersForContentReader =
@@ -763,13 +770,6 @@ private:
                          ContentReceiver multipart_receiver);
 
   virtual bool process_and_close_socket(socket_t sock);
-
-  struct MountPointEntry {
-    std::string mount_point;
-    std::string base_dir;
-    Headers headers;
-  };
-  std::vector<MountPointEntry> base_dirs_;
 
   std::atomic<bool> is_running_;
   std::map<std::string, std::string> file_extension_and_mimetype_map_;


yhirose:
@eaglexmw-gmail, thank you for the pull request. Could you show more information about why you need it? Thanks.

eaglexmw-gmail:
I want add a feature: memory(or Zip) File system support for cpp-httplib,
Like these code:
`class PatchedServer: public Server
{
protected:
  FileSystem* fs_handle;

public:
  PatchedServer(): Server(), fs_handle(nullptr) {
      // 自定义文件处理器
      set_pre_routing_handler(std::bind(&PatchedServer::pre_file_request, this, 
                                            std::placeholders::_1, std::placeholders::_2));
  }

  void set_filesystem_handle(FileSystem* fs) {
      fs_handle = fs;
  }

  HandlerResponse pre_file_request(const Request &req, Response &res){
    // File handler
    bool is_head_request = req.method == "HEAD";
    if ((req.method == "GET" || is_head_request) &&
        ex_handle_file_request(req, res, is_head_request)) {
        return HandlerResponse::Handled;
    }
    return HandlerResponse::Unhandled;
  };

  bool ex_handle_file_request(const Request &req, Response &res, bool head) {
    if (nullptr == fs_handle) {
      return false;
    }
   ...............
  }
};


int main(int argc, char* argv[])
{
    ZipFileSystem zfs;
    // HTTP
    PatchedServer svr;

    const uint8_t* assets = NULL;
    uint32_t assets_len = 0;
    get_web_assets(&assets, &assets_len);

    if (!zfs.init(assets, assets_len)) {
        printf("init zip file system fail.\n");
    }
    svr.set_filesystem_handle(&zfs);
    svr.set_mount_point("/static", "@/");

    svr.listen("0.0.0.0", 8080);

    return 0;
}
`

memory fs or zip fs will provide fs access interface with same folder tree like real fs.
so, I think, We can reuse the set_mount_point and other function.
Just change base_dir_ to protected.

yhirose:
@eaglexmw-gmail, thanks for the pull request. I understand you would like to use `MountPointEntry` and `base_dirs_` with other file systems rather than the OS file system though, I would still prefer keeping them in the private section since it's strongly tied in `Server::handle_file_request` method. So, I suggest to make a similar structure and a member variable. Thanks for your understanding!

PR #1418, opened 2022-11-03
Get client process id over ip/port when server runs on UNIX socket.

gh4ck3r:
I put client process id into `REMOTE_PORT` in http request header when server runs on UNIX socket.
The name might be awkward for process id but, I think, it's acceptable because process id can be used to address client like ip/port. As a matter of fact, process id and ip/port never coexists.

@AhnLab-OSSG


gh4ck3r:
Somebody help me please with restarting tests. It looks failed because environment issue but I can't restart the test.

yhirose:
@gh4ck3r yes, I think it's acceptable to use a process id for port. Thanks again for the fine contribution!

PR #1427, opened 2022-11-13
undef poll at the end if CPPHTTPLIB_USE_POLL

artpaul:
The replacement of `poll` with `WSAPoll` should be reverted to avoid propagation to other source files.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -8405,8 +8405,8 @@ inline SSL_CTX *Client::ssl_context() const {
 
 } // namespace httplib
 
-#ifdef CPPHTTPLIB_USE_POLL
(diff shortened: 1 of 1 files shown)


httplib.h
```diff
@@ -8405,4 +8405,8 @@ inline SSL_CTX *Client::ssl_context() const {
 
 } // namespace httplib
 
+#ifdef CPPHTTPLIB_USE_POLL
```
yhirose, line 8408: Could you change it to `#if defined(_WIN32) && defined(CPPHTTPLIB_USE_POLL)`?

artpaul: Done.

yhirose:
Thanks for your contribution!

Issue #1426, opened 2022-11-12
duration_to_sec_and_usec narrowing


stingray-11:
This occurs with `-Werror=conversion` on a 32-bit Linux system (time_t is 32-bits) and GCC 10.  

```
/home/user/andromeda-desktop-clients/build/_deps/cpp-httplib-build/out/httplib.h: In instantiation of 'void httplib::detail::duration_to_sec_and_usec(const T&, U) [with T = std::chrono::duration<long long int>; U = httplib::ClientImpl::set_read_timeout<long long int, std::ratio<1> >::<lambda(time_t, time_t)>]':
/home/user/andromeda-desktop-clients/build/_deps/cpp-httplib-build/out/httplib.h:1681:35:   required from 'void httplib::ClientImpl::set_read_timeout(const std::chrono::duration<_Rep1, _Period1>&) [with Rep = long long int; Period = std::ratio<1>]'
/home/user/andromeda-desktop-clients/build/_deps/cpp-httplib-build/out/httplib.h:1701:25:   required from 'void httplib::Client::set_read_timeout(const std::chrono::duration<_Rep1, _Period1>&) [with Rep = long long int; Period = std::ratio<1>]'
/home/user/andromeda-desktop-clients/src/lib/andromeda/HTTPRunner.cpp:19:54:   required from here
/home/user/andromeda-desktop-clients/build/_deps/cpp-httplib-build/out/httplib.h:1536:12: error: conversion from 'long long int' to 'time_t' {aka 'long int'} may change value [-Werror=conversion]
 1536 |   callback(sec, usec);
      |            ^~~
```

I'm calling `set_read_timeout(std::chrono::seconds(120))` in my code.  It looks like the std::chrono Rep is a long long int (64-bits) and httplib is assigning it to a time_t (32-bits).  Probably will need a `static_cast<time_t>` somewhere as long as that yields the correct behavior.

stingray-11:
A good solution in general to prevent these warnings from showing up in projects would be to change `CMakeLists.txt` to read `target_include_directories(${PROJECT_NAME} SYSTEM ${_INTERFACE_OR_PUBLIC}`.  This will make the compiler use `-isystem` rather than `-I` which prevents warnings.

yhirose:
@stingray-11, thanks for the report!

PR #1428, opened 2022-11-13
add SYSTEM to include to prevent warnings

stingray-11:
This prevents compile warnings within cpp-httplib from showing up in an end user's build.  

https://github.com/yhirose/cpp-httplib/issues/1426

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -193,7 +193,7 @@ target_compile_features(${PROJECT_NAME} ${_INTERFACE_OR_PUBLIC}
 	cxx_std_11
 )
 
-target_include_directories(${PROJECT_NAME} ${_INTERFACE_OR_PUBLIC}
+target_include_directories(${PROJECT_NAME} SYSTEM ${_INTERFACE_OR_PUBLIC}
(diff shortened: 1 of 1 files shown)


yhirose:
@sum01 do you have any comment on this pull request? Thanks!

sum01:
I've never actually seen `SYSTEM` used, so I wasn't sure of its purpose.

Looking it up, it doesn't seem like it really applies to this project, but maybe I'm wrong? To me, it seems like `SYSTEM` is meant for external includes, not internal ones. @stingray-11

stingray-11:
It is, but this project itself is "external" for other projects that use it.  Other projects might want to be able to enable extra compiler warnings that httplib does not yet pass, and use -Werror

I cannot find the Stack overflow post I was looking at before but this one on the [cmake forum](https://discourse.cmake.org/t/declare-imported-targets-as-system-to-suppress-warnings/555) says "CMake expects the library author not the consumer to set the which include directories should be considered SYSTEM".

sum01:
I guess that makes sense. Weird I've not seen others using it before though.

yhirose:
Thanks for the contribution!

Issue #287, opened 2019-12-12
Is there a non-header-only version of this library available?


KowalskiThomas:
Hi,

I know this might be a pretty naive question, but I was wondering whether a non-header-only version was available for cpp-httplib. I am compiling C++ code in pretty constrained environments and my compile times have skyrocketed since I've started using this library. 

Thanks a lot

yhirose:
@KowalskiThomas, thanks for the request. I am not planning to make a no-header-only version, but I made `split.py` to make `httplib.h` and `httplib.cc` in `out` directory. Does it work for you?

KowalskiThomas:
It's exactly what I needed and what I would have done if you had not. 
Thank you very much!

hak8or:
For anyone googling around and coming to here and wondering if it's worth it to get this going, yes, and it's pretty simple.

To start off, as suggested in the readme, you go into the cpp-httplib directory and run split.py, which gives you ```httplib.cc``` and ```httplib.h```. In your code, you just include the header and link against the cc variant.

On my Ryzen 2700x running GCC 10.2.0, with faa5f1d8023746a3da9f275c51867ded2a672ee9 ```v0.8.6``` of this library, I was able to get compilation from roughly 22.5 seconds down to a blistering **2.8 seconds**!

I cannot thank you enough, both to @yhirose for allowing this to be in the library (and putting it into the readme), and @KowalskiThomas for asking about this so Google indexer bot sees it. The QoL improvement of allowing a non header only variant of your library is genuinely huge via saving time when compiling. Now I just wish this can be somehow integrated into the ```CMakeLists.txt```for this library to step upon #470 (please don't remove CMake support, many of us do actually use it). I might play around with that, but my CMake skills aren't exactly stellar.

yhirose:
@hak8or, thanks for the feedback. `CMakeLists.txt` is kindly maintained by @sum01, and I think it has already the support for `split.py`. Please see the following code.

https://github.com/yhirose/cpp-httplib/blob/faa5f1d8023746a3da9f275c51867ded2a672ee9/CMakeLists.txt#L158-L196

> please don't remove CMake support, many of us do actually use it

The current version of CMakeLists.txt is well maintained even if I didn't make the file. So don't worry, I am not going to remove it. 😄

sum01:
Correct, all you need to do is enable the `HTTPLIB_COMPILE` option when using Cmake and it will manage splitting the file and compiling it for you (requires Python v3 installed).

[...]

PR #1434, opened 2022-11-19
Windows CMake directory install fix

themarpe:
Hi @yhirose @sum01 

Thanks for your initial work on https://github.com/yhirose/cpp-httplib/pull/470

There is a comment which seems to be the rational behind the different CMake installation directory, but I don't think it is relevant (issue: https://github.com/yhirose/cpp-httplib/issues/287)

I therefore assume that this initial split in installation directory was done "from spec", not from practical issue bubbling up - but it's not (was not) correct.

Proposed are two commits. 
1. First one fixes the original behavior - it was close, but that means of searching for CMake config files was only introduced in CMake version 3.25, predating the date of original PR. See: https://cmake.org/cmake/help/latest/command/find_package.html#id3
2. Simplification, which removes the differentiation altogether, and settles for `lib/cmake/[project]/` path - I've integrated many CMake dependencies to date in various projects (one bigger being https://github.com/luxonis/depthai-core) and have never seen this installation path used by any dependency on Windows. Never had issues before.

Issue popped up during integration of the library using Hunter package manager.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -221,15 +221,8 @@ target_compile_definitions(${PROJECT_NAME} ${_INTERFACE_OR_PUBLIC}
 	$<$<BOOL:${HTTPLIB_IS_USING_OPENSSL}>:CPPHTTPLIB_OPENSSL_SUPPORT>
 )
 
-# Cmake's find_package search path is different based on the system
-# See https://cmake.org/cmake/help/latest/command/find_package.html for the list
-if(CMAKE_SYSTEM_NAME STREQUAL "Windows")
-	set(_TARGET_INSTALL_CMAKEDIR "${CMAKE_INSTALL_PREFIX}/cmake")
-else()
-	# On Non-Windows, it should be /usr/lib/cmake/<name>/<name>Config.cmake
-	# NOTE: This may or may not work for macOS...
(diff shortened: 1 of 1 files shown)


sum01:
That distinction was born from my experience trying to get Cmake to find installed things on Windows. Admitedly it was something from a while ago, so maybe it changed by now.

Though I don't think I can really depend on Cmake v3.25 this early. I usually use slower releases like Ubuntu (v3.22) or Debian (v3.18) as a barometer. But if you say it works, I'll plan on testing it on Windows then, although I'd want it to work with the declared minimum Cmake version (v3.14 currently).

themarpe:
> Though I don't think I can really depend on Cmake v3.25 this early. I usually use slower releases like Ubuntu (v3.22) or Debian (v3.18) as a barometer.

I agree, this is what the first commit addresses. The current state is `cmake/[package]/` which is a 3.25 feature (as linked), the commit changing it to: `cmake/` is pre 3.25, which is what i think you intended in the first place

> But if you say it works, I'll plan on testing it on Windows then, although I'd want it to work with the declared minimum Cmake version (v3.14 currently).

Great, do let me know, the latter commit (simplifyed one) is the one I'm in favor of if it works as expected in various cases.

yhirose:
Thanks for the contribution!

PR #1438, opened 2022-11-24
Add EINTR and EAGAIN judge for accept

fishpenguin:
This pr add EINTR and EAGAIN process for socket accept function.

When program is blocked by accept, Capturing some signals like SIGUSR2 may interrupt the accept and return error(-1), set errno to EINTR. When this happens, http service shouldn't be shutdown but should continue to call accept.

I propose this pr because I met such problem, I want to shutdown my http service by stop function when I capture signals, but this will shutdown straightly.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -5484,6 +5484,8 @@ inline bool Server::listen_internal() {
           // Try to accept new connections after a short sleep.
           std::this_thread::sleep_for(std::chrono::milliseconds(1));
           continue;
+        } else if (errno == EINTR || errno == EAGAIN) {
+          continue;
         }
         if (svr_sock_ != INVALID_SOCKET) {
           detail::close_socket(svr_sock_);
--- a/test/test.cc
+++ b/test/test.cc
@@ -943,6 +943,7 @@ TEST(UrlWithSpace, Redirect_Online) {
 
 #endif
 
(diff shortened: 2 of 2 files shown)


yhirose:
@fishpenguin thanks for the pull request. Could you explain more about EAGAIN? In what situation did you receive EAGAIN?

fishpenguin:
@yhirose I didn't meet EAGAIN error, it usually happens when the socket network resource temporarily unavailable. EAGAIN shouldn't be treat as an error, it won't destroy socket. When EAGAIN happens, socket should retry as [man accept](https://man7.org/linux/man-pages/man2/accept.2.html) said.

yhirose:
@fishpenguin, thank you for the reply. Could you add at least one unit test case to validate your modification works? Then, I'll do code review. Thanks!

fishpenguin:
@yhirose sorry, I was busy with other things these days and forgot this pr. I added a unit test: the unit test send SIGINT to the thread which is in listening, and the listen_after_bind() won't quit. If the EINTR isn't judged, it will quit when receive the signal.

yhirose:
@fishpenguin thanks for adding the unit test. Could you take a look at the build error on GitHub Action?

yhirose:
@fishpenguin thanks for the fine contribution!

PR #1448, opened 2022-12-02
std::condition_variable::notify_one/all() should be called after unlocking mutex

jimmy-park:
The notifying thread doesn't need to hold the lock because the notified thread may immediately block again in some implementations. (see description and example in [cppreference](https://en.cppreference.com/w/cpp/thread/condition_variable/notify_one))
There is another [caveat](https://en.cppreference.com/w/cpp/thread/condition_variable/notify_one#:~:text=Notifying%20while%20under,a%20destroyed%20object.) about notifying while holding a lock, but I think it isn't relevant to current implementation of `ThreadPool`.

I also recommend the implementation of task system which presented in [Sean Parent's talk](https://www.youtube.com/watch?v=zULU6Hhp42w).

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -580,7 +580,7 @@ private:
 
           if (pool_.shutdown_ && pool_.jobs_.empty()) { break; }
 
-          fn = pool_.jobs_.front();
+          fn = std::move(pool_.jobs_.front());
           pool_.jobs_.pop_front();
         }
 
--- a/httplib.h
+++ b/httplib.h
@@ -546,8 +546,11 @@ public:
   ~ThreadPool() override = default;
 
   void enqueue(std::function<void()> fn) override {
-    std::unique_lock<std::mutex> lock(mutex_);
-    jobs_.push_back(std::move(fn));
+    {
+      std::unique_lock<std::mutex> lock(mutex_);
+      jobs_.push_back(std::move(fn));
+    }
+
     cond_.notify_one();
   }
 
@@ -556,8 +559,9 @@ public:
     {
       std::unique_lock<std::mutex> lock(mutex_);
       shutdown_ = true;
-      cond_.notify_all();
     }
+
+    cond_.notify_all();
 
     // Join...
     for (auto &t : threads_) {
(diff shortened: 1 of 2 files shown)


yhirose:
@jimmy-park thank you for the reply. Could you add at least one unit test case to verify your modification works? Then, I'll do code review. Thanks!

jimmy-park:
@yhirose I added a simple test which increases an atomic counter using `TaskQueue` . Since `TaskQueue` is only used in `listen_internal()`, I think it acts like a SPMC (single producer multiple consumer) queue and test code follows this behavior. I know this is an oversimplified example. Do we need more test cases for concurrency?

yhirose:
@jimmy-park, the reason why I need the unit test is that I would like to reproduce the problem you mentioned above on my machine and confirm that your modification fixes it. Otherwise, there is no way for me to verify your fix is acceptable. :)

jimmy-park:
@yhirose Understood :smile:

yhirose:
@jimmy-park, I just tried to run `TaskQueueTest.IncreaseAtomicInteger` with the current httplib.h without your change, but it worked fine with no problem. How did you find the current implementation of TaskQueue has a problem?

jimmy-park:
@yhirose The unit test I added was a simple check of the basic functionality of `TaskQueue`. It works in current implementation because all waiting threads will wake up on `TaskQueue::shutdown()`. I think it maybe a performance issue in **some** platforms as [cppreference](https://en.cppreference.com/w/cpp/thread/condition_variable/notify_one#:~:text=However%2C%20some%20implementations%20(in%20particular%20many%20implementations%20of%20pthreads)%20recognize%20this%20situation%20and%20avoid%20this%20%22hurry%20up%20and%20wait%22%20scenario%20by%20transferring%20the%20waiting%20thread%20from%20the%20condition%20variable%27s%20queue%20directly%20to%20the%20queue%20of%20the%20mutex%20within%20the%20notify%20call%2C%20without%20waking%20it%20up.) says, not a functionality issue. It is hard to point out which platform has this issue, so I recommend to follow cppreference's example.

yhirose:
@jimmy-park, thanks for your replay. I read the cppreference and now understand what you were trying to do. Your change looks good to me, and I just merged it. Thanks a lot!

PR #1450, opened 2022-12-04
Support LOCAL_ADDR and LOCAL_PORT header in client Request

ibauersachs:
Having the local address/port is useful if the server is bound to all interfaces, e.g. to serve different content for developers on localhost only.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -413,6 +413,8 @@ struct Request {
 
   std::string remote_addr;
   int remote_port = -1;
+  std::string local_addr;
+  int local_port = -1;
 
   // for server
   std::string version;
@@ -514,6 +516,7 @@ public:
   virtual ssize_t read(char *ptr, size_t size) = 0;
   virtual ssize_t write(const char *ptr, size_t size) = 0;
   virtual void get_remote_ip_and_port(std::string &ip, int &port) const = 0;
+  virtual void get_local_ip_and_port(std::string &ip, int &port) const = 0;
   virtual socket_t socket() const = 0;
 
   template <typename... Args>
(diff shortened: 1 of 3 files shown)


yhirose:
@ibauersachs thank you for the reply. Could you add at least one unit test case to verify your modification works? Then, I'll do code review. Thanks!

ibauersachs:
@yhirose Updated to include a simple test. I would have liked to test against e.g. 127.0.0.2 in addition to 127.0.0.1, but that would require a larger setup for the entire `ServerTest` class or an additional class. Please let me know if it's enough like this or if I should create a distinct test class.

yhirose:
@ibauersachs thanks for adding the test. That's enough for me. After I approved the change, the build on GitHub Actions failed at `FuzzedStream`. Could you fix it and update your pull request? Thanks!

ibauersachs:
I'll look at it.
I didn't notice this because the fuzzer source file unfortunately isn't included in the CMake build.

ibauersachs:
@yhirose Updated. The workflows passed in my fork.

As a side note, when running the tests locally in WSL2 - which doesn't properly support `AF_UNIX` - the `UnixSocketTest` hangs (it waits forever until the server is running, which becomes true). A possible workaround is to assert on a future, e.g. similar to this:

```cpp
TEST_F(UnixSocketTest, pathname) {
  httplib::Server svr;
  svr.Get(pattern_, [&](const httplib::Request &, httplib::Response &res) {
    res.set_content(content_, "text/plain");
  });

  std::packaged_task<bool()> listen_task([&]{
    return svr.set_address_family(AF_UNIX).listen(pathname_, 80);
  });
  auto future = listen_task.get_future();

  std::thread t(std::move(listen_task));
  t.detach();
  ASSERT_EQ(future.wait_for(std::chrono::milliseconds(1000)), std::future_status::ready);
  ASSERT_TRUE(future.get());
  ASSERT_TRUE(svr.is_running());

  client_GET(pathname_);

  svr.stop();
}
```

yhirose:
@ibauersachs thanks for your fine contribution! As for `UnixSocketTest.pathname`, I tested it on my WSL2 and the test passed successfully. So I am leave it as it is until it becomes an issue on my machine.

PR #1454, opened 2022-12-09
add support for requests with both MultipartFormDataItems and Content Providers

stingray-11:
I had a situation where I needed to send a POST request with some multipart parameters, some of which needed to be fed by content providers (chunked transfer) rather than just a string.  I implemented just that.

I tried to get it working with the most idiomatic code for this library but there are still a few combinations missing, e.g. right now it only works with ContentProviderWithoutLength not just ContentProvider.  Maybe it would be a better idea to just use ContentProvider and then users can use the ContentProviderAdapter?  Only problem is this adapter class doesn't seem to be accessible outside httplib.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -6690,7 +6690,7 @@ inline bool ClientImpl::process_request(Stream &strm, Request &req,
   return true;
 }
 
-ContentProviderWithoutLength ClientImpl::get_multipart_content_provider(
+inline ContentProviderWithoutLength ClientImpl::get_multipart_content_provider(
     const std::string &boundary, const MultipartFormDataItems &items,
     const MultipartFormDataProviderItems &provider_items) {
   size_t cur_item = 0, cur_start = 0;
--- a/httplib.h
+++ b/httplib.h
@@ -974,6 +974,9 @@ public:
              const MultipartFormDataItems &items);
   Result Put(const std::string &path, const Headers &headers,
              const MultipartFormDataItems &items, const std::string &boundary);
+  Result Put(const std::string &path, const Headers &headers,
+             const MultipartFormDataItems &items,
+             const MultipartFormDataProviderItems &provider_items);
 
   Result Patch(const std::string &path);
   Result Patch(const std::string &path, const char *body, size_t content_length,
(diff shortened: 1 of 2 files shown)


yhirose:
@stingray-11, thanks for the pull request. Could you provide some unit test cases for the enhancement? Then, I'll take a look at the code. Thanks!

stingray-11:
I need to do some more work on the code, I don't think it's buffering quite the way I thought :)

stingray-11:
Can you have a quick look at the code before I write a formal test?  Just in case you don't like the general layout or it needs major changes, it would be easier to get that together first.

yhirose:
@stingray-11, I took a look at the code briefly, and basically it looks good to me. So could you go a head to add some unit tests, so that we could see actual usages for the enhancement?

Also, could you use 'snake case' instead of 'camel case' for variables and apply clang-format with the `.clang-format` in the root folder? That will help me to see the code more clearly. Thanks!

stingray-11:
Okay, I fixed the snake_case and ran clang-format.  It changed a couple pre-existing lines.  I will write a unit test in the near future.  If you are curious how the code is being used, see - https://github.com/irondrive/andromeda-desktop-clients/blob/master/src/lib/andromeda/backend/HTTPRunner.cpp

stingray-11:
I've been a bit busy lately but I have not forgotten about this.

httplib.h
```diff
@@ -6647,6 +6690,49 @@ inline bool ClientImpl::process_request(Stream &strm, Request &req,
   return true;
 }
 
+inline ContentProviderWithoutLength ClientImpl::get_multipart_content_provider(
```
stingray-11, line 6693: This seems like a lot to be inlining but for some reason I get a linker multiple definition error if I don't.

stingray-11: Might have something to do with the way split.py works?

stingray-11:
The unit test is up!

httplib.h
yhirose: This usage of 'inline' isn't for actual inlining, but avoiding the multiple definition error when more than one .cpp files include httplib.h. So no problem!

yhirose:
Great job! Thanks for your fine contribution. :)

PR #1467, opened 2023-01-13
add to_human_string

stingray-11:
Adds to_human_string() for better error strings.  Some may need tweaking, I took an educated guess on a few.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -826,7 +826,6 @@ enum class Error {
 };
 
 std::string to_string(const Error error);
-std::string to_human_string(const Error error);
 
 std::ostream &operator<<(std::ostream &os, const Error &obj);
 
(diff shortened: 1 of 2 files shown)


stingray-11:
Not sure why the windows test failed, doesn't look related to this.

yhirose:
@stingray-11 sorry for the late reply. I don't mind to replace the existing `to_string` method with your more human readable string instead of introducing `to_human_string`. Could you make the change? Thanks!

stingray-11:
> @stingray-11 sorry for the late reply. I don't mind to replace the existing `to_string` method with your more human readable string instead of introducing `to_human_string`. Could you make the change? Thanks!

Done!

yhirose:
@stingray-11 thanks for your contribution!

PR #1468, opened 2023-01-14
Support CTest

jimmy-park:
I added `test/CMakeLists.txt` because the current testing workflow is separate from the CMake build system.
Now we can do this for testing in a cross-platform way.

```sh
cmake -S . -B build -DHTTPLIB_TEST=ON
cmake --build build && cd build
ctest
```

I also tested `httplib` with `OpenSSL` as shown in `test/Makefile`.
This requires me to generate some cert files in the build directory.
Here is results on my machine (Windows 10)

result-windows

It may test with `Brotli` and `zlib` as well if libraries are installed correctly.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -9,8 +9,7 @@ set(gtest_force_shared_crt ON)
 
 FetchContent_Declare(
     gtest
-    URL https://github.com/google/googletest/archive/release-1.12.1.tar.gz
-    URL_HASH SHA256=81964fe578e9bd7c94dfdb09c8e4d6e6759e19967e397dbea48d1c10e45d0df2
+    URL https://github.com/google/googletest/archive/main.tar.gz
 )
 FetchContent_MakeAvailable(gtest)
 
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -1,0 +1,94 @@
+cmake_policy(SET CMP0135 NEW)
+
+include(FetchContent)
+include(GoogleTest)
+
+set(BUILD_GMOCK OFF)
+set(INSTALL_GTEST OFF)
+set(gtest_force_shared_crt ON)
+
+FetchContent_Declare(
+    gtest
+    URL https://github.com/google/googletest/archive/release-1.12.1.tar.gz
+    URL_HASH SHA256=81964fe578e9bd7c94dfdb09c8e4d6e6759e19967e397dbea48d1c10e45d0df2
+)
+FetchContent_MakeAvailable(gtest)
+
+add_executable(httplib-test test.cc)
+target_compile_options(httplib-test PRIVATE "$<$<CXX_COMPILER_ID:MSVC>:/utf-8>")
+target_link_libraries(httplib-test PRIVATE httplib GTest::gtest_main)
+gtest_discover_tests(httplib-test)
+
+execute_process(
+    COMMAND ${CMAKE_COMMAND} -E copy_directory ${CMAKE_CURRENT_LIST_DIR}/www www
+    COMMAND ${CMAKE_COMMAND} -E copy_directory ${CMAKE_CURRENT_LIST_DIR}/www2 www2
+    COMMAND ${CMAKE_COMMAND} -E copy_directory ${CMAKE_CURRENT_LIST_DIR}/www3 www3
+    COMMAND ${CMAKE_COMMAND} -E copy_if_different ${CMAKE_CURRENT_LIST_DIR}/ca-bundle.crt ca-bundle.crt
+    COMMAND ${CMAKE_COMMAND} -E copy_if_different ${CMAKE_CURRENT_LIST_DIR}/image.jpg image.jpg
+    WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+    COMMAND_ERROR_IS_FATAL ANY
+)
+
+if(HTTPLIB_IS_USING_OPENSSL)
+    find_program(OPENSSL_COMMAND
+        NAMES openssl
+        PATHS ${OPENSSL_INCLUDE_DIR}/../bin
+        REQUIRED
+    )
+    execute_process(
+        COMMAND ${OPENSSL_COMMAND} genrsa 2048
+        OUTPUT_FILE key.pem
+        WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+        COMMAND_ERROR_IS_FATAL ANY
+    )
+    execute_process(
+        COMMAND ${OPENSSL_COMMAND} req -new -batch -config ${CMAKE_CURRENT_LIST_DIR}/test.conf -key key.pem
+        COMMAND ${OPENSSL_COMMAND} x509 -days 3650 -req -signkey key.pem
+        OUTPUT_FILE cert.pem
+        WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+        COMMAND_ERROR_IS_FATAL ANY
+    )
+    execute_process(
+        COMMAND ${OPENSSL_COMMAND} req -x509 -config ${CMAKE_CURRENT_LIST_DIR}/test.conf -key key.pem -sha256 -days 3650 -nodes -out cert2.pem -extensions SAN
+        WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+        COMMAND_ERROR_IS_FATAL ANY
+    )
+    execute_process(
+        COMMAND ${OPENSSL_COMMAND} genrsa 2048
+        OUTPUT_FILE rootCA.key.pem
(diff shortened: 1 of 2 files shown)


jimmy-park:
Testing with `OpenSSL`, `Brotli` and `zlib` was successful in my VM (Ubuntu 22.04).
But it needed to install following libraries.

```sh
sudo apt install -y libssl-dev libbrotli-dev zlib1g-dev
```

jimmy-park:
For some reason, generating `cert2.pem` only fails on macOS unless we add the `-new` option.
1a41e714f76cf57343b2312a1fc3075a5cd99a3d will fix this issue.  

PS. tests in CI sometimes fail due to network issues.

yhirose:
@jimmy-park, thanks for the pull request. I found some embedded strings, 1) physical paths such as `www`, `www2`..., 2) version number like `https://github.com/google/googletest/archive/release-1.12.1.tar.gz`. These my cause problems eventually and need to be taken care of. Please see more information in #955, #1033, #1044.

I actually don't maintain any CMake related files. @sum01 kindly takes care of those files, so that I don't need to worry about it. Could you willingly take a responsibility to maintain `test/CMakeLists.txt` whenever problems happen with it in the future? If so, I'll accept this pull request. Please see my policy about CMake and Meson in the following comment.
https://github.com/yhirose/cpp-httplib/pull/955#issuecomment-855280013

Thanks for your understanding!

jimmy-park:
@yhirose Thanks for replying.

> 1) physical paths such as www, www2...

Since `test.cc` requires some files where it located in and `httplib-test` will be executed in `CMAKE_CURRENT_BINARY_DIR`, I needed to copy these files from the `test` directory to the `CMAKE_CURRENT_BINARY_DIR`. Cert files are created the same way.

> 2) version number like https://github.com/google/googletest/archive/release-1.12.1.tar.gz

We can choose the latest main branch using `https://github.com/google/googletest/archive/main.tar.gz`. It seems like Google also [recommends this way](https://github.com/google/googletest/#live-at-head). Is it okay with this library?

I saw comments and understand what you are concerned about. I'm willing to maintain `test/CMakeLists.txt` and update it when `test/Makefile` changes.

Thanks for the useful library 😄

yhirose:
@jimmy-park thanks for the answers, and accept maintaining it! As you mentioned, could you try to use Google Test from the latest main branch?

jimmy-park:
@yhirose Sure!

yhirose:
@jimmy-park, thanks for the fine contribution!

test/CMakeLists.txt
```diff
@@ -0,0 +1,93 @@
...
+
+execute_process(
+    COMMAND ${CMAKE_COMMAND} -E copy_directory ${CMAKE_CURRENT_LIST_DIR}/www www
+    COMMAND ${CMAKE_COMMAND} -E copy_directory ${CMAKE_CURRENT_LIST_DIR}/www2 www2
+    COMMAND ${CMAKE_COMMAND} -E copy_directory ${CMAKE_CURRENT_LIST_DIR}/www3 www3
+    COMMAND ${CMAKE_COMMAND} -E copy_if_different ${CMAKE_CURRENT_LIST_DIR}/ca-bundle.crt ca-bundle.crt
+    COMMAND ${CMAKE_COMMAND} -E copy_if_different ${CMAKE_CURRENT_LIST_DIR}/image.jpg image.jpg
+    WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+    COMMAND_ERROR_IS_FATAL ANY
+)
```
Tachi107, line 29: Why not use [`file(COPY)`](https://cmake.org/cmake/help/latest/command/file.html#copy)?

test/CMakeLists.txt
```diff
@@ -0,0 +1,93 @@
...
+        WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+        COMMAND_ERROR_IS_FATAL ANY
+    )
+    execute_process(
+        COMMAND ${OPENSSL_COMMAND} req -new -batch -config ${CMAKE_CURRENT_LIST_DIR}/test.conf -key key_encrypted.pem
+        COMMAND ${OPENSSL_COMMAND} x509 -days 3650 -req -signkey key_encrypted.pem
+        OUTPUT_FILE cert_encrypted.pem
+        WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+        COMMAND_ERROR_IS_FATAL ANY
+    )
```
Tachi107, line 92: Since these commands are generating output files, it may be more appropriate to use [`add_custom_command(OUTPUT)`](https://cmake.org/cmake/help/latest/command/add_custom_command.html#generating-files), similar to how the Meson script does:

https://github.com/yhirose/cpp-httplib/blob/4e6ded1f36ec24d6428ab560f8c0ddfb653651e6/test/meson.build#L9-L80

test/CMakeLists.txt
```diff
@@ -0,0 +1,93 @@
...
+
+set(BUILD_GMOCK OFF)
+set(INSTALL_GTEST OFF)
+set(gtest_force_shared_crt ON)
+
+FetchContent_Declare(
+    gtest
+    URL https://github.com/google/googletest/archive/main.tar.gz
+)
+FetchContent_MakeAvailable(gtest)
```
Tachi107, line 14: Instead of unconditionally downloading and building GoogleTest from source, could you please try to look for the library on the build machine with `find_package(GTest)` and fall back to FetchContent only if not found?

Tachi107 (commented):
Sorry for my late review, but I wasn't notified about this patch until the 0.12.0 release.

Mostly minor issues/suggestions

yhirose:
@jimmy-park could you replay to @Tachi107's comments and suggestions?

test/CMakeLists.txt
jimmy-park: I forgot the file command at that time. The same thing can be done with less code.

```CMake
file(
    COPY
    ${CMAKE_CURRENT_LIST_DIR}/www
    ${CMAKE_CURRENT_LIST_DIR}/www2
    ${CMAKE_CURRENT_LIST_DIR}/www3
    ${CMAKE_CURRENT_LIST_DIR}/ca-bundle.crt
    ${CMAKE_CURRENT_LIST_DIR}/image.jpg
    DESTINATION ${CMAKE_CURRENT_BINARY_DIR}
)
```

test/CMakeLists.txt
jimmy-park: I tested with `add_custom_command`, but it doesn't provide a pipeline between multiple commands. So i choose `execute_process` to mimic what `Makefile` does.

test/CMakeLists.txt
jimmy-park: Yes, fallback option would be nice.

```cmake
find_package(GTest)

if(NOT GTest_FOUND)
    if(POLICY CMP0135)
        cmake_policy(SET CMP0135 NEW)
    endif()

    include(FetchContent)
    include(GoogleTest)

    set(BUILD_GMOCK OFF)
    set(INSTALL_GTEST OFF)
    set(gtest_force_shared_crt ON)

    FetchContent_Declare(
        gtest
        URL https://github.com/google/googletest/archive/main.tar.gz
    )
    FetchContent_MakeAvailable(gtest)
endif()
```

jimmy-park:
@Tachi107 Thanks for suggestions. I commented my thoughts and some possible modifications.

@yhirose I think applying @Tachi107's suggestion is required. However, since this pull request had been merged, I will open a new pull request after resolving these conversations.

yhirose:
@jimmy-park yes, please go ahead to make another pull request for the adjustments. Thanks!

test/CMakeLists.txt
Tachi107: Yes, and also note that `${CMAKE_CURRENT_LIST_DIR}/` is unneeded since paths are already relative to the current source directory

test/CMakeLists.txt
Tachi107: Yeah this looks good

jimmy-park: It turns out that `FindGTest.cmake` will include the `GoogleTest` module also. So, `include(GoogleTest)` is actually redundant.

https://github.com/Kitware/CMake/blob/df52b85b0795404eec9e076ba4e6df7caaf40490/Modules/FindGTest.cmake#L102

test/CMakeLists.txt
jimmy-park: Not the actual shell pipe `|`, but CMake provides the functionality.
When we run multiple commands using `execute_process`, [they are pipelined](https://cmake.org/cmake/help/latest/command/execute_process.html#:~:text=Commands%20are%20executed%20concurrently%20as%20a%20pipeline%2C%20with%20the%20standard%20output%20of%20each%20process%20piped%20to%20the%20standard%20input%20of%20the%20next.%20A%20single%20standard%20error%20pipe%20is%20used%20for%20all%20processes.).

PR #1474, opened 2023-01-24
Support loading system certs from Keychein on MacOS

yerseg:
Added support for loading certificates from the Keychain on MacOS, similar to Windows.
I have tested it in my project and it works. 

I am not sure how to properly add the necessary MacOS SDK frameworks to cmake and make, I will be glad to comments.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -207,7 +207,7 @@ target_link_libraries(${PROJECT_NAME} ${_INTERFACE_OR_PUBLIC}
 		$<$<PLATFORM_ID:Windows>:crypt32>
 		$<$<PLATFORM_ID:Windows>:cryptui>
 		# Needed for API from MacOS Security framework
-		$<$<PLATFORM_ID:Darwin>:"-framework CoreFoundation -framework Security">
+		$<$<AND:$<PLATFORM_ID:Darwin>,$<BOOL:${HTTPLIB_IS_USING_OPENSSL}>>:"-framework CoreFoundation -framework Security">
 		# Can't put multiple targets in a single generator expression or it bugs out.
 		$<$<BOOL:${HTTPLIB_IS_USING_BROTLI}>:Brotli::common>
 		$<$<BOOL:${HTTPLIB_IS_USING_BROTLI}>:Brotli::encoder>
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -207,7 +207,7 @@ target_link_libraries(${PROJECT_NAME} ${_INTERFACE_OR_PUBLIC}
 		$<$<PLATFORM_ID:Windows>:crypt32>
 		$<$<PLATFORM_ID:Windows>:cryptui>
 		# Needed for API from MacOS Security framework
-		$<$<AND:$<PLATFORM_ID:Darwin>,$<BOOL:${HTTPLIB_IS_USING_OPENSSL}>>:"-framework CoreFoundation -framework Security">
+		"$<$<AND:$<PLATFORM_ID:Darwin>,$<BOOL:${HTTPLIB_IS_USING_OPENSSL}>>:-framework CoreFoundation -framework Security>"
 		# Can't put multiple targets in a single generator expression or it bugs out.
 		$<$<BOOL:${HTTPLIB_IS_USING_BROTLI}>:Brotli::common>
 		$<$<BOOL:${HTTPLIB_IS_USING_BROTLI}>:Brotli::encoder>
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -206,6 +206,8 @@ target_link_libraries(${PROJECT_NAME} ${_INTERFACE_OR_PUBLIC}
 		$<$<PLATFORM_ID:Windows>:ws2_32>
 		$<$<PLATFORM_ID:Windows>:crypt32>
 		$<$<PLATFORM_ID:Windows>:cryptui>
+		# Needed for API from MacOS Security framework
+		$<$<PLATFORM_ID:Darwin>:"-framework CoreFoundation -framework Security">
 		# Can't put multiple targets in a single generator expression or it bugs out.
 		$<$<BOOL:${HTTPLIB_IS_USING_BROTLI}>:Brotli::common>
 		$<$<BOOL:${HTTPLIB_IS_USING_BROTLI}>:Brotli::encoder>
--- a/httplib.h
+++ b/httplib.h
@@ -7928,7 +7928,7 @@ inline bool SSLClient::load_certs() {
         ret = false;
       }
     } else {
-      bool loaded = false;
+      auto loaded = false;
 #ifdef _WIN32
       loaded =
           detail::load_system_certs_on_windows(SSL_CTX_get_cert_store(ctx_));
--- a/httplib.h
+++ b/httplib.h
@@ -4426,7 +4426,7 @@ inline bool retrieve_certs_from_keychain(CFObjectPtr<CFArrayRef> &certs) {
   CFTypeRef values[] = {kSecClassCertificate, kSecMatchLimitAll,
                         kCFBooleanTrue};
 
-  CFObjectPtr<const __CFDictionary> query(
+  CFObjectPtr<CFDictionaryRef> query(
       CFDictionaryCreate(nullptr, reinterpret_cast<const void **>(keys), values,
                          sizeof(keys) / sizeof(keys[0]),
                          &kCFTypeDictionaryKeyCallBacks,
--- a/httplib.h
+++ b/httplib.h
@@ -4414,8 +4414,7 @@ inline bool load_system_certs_on_windows(X509_STORE *store) {
 
   return true;
 }
-#endif
-#ifdef __APPLE__
+#elif defined(__APPLE__)
 using CFObjectDeleter = std::function<void(CFTypeRef)>;
 template <typename T> using CFObjectPtr = std::unique_ptr<T, CFObjectDeleter>;
 
@@ -7904,8 +7903,7 @@ inline bool SSLClient::load_certs() {
       SSL_CTX_set_default_verify_paths(ctx_);
 #ifdef _WIN32
       detail::load_system_certs_on_windows(SSL_CTX_get_cert_store(ctx_));
-#endif
-#ifdef __APPLE__
+#elif defined(__APPLE__)
       detail::load_system_certs_on_apple(SSL_CTX_get_cert_store(ctx_));
 #endif
     }
--- a/httplib.h
+++ b/httplib.h
@@ -7898,12 +7898,14 @@ inline bool SSLClient::load_certs() {
         ret = false;
       }
     } else {
-      SSL_CTX_set_default_verify_paths(ctx_);
+      bool loaded = false;
 #ifdef _WIN32
-      detail::load_system_certs_on_windows(SSL_CTX_get_cert_store(ctx_));
+      loaded =
+          detail::load_system_certs_on_windows(SSL_CTX_get_cert_store(ctx_));
 #elif defined(__APPLE__)
-      detail::load_system_certs_on_apple(SSL_CTX_get_cert_store(ctx_));
+      loaded = detail::load_system_certs_on_apple(SSL_CTX_get_cert_store(ctx_));
 #endif
+      if (!loaded) { SSL_CTX_set_default_verify_paths(ctx_); }
     }
   });
 
--- a/httplib.h
+++ b/httplib.h
@@ -239,12 +239,10 @@ using socket_t = int;
 #pragma comment(lib, "crypt32.lib")
 #pragma comment(lib, "cryptui.lib")
 #endif
-#endif //_WIN32
-
-#ifdef __APPLE__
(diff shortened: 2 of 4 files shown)


Tachi107:
Could you please also update the meson.build files? It should be enough to add the following under the Brotli checks:

```meson
if host_machine.system() == 'darwin'
  deps += dependency('appleframeworks', modules : ['CoreFoundation', 'Security'])
endif
```

Thanks!

httplib.h
```diff
@@ -4410,7 +4415,69 @@ inline bool load_system_certs_on_windows(X509_STORE *store) {
   return true;
 }
 #endif
+#ifdef __APPLE__
```
Tachi107, line 4418: What about #elif?

yerseg: fixed
used `#elif` instead of a couple of `#ifdef`-s

yerseg:
@Tachi107 thank you for review! I added the necessary deps to meson.build as you said

httplib.h
```diff
@@ -4409,8 +4414,69 @@ inline bool load_system_certs_on_windows(X509_STORE *store) {
...
+template <typename T> using CFObjectPtr = std::unique_ptr<T, CFObjectDeleter>;
+
+inline bool load_system_certs_on_apple(X509_STORE *store) {
+  CFStringRef keys[] = {kSecClass, kSecMatchLimit, kSecReturnRef};
+  CFTypeRef values[] = {kSecClassCertificate, kSecMatchLimitAll,
+                        kCFBooleanTrue};
+
+  auto deleter = [](CFTypeRef p) {
+    if (p) CFRelease(p);
+  };
```
Tachi107, line 4428: A simple function reference should suffice, there's no need to use `std::function`:

```suggestion
static void CFObjectDeleter(CFTypeRef p) {
  if (p) CFRelease(p);
}
template <typename T> using CFObjectPtr = std::unique_ptr<T, decltype(CFObjectDeleter)*>;

inline bool load_system_certs_on_apple(X509_STORE *store) {
  CFStringRef keys[] = {kSecClass, kSecMatchLimit, kSecReturnRef};
  CFTypeRef values[] = {kSecClassCertificate, kSecMatchLimitAll,
                        kCFBooleanTrue};
```

Additionally, since the `p` is always checked and used before destruction, it cannot possibly be `NULL`, so you could drop the `CFObjectDeleter` function entirely and directly pass `CFRelease` to the `decltype()` call.

meson.build
```diff
@@ -58,6 +58,10 @@ if brotli_found_all
   args += '-DCPPHTTPLIB_BROTLI_SUPPORT'
 endif
 
+if host_machine.system() == 'darwin'
+  deps += dependency('appleframeworks', modules: ['CoreFoundation', 'Security'])
+endif
```
Tachi107, line 63: I didn't notice this earlier, but putting the `dependency()` call here would make the `CoreFoundation` and `Security` frameworks required even when not using building with TLS support. Could you please move these into the `if openssl_dep.found()` block? Thanks again :)

Tachi107 (commented):
Some minor feedback

httplib.h
```diff
@@ -241,6 +241,11 @@ using socket_t = int;
 #endif
 #endif //_WIN32
 
+#ifdef __APPLE__
```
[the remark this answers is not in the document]
yerseg: done

meson.build
yerseg: fixed

httplib.h
yerseg: I would prefer to keep the lambda, but I agree that we can not use std::function. A simple func ptr is enough.

Tachi107: Yeah, a lambda is cleaner, well done :)

httplib.h
```diff
@@ -7834,10 +7898,11 @@ inline bool SSLClient::load_certs() {
         ret = false;
       }
     } else {
+      SSL_CTX_set_default_verify_paths(ctx_);
```
yhirose, line 7901: This call should be in `#else` block. Please see the original code.

yerseg: I am not very familiar with openssl, but I have come to the following. 
If we fail to load certs from the system storage, then it would be correct to call SSL_CTX_set_default_verify_paths. Am I wrong?

yerseg: On Windows maybe it didn't matter, but for MacOS it has value at all I guess…

yhirose: Here is the link to `SSL_CTX_set_default_verify_paths`:
https://www.openssl.org/docs/man3.1/man3/SSL_CTX_set_default_verify_paths.html
image
So if we successfully load the OS certs, we shouldn't call `SSL_CTX_set_default_verify_paths`.

You probably could do as below:

```cpp
      auto loaded = false;
#ifdef _WIN32
      loaded = detail::load_system_certs_on_windows(SSL_CTX_get_cert_store(ctx_));
#elif defined(__APPLE__)
      loaded = detail::load_system_certs_on_apple(SSL_CTX_get_cert_store(ctx_));
#endif
      if (!loaded) { SSL_CTX_set_default_verify_paths(ctx_); }
```

This means, always `SSL_CTX_set_default_verify_paths` gets called on linux. As for Windows and MacOS, if the `detail::load_system_certs_on_???` fails, then `SSL_CTX_set_default_verify_paths` will be called as the last resort function.

yerseg: Yeah, you're right. But when we first use SSL_CTX_set_default_verify_paths, it means that if something screws up, the system will at least try to find certificates in the default places. As far as I can tell, adding extra certificates using X509_STORE_add_cert after setting the default paths shouldn't cause any issues.

yerseg: To put it another way, I believe that calling SSL_CTX_set_default_verify_paths first and then load_system_certs will either result in the same or an even better state than if load_system_certs were called first and then set_default_verify_paths.

yhirose: In the original code for Windows, we don't have to call `SSL_CTX_set_default_verify_paths` before/after `load_system_certs_on_windows`. I simply would like to do the same for MacOS since it is absolutely unnecessary to call `SSL_CTX_set_default_verify_paths` if `load_system_certs_on_apple` succeeds.

yerseg: ok, done

httplib.h
```diff
@@ -4409,8 +4412,69 @@ inline bool load_system_certs_on_windows(X509_STORE *store) {
 
   return true;
 }
+#elif defined(__APPLE__)
+template <typename T>
+using CFObjectPtr = std::unique_ptr<T, void (*)(CFTypeRef)>;
```
yhirose, line 4417: Can you move `CFObjectPtr` into `load_system_certs_on_apple`?

yerseg: It is template alias. It is not allowed to declare a template inside a function

yhirose:
@yerseg I have just approved the change, but the CI test on GitHub Actions reported a number of failures... Could you run the unit tests on your local machine by `cd test; make`? Thanks!

httplib.h
jimmy-park: If you really want to set the default path before loading certs, just call it after initializing `httplib::SSLClient`.

```cpp
httplib::SSLClient client;
SSL_CTX_set_default_verify_paths(client.ssl_context());
```

jimmy-park: If you really want to set the default path before loading certs, just call it after initializing httplib::SSLClient.

```cpp
httplib::SSLClient client;
SSL_CTX_set_default_verify_paths(client.ssl_context());
```

CMakeLists.txt
```diff
@@ -206,6 +206,8 @@ target_link_libraries(${PROJECT_NAME} ${_INTERFACE_OR_PUBLIC}
 		$<$<PLATFORM_ID:Windows>:ws2_32>
 		$<$<PLATFORM_ID:Windows>:crypt32>
 		$<$<PLATFORM_ID:Windows>:cryptui>
+		# Needed for API from MacOS Security framework
+		$<$<AND:$<PLATFORM_ID:Darwin>,$<BOOL:${HTTPLIB_IS_USING_OPENSSL}>>:"-framework CoreFoundation -framework Security">
```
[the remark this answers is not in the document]
yerseg: fixed it, thanks

yerseg:
@yhirose It turned out that there is another keychain from which we need to load certificates. Now all the tests pass.

httplib.h
```diff
@@ -4409,8 +4412,98 @@ inline bool load_system_certs_on_windows(X509_STORE *store) {
...
+      CFDictionaryCreate(nullptr, reinterpret_cast<const void **>(keys), values,
+                         sizeof(keys) / sizeof(keys[0]),
+                         &kCFTypeDictionaryKeyCallBacks,
+                         &kCFTypeDictionaryValueCallBacks),
+      cf_object_ptr_deleter);
+
+  if (!query) { return false; }
+
+  CFTypeRef security_items = nullptr;
+  OSStatus err = SecItemCopyMatching(query.get(), &security_items);
```
yhirose, line 4439: You can use `auto`.

httplib.h
```diff
@@ -4409,8 +4412,98 @@ inline bool load_system_certs_on_windows(X509_STORE *store) {
...
+    return false;
+  }
+
+  certs = std::move(security_items_ptr);
+  return true;
+}
+
+inline bool retrieve_root_certs_from_keychain(CFObjectPtr<CFArrayRef> &certs) {
+  CFArrayRef root_security_items = nullptr;
+  OSStatus err = SecTrustCopyAnchorCertificates(&root_security_items);
```
yhirose, line 4455: You can use `auto`.

httplib.h
```diff
@@ -4409,8 +4412,98 @@ inline bool load_system_certs_on_windows(X509_STORE *store) {
...
+                                                  cf_object_ptr_deleter);
+
+  if (err != errSecSuccess) { return false; }
+
+  certs = std::move(root_security_items_ptr);
+  return true;
+}
+
+inline void add_certs_to_x509_store(CFArrayRef certs, X509_STORE *store) {
+  OSStatus err = errSecSuccess;
```
yhirose, line 4466: You can use `auto`.

httplib.h
```diff
@@ -4409,8 +4412,98 @@ inline bool load_system_certs_on_windows(X509_STORE *store) {
...
+    const auto cert = reinterpret_cast<const __SecCertificate *>(
+        CFArrayGetValueAtIndex(certs, i));
+
+    if (SecCertificateGetTypeID() != CFGetTypeID(cert)) { continue; }
+
+    CFDataRef cert_data = nullptr;
+    err = SecItemExport(cert, kSecFormatX509Cert, 0, nullptr, &cert_data);
+    CFObjectPtr<CFDataRef> cert_data_ptr(cert_data, cf_object_ptr_deleter);
+
+    if (err != errSecSuccess) { continue; }
```
yhirose, line 4477: The above 4 lines can be as below:

```cpp
    CFDataRef cert_data = nullptr;
    if (SecItemExport(cert, kSecFormatX509Cert, 0, nullptr, &cert_data) != errSecSuccess) { continue; }

    CFObjectPtr<CFDataRef> cert_data_ptr(cert_data, cf_object_ptr_deleter);

```

httplib.h
```diff
@@ -4409,8 +4412,98 @@ inline bool load_system_certs_on_windows(X509_STORE *store) {
...
+      cf_object_ptr_deleter);
+
+  if (!query) { return false; }
+
+  CFTypeRef security_items = nullptr;
+  OSStatus err = SecItemCopyMatching(query.get(), &security_items);
+  CFObjectPtr<CFArrayRef> security_items_ptr(
+      reinterpret_cast<CFArrayRef>(security_items), cf_object_ptr_deleter);
+
+  if (err != errSecSuccess) { return false; }
```
yhirose, line 4443: This line should be moved above the previous line.

httplib.h
```diff
@@ -7834,11 +7927,14 @@ inline bool SSLClient::load_certs() {
         ret = false;
       }
     } else {
+      bool loaded = false;
```
yhirose, line 7930: You can use `auto`.

yerseg:
@yhirose fixed your comments

yhirose:
@yerseg, It has been merged. Thanks for your fine contribution!

yerseg:
@yhirose thank you too!

yhirose:
@yerseg please see #1522. it seems like a number of users on MacOS will encounter the link problem from now on. I am now thinking to make this feature 'opt in', since `SSL_CTX_set_default_verify_paths` works anyway with no problem on MacOS.  Or do you have any solution to reduce the expected user support caused by this? (On Windows, we embed `#pragma comment(lib, "ws2_32.lib")` in httplib.h, so that we can force the compiler to link the system library.) Thanks!

yerseg:
@yhirose I have been thinking about this problem, and I’m agree that we should make this feature optional. 
It will be great in case Unix toolchains provide some stuff like “#pragma comment lib” in MSVC, but we have not. 
So I will add the macro in new Pull request soon.

yhirose:
I added note for macOS on README for now. [d1b6162](https://github.com/yhirose/cpp-httplib/commit/d1b616286f0921efa9e1c5b8e619c9fd7078e348)

PR #1482, opened 2023-02-02
Update httplib.h

chenlimingcn:
capatables: this, sock, others variables is not neccessary.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -5596,7 +5596,7 @@ inline bool Server::listen_internal() {
 #if __cplusplus > 201703L
       task_queue->enqueue([=, this]() { process_and_close_socket(sock); });
 #else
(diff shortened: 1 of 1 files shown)


httplib.h
```diff
@@ -5596,7 +5596,7 @@ inline bool Server::listen_internal() {
 #if __cplusplus > 201703L
```
yhirose, line 5596: If your change is applied, the `#if __cplusplus > 201703L` block seems to be no longer needed. Could you make remove the block and check with C++17 compile option? Thanks!

yhirose:
Also could you rename the title to be more meaningful? Thanks!

yhirose:
@chenlimingcn any response?

PR #1493, opened 2023-02-14
Reuse gtest on system

pemensik:
Try to use gtest on the system if found.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -1,14 +1,16 @@
-cmake_policy(SET CMP0135 NEW)
-
-include(FetchContent)
-include(GoogleTest)
-
-set(BUILD_GMOCK OFF)
-set(INSTALL_GTEST OFF)
-set(gtest_force_shared_crt ON)
-
 find_package(GTest)
-if (NOT GTest_FOUND)
+
+if(NOT GTest_FOUND)
+    if(POLICY CMP0135)
+        cmake_policy(SET CMP0135 NEW)
+    endif()
+
+    include(FetchContent)
+
+    set(BUILD_GMOCK OFF)
+    set(INSTALL_GTEST OFF)
+    set(gtest_force_shared_crt ON)
+
     FetchContent_Declare(
         gtest
         URL https://github.com/google/googletest/archive/main.tar.gz
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -7,11 +7,14 @@ set(BUILD_GMOCK OFF)
 set(INSTALL_GTEST OFF)
 set(gtest_force_shared_crt ON)
 
-FetchContent_Declare(
-    gtest
-    URL https://github.com/google/googletest/archive/main.tar.gz
-)
-FetchContent_MakeAvailable(gtest)
+find_package(GTest)
+if (NOT GTest_FOUND)
+    FetchContent_Declare(
+        gtest
+        URL https://github.com/google/googletest/archive/main.tar.gz
+    )
+    FetchContent_MakeAvailable(gtest)
+endif()
 
 add_executable(httplib-test test.cc)
 target_compile_options(httplib-test PRIVATE "$<$<CXX_COMPILER_ID:MSVC>:/utf-8>")
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -19,7 +19,7 @@ if(NOT GTest_FOUND)
 endif()
 
 add_executable(httplib-test test.cc)
-target_compile_options(httplib-test PRIVATE "$<$<CXX_COMPILER_ID:MSVC>:/utf-8>")
+target_compile_options(httplib-test PRIVATE "$<$<CXX_COMPILER_ID:MSVC>:/utf-8;/bigobj>")
 target_link_libraries(httplib-test PRIVATE httplib GTest::gtest_main)
 gtest_discover_tests(httplib-test)
 
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -1,6 +1,12 @@
 find_package(GTest)
 
-if(NOT GTest_FOUND)
+if(GTest_FOUND OR GTEST_FOUND)
+    if(NOT TARGET GTest::gtest_main AND TARGET GTest::Main)
+	# CMake <3.20
+        add_library(GTest::gtest_main INTERFACE IMPORTED)
+        target_link_libraries(GTest::gtest_main INTERFACE GTest::Main)
+    endif()
+else()
     if(POLICY CMP0135)
         cmake_policy(SET CMP0135 NEW)
     endif()
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -1,6 +1,6 @@
 find_package(GTest)
 
-if(GTest_FOUND OR GTEST_FOUND)
+if(GTest_FOUND)
     if(NOT TARGET GTest::gtest_main AND TARGET GTest::Main)
 	# CMake <3.20
         add_library(GTest::gtest_main INTERFACE IMPORTED)


yhirose:
@sum01, @jimmy-park, could you review the change? It looks good to me, but I would like to check with CMake experts. :)

jimmy-park:
@pemensik `FetchContent_Declare(FIND_PACKAGE_ARGS)` will not work on older versions of CMake (it requires CMake 3.24+). Related suggestion https://github.com/yhirose/cpp-httplib/pull/1468#discussion_r1102561238 was mentioned by @Tachi107 recently. I think `find_package(GTest)` before `FetchContent` is more appropriate in this case. I will soon open a pull request for an updated `test/CMakeLists.txt` which addresses this issue also.

Tachi107:
@pemensik alternatively you could use Meson, which already uses the system-installed GTest version ;)

pemensik:
@jimmy-park feel free to push into my branch, I have allowed that. Pushed modification, is that what you meant?

@Tachi107 my project is using cmake, which requires cpp-httplib. If I build it with meson, I will not have cmake module to use. Just pkg-config module, which requires modification. It is unfortunate used build system causes built output to be different. Especially when both variants generate exclusive content and no build system generates both ways compatible with any of them.

Tachi107:
> @Tachi107 my project is using cmake, which requires cpp-httplib. If I build it with meson, I will not have cmake module to use. Just pkg-config module, which requires modification. It is unfortunate used build system causes built output to be different. Especially when both variants generate exclusive content and no build system generates both ways compatible with any of them.

It's not that hard to generate a CMake Config file from Meson, but really the issue is that CMake's `find_package()` does not provide a way to automatically use .pc files, which are the closest thing we have to a standard for C and C++ dependency discovery. I'll eventually send a patch to add this extra functionality to the meson.build file, but in the mean time I'd suggest you to just use drop [this Find Module](https://github.com/yuzu-emu/yuzu/blob/770a49616d9f9772ffd6c1648a25da4abd4338e2/CMakeModules/Findhttplib.cmake) in your CMake project; it tries to find cpp-httplib using a CMake Config file, and falls back to pkg-config otherwise.

jimmy-park:
@pemensik yes, that's what I thought. I will add a small change to move some variables.

jimmy-park:
@pemensik I couldn't push to this branch because of permission denial. Could you please check it again? or my git settings have some problem.

pemensik:
*Allow edits and access to secrets by maintainers* is checked. Not sure how it is supposed to work. Do you have write access into this repository?

Tachi107:
No, yhirose is the only one with write access. Maybe you can give @jimmy-park write access to your fork, though

sum01:
I've never used GTest with Cmake, but just at a glance, the docs seem to specify different variables based on the version.

For example:
| v3.14 | Latest release (v3.26) |
| --- | --- |
| `GTEST_FOUND` | `GTest_FOUND` | 
| `GTest::Main` | `GTest::Main` is deprecated, now uses `GTest::gtest_main` |

Even though this is kind of silly, won't this cause some issues if the builder has an older Cmake version?

jimmy-park:
@sum01 hmm.. apparently they deprecated those targets since version `3.20`. I'm not sure this workaround works on older versions.

```cmake
find_package(GTest)

if(GTest_FOUND OR GTEST_FOUND)
    if(TARGET GTest::Main AND NOT TARGET GTest::gtest_main)
        add_library(GTest::gtest_main ALIAS GTest::Main)
    endif()
else()
    # FetchContent routine
endif()
```

eli-schwartz:
What is wrong with:
```
include(FindPkgConfig)
pkg_search_module(HTTPLIB IMPORTED_TARGET GLOBAL cpp-httplib)
```

Then simply link to `PkgConfig::HTTPLIB`.

sum01:
While PkgConf is nice, I guess since it's not commonly installed on things like Windows (and maybe macOS?) it Just requires more workarounds. That's why the built-in modules that come with Cmake are useful.

Just more things to keep in mind stemming from trying to do multi-platform Cmake I suppose.

eli-schwartz:
Well, at the very least it should be a solution for fedora... For @pemensik's cmake project which depends on httplib, for example.

Although really, cmake isn't commonly installed either -- but people who need it, install it, and the same applies to pkg-config. It's trivially available in macOS via e.g. homebrew, and automatically wrapped on Windows by vcpkg.

pemensik:
I think the option would be to generate also pkg-config file from cmake build. For example by *configure_file* cmake statement and a template. That would make pkg-config generated always, maybe just with minor differences. I think generating cmake config from Meson would be more difficult. But at least one way present always would be useful.

But the main topic of this PR is different, maybe separate issue should be created for cmake module vs pkg-config file and discuss it there.

test/CMakeLists.txt
```diff
@@ -1,14 +1,16 @@
...
+if(NOT GTest_FOUND)
+    if(POLICY CMP0135)
+        cmake_policy(SET CMP0135 NEW)
+    endif()
 
-set(BUILD_GMOCK OFF)
-set(INSTALL_GTEST OFF)
-set(gtest_force_shared_crt ON)
+    include(FetchContent)
+
```
pemensik, line 9: Where is ``include(GoogleTest)``? Isn't it required later?

jimmy-park: `find_package(GTest)` will include `GoogleTest` as shown in [FindGTest.cmake](https://github.com/Kitware/CMake/blob/df52b85b0795404eec9e076ba4e6df7caaf40490/Modules/FindGTest.cmake#L102).

pemensik: Ah, sure, makes sense to not require it manually.

sum01:
> Although really, cmake isn't commonly installed either -- but people who need it, install it, and the same applies to pkg-config. It's trivially available in macOS via e.g. homebrew, and automatically wrapped on Windows by vcpkg.

@eli-schwartz 

Well the main difference is you need Cmake to build with Cmake, which means the built-ins are guarnteed to be available, whereas adding PkgConf to the mix means you need to also make sure the user has that too. Just an unnecessary complication if Cmake already has a module to use.

---

@pemensik I'd suggest implementing some form of this https://github.com/yhirose/cpp-httplib/pull/1493#issuecomment-1430577511 to avoid the issues I mentioned in https://github.com/yhirose/cpp-httplib/pull/1493#issuecomment-1430247102

I'm also not sure trying to generate PkgConf files with Cmake is a good idea, if the internet is to be believed ([see here](https://stackoverflow.com/questions/44292462/how-to-auto-generate-pkgconfig-files-from-cmake-targets)) considering the loss of information vs Cmake. You'd probably have better luck writing your own PkgConf file than generating one.

Tachi107:
@sum01 `GTest::Main` is guaranteed to always be available when using `find_package(GTest)` for compatibility reasons (it's defined by CMake's FindGTest built-in module before calling the upstream GTestConfig)

sum01:
> @sum01 `GTest::Main` is guaranteed to always be available when using `find_package(GTest)` for compatibility reasons (it's defined by CMake's FindGTest built-in module before calling the upstream GTestConfig)

My impression was that since it's listed as "deprecated since v3.20" it'll eventually disappear, and OP may as well take action now instead of waiting for vX.Y.Z when it's gone.

eli-schwartz:
> I'm also not sure trying to generate PkgConf files with Cmake is a good idea, if the internet is to be believed ([see here](https://stackoverflow.com/questions/44292462/how-to-auto-generate-pkgconfig-files-from-cmake-targets)) considering the loss of information vs Cmake. 

What loss of information? The information was never present in cmake to begin with -- cmake doesn't even try to detect e.g. openssl, zlib, brotli via pkg-config in order to write proper "requires" metadata so one pkg-config file can depend on another pkg-config file.

And cmake itself is doing considerably worse and dropping information ***all*** over the place. The cmake-generated export Targets file provides INTERFACE_LINK_LIBRARIES for ZLIB::ZLIB, OpenSSL::SSL, OpenSSL::Crypto, Brotli::encoder, Brotli::decoder, etc. -- but never *defines* them, which is why httplibConfig.cmake.in has to go to some very verbose double-handling of this.

With pkg-config it's very simple, try to find the pkg-config dependency "foo", if it is found then link to PkgConfig::FOO and add "foo" to a list `pc_requires`. Format the pkg-config file with `Requires: @pc_requires@`. It's the same information in both places.

If you found a dependency using find_package() instead of pkg_check_modules() then you need to write out its `-lfoo` flags instead.

> You'd probably have better luck writing your own PkgConf file than generating one.

No, you absolutely want to sync it with cmake's found packages state, which means encoding the information about what to include in the pkg-config file as CMakeLists.txt logic.

test/CMakeLists.txt
```diff
@@ -1,14 +1,16 @@
...
+    set(BUILD_GMOCK OFF)
+    set(INSTALL_GTEST OFF)
+    set(gtest_force_shared_crt ON)
```
pemensik, line 12: Why is this variable set? If gtest is not installed, does it make sense to explicitly use shared runtime? Haven't found why it is useful.

test/CMakeLists.txt
```diff
@@ -1,14 +1,16 @@
-cmake_policy(SET CMP0135 NEW)
+find_package(GTest)
 
-include(FetchContent)
-include(GoogleTest)
+if(NOT GTest_FOUND)
```
pemensik, line 3: Seems better, thanks.

test/CMakeLists.txt
```diff
@@ -19,7 +19,7 @@ if(NOT GTest_FOUND)
 endif()
 
 add_executable(httplib-test test.cc)
-target_compile_options(httplib-test PRIVATE "$<$<CXX_COMPILER_ID:MSVC>:/utf-8>")
+target_compile_options(httplib-test PRIVATE "$<$<CXX_COMPILER_ID:MSVC>:/utf-8;/bigobj>")
```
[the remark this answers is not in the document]
jimmy-park: I tested on Windows and got a specific error about [`/bigobj`](https://learn.microsoft.com/en-us/cpp/build/reference/bigobj-increase-number-of-sections-in-dot-obj-file?view=msvc-170). `test.cc` exceeded the threshold at some point. I don't think it is related to this issue, but I had to set the option to compile.

test/CMakeLists.txt
jimmy-park: Unless we set `gtest_force_shared_crt` to `ON`, GoogleTest's CMake proejct will override compiler options that cause an linker error on Windows. Please see [googletest/README.md](https://github.com/google/googletest/blob/main/googletest/README.md#visual-studio-dynamic-vs-static-runtimes) and [googletest/internal_utils.cmake](https://github.com/google/googletest/blob/f063cd25c90cbd4089a0ff96f5991df4f2721338/googletest/cmake/internal_utils.cmake#L34-L47).

yhirose:
@pemensik, what is the status of this pull request? It seems like two comments from @jimmy-park are not resolved, yet.

yhirose:
@pemensik, @jimmy-park, can I merge this pull request, or should I close it? Please let me know.

test/CMakeLists.txt
pemensik: Okay, it might be set only for windows in that case. Or at least had comment referencing this is MSVC and Windows specific issue. ``if (MSVC) ...`` should be used if it does not make sense for any other platform. But not important for this PR.

jimmy-park:
> @sum01 hmm.. apparently they deprecated those targets since version `3.20`. I'm not sure this workaround works on older versions.
> 
> ```cmake
> find_package(GTest)
> 
> if(GTest_FOUND OR GTEST_FOUND)
>     if(TARGET GTest::Main AND NOT TARGET GTest::gtest_main)
>         add_library(GTest::gtest_main INTERFACE IMPORTED)
>         target_link_libraries(GTest::gtest_main INTERFACE GTest::Main)
>     endif()
> else()
>     # FetchContent routine
> endif()
> ```

@pemensik How about applying this workaround? I think https://github.com/yhirose/cpp-httplib/pull/1493#issuecomment-1430247102 should be handled in this pull request for the backward compatibility.

yhirose:
@pemensik, please let me know when you apply the above @jimmy-park's comment.

pemensik:
Is that forward or backward compatibility? I have looked into our gtest package for RHEL8, version 1.8.0 does not yet have any cmake module. RHEL9 version 1.11.0 does have only gtest_main target. Latest release on https://github.com/google/googletest is 1.13.0. The same is in Fedora Rawhide and it produces still only ``GTest::gtest_main`` target. I haven't found GTest::Main in the whole google organization. Can you provide link to version without GTest::gtest_main target, @jimmy-park ?

pemensik:
Oh, found that, it is part of cmake-data package for older versions. RHEL8 has version barely required for modern targets, RHEL7 does not even parse main CMakeLists.txt. So RHEL is not a good distribution to test the legacy target. Okay, it may not hurt providing additional target.

test/CMakeLists.txt
```diff
@@ -1,20 +1,31 @@
-cmake_policy(SET CMP0135 NEW)
+find_package(GTest)
 
-include(FetchContent)
-include(GoogleTest)
+if(GTest_FOUND OR GTEST_FOUND)
```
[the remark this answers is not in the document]
jimmy-park: Yes, I know that's true. But I was confused when I saw the older version of `FindGTest` module. It uses `GTEST_FOUND` and its document says the same thing. So I'm not sure this is redundant or not.

- https://cmake.org/cmake/help/v3.19/module/FindGTest.html
- https://github.com/Kitware/CMake/blob/c9b27dace7e4110b25d5889e9c6952d0f4a7ca58/Modules/FindGTest.cmake#L204-L211

test/CMakeLists.txt
```diff
@@ -1,20 +1,31 @@
...
+    if(NOT TARGET GTest::gtest_main AND TARGET GTest::Main)
+	# CMake <3.20
+        add_library(GTest::gtest_main INTERFACE IMPORTED)
+        target_link_libraries(GTest::gtest_main INTERFACE GTest::Main)
```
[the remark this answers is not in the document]
jimmy-park: Unfortunately, before CMake 3.18, we can not use `ALIAS` target for `GTest::Main` that is a non-global `IMPORTED` library target.

https://cmake.org/cmake/help/latest/command/add_library.html#alias-libraries:~:text=New%20in%20version%203.18%3A%20An%20ALIAS%20can%20target%20a%20non%2DGLOBAL%20Imported%20Target.%20Such%20alias%20is%20scoped%20to%20the%20directory%20in%20which%20it%20is%20created%20and%20below.%20The%20ALIAS_GLOBAL%20target%20property%20can%20be%20used%20to%20check%20if%20the%20alias%20is%20global%20or%20not.

test/CMakeLists.txt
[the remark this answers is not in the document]
Tachi107: Yeah, the module may be setting `GTEST_FOUND`, but I was specifically talking about `find_package()` - its documentation says:

> Regardless of the mode used, a `<PackageName>_FOUND` variable will be set to indicate whether the package was found.

So, even if the FindGTest module doesn't set `GTest_FOUND`, `find_package()` itself does.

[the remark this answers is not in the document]
jimmy-park: I tested under CMake 3.19 and it seems like both variables are set correctly.
```cmake
cmake_minimum_required(VERSION 3.19)

project(test)

find_package(GTest)

message(STATUS "GTest_FOUND : ${GTest_FOUND}")
message(STATUS "GTEST_FOUND : ${GTEST_FOUND}")
```

Result

result

So `GTEST_FOUND` is redundant for sure.

```suggestion
if(GTest_FOUND)
```

yhirose:
It's pretty hard for me to keep track of this thread... So is it now ready for merge?

jimmy-park:
@yhirose Yes, I think it is okay to merge. I have no further discussion about this pull request.

yhirose:
Thanks all!

PR #1499, opened 2023-02-18
Threadpool: don't allow enqueueing after shutdown.

Xiaotian0726:
A threadpool should ignore all the enqueueing tasks after its shutdown rather than pushing them into the `job_` list.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -559,6 +559,10 @@ public:
   void enqueue(std::function<void()> fn) override {
     {
       std::unique_lock<std::mutex> lock(mutex_);
+
+      // Don't allow enqueueing after shutdown
+      if (shutdown_) { return; }
+
       jobs_.push_back(std::move(fn));
     }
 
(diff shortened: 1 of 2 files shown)


yhirose:
@Xiaotian0726, thanks for the pull request. Could you explain more about this change, and provide at least one unit test in `test/test/cc` for this issue? Thanks!

Xiaotian0726:
> @Xiaotian0726, thanks for the pull request. Could you explain more about this change, and provide at least one unit test in `test/test/cc` for this issue? Thanks!

@yhirose , thanks for you reply. The explanation has been updated and a new PR of unit test is on.

yhirose:
@Xiaotian0726, thanks for adding the unit test, however, the unit test didn't fail when I tested it with the original httplib.h without your change...

Xiaotian0726:
> @Xiaotian0726, thanks for adding the unit test, however, the unit test didn't fail when I tested it with the original httplib.h without your change...
> 
> image

Thank you for your testing, I'll check this again later...

PR #1504, opened 2023-02-24
cmake: support components

abouvier:
This will allow things like `find_package(httplib REQUIRED OpenSSL)` to reject library not supporting SSL functionalities, and prevent the creation of the `httplib::httplib` imported target when this is the case.

It's shorter than:
```
find_package(httplib REQUIRED)
if(NOT HTTPLIB_IS_USING_OPENSSL)
  message(FATAL_ERROR "...")
endif()
```
And it allows the creation of a fallback imported target (e.g. from bundled source code) with the same name than `httplib::httplib`.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,8 +14,8 @@
 
 	-------------------------------------------------------------------------------
 
-	After installation with Cmake, a find_package(httplib) is available.
-	This creates a httplib::httplib target (if found).
+	After installation with Cmake, a find_package(httplib COMPONENTS OpenSSL ZLIB Brotli) is available.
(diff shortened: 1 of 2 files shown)


sum01 (changes requested):
Looks mostly good, just would like some documentation at the top of the main `CMakeLists.txt` file to show the three components we have available. Maybe something as simple as `find_package(httplib COMPONENTS OpenSSL ZLIB Brotli REQUIRED)` with a short list of the possible components or whatever.

yhirose:
@abouvier, @sum01, thanks for your contribution! I have just merged it.

PR #1509, opened 2023-02-28
cmake: fix find_dependency

abouvier:
`QUIET`/`REQUIRED` are not recommended in `find_dependency` because they will be added by CMake based on the keyword used in `find_package(httplib ...)`.

For example currently `find_package(httplib QUIET COMPONENTS OpenSSL)` will produce a fatal error if OpenSSL is not found instead of failing silently (httplib is not a required dependency in this case).


yhirose:
@sum01 does this change look good to you? If so, I'll merge it.

sum01 (approved):

yhirose:
Thanks for the contribution!

PR #1512, opened 2023-03-02
Modify where post_routing_handler_ is called.

zhaohuijun:
If res is modified in function post_routing_handler_, the original position causes the header to be incorrect.


yhirose:
@zhaohuijun, thanks for the pull request. As I always ask any contributors, could you provide at least one unit test which verifies your change is correct. Then, I'll review it. Thanks for your understanding!

yhirose:
@zhaohuijun any response?

yhirose:
@zhaohuijun it took a look at your code, but I don't agree the change since the post request handler should be called with a fully prepared response object.

Issue #1515, opened 2023-03-07
parse_multipart_boundary() fails if Content Type header contains a charset attribute


mgaillard:
When processing a multipart request or response, the following function is called to detect the boundary.
`bool parse_multipart_boundary(const std::string &content_type, std::string &boundary)`
I do not know if this is standard or not but I receive the following header:
`multipart/mixed; boundary=Boundary_69003_722396156_1678219376089;charset=UTF-8`
When using the parse_multipart_boundary function, it currently gets the charset as part of the boundary.
`boundary == "Boundary_69003_722396156_1678219376089;charset=UTF-8"`
For reference, I found some documentation online:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type

PR #1516, opened 2023-03-07
Fix multipart Content-Type headers with both boundary and charset parameters

mgaillard:
This PR is for Issue #1515.

If the webserver returns a multipart response and include both boundary and charset parameters in the Content-Type header, then only the boundary needs to be read by `parse_multipart_boundary()`. I am not sure that including boundary + charset is something that follows the standard, but when querying an _openresty_ server, we get this case.
I tried my best to read the RFC standards and do something that works in the general case. According to section [5.1.1](https://www.rfc-editor.org/rfc/rfc2046.html#section-5.1.1) of the RFC2046, the boundary cannot contain a semicolon, so I just search for the first semicolon after the "boundary=" keyword and cut the substring at this position. I added four tests to check that it works.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -1823,6 +1823,8 @@ std::string params_to_query_str(const Params &params);
 
 void parse_query_text(const std::string &s, Params &params);
 
+bool parse_multipart_boundary(const std::string &content_type, std::string &boundary);
+
 bool parse_range_header(const std::string &s, Ranges &ranges);
 
 int close_socket(socket_t sock);
--- a/httplib.h
+++ b/httplib.h
@@ -3888,10 +3888,12 @@ inline void parse_query_text(const std::string &s, Params &params) {
 
 inline bool parse_multipart_boundary(const std::string &content_type,
                                      std::string &boundary) {
-  auto pos = content_type.find("boundary=");
+  auto boundary_keyword = "boundary=";
+  auto pos = content_type.find(boundary_keyword);
   if (pos == std::string::npos) { return false; }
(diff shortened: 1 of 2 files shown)


httplib.h
```diff
@@ -3890,7 +3890,8 @@ inline bool parse_multipart_boundary(const std::string &content_type,
                                      std::string &boundary) {
   auto pos = content_type.find("boundary=");
   if (pos == std::string::npos) { return false; }
-  boundary = content_type.substr(pos + 9);
+  auto end = content_type.find(';', pos);
+  boundary = content_type.substr(pos + 9, end - pos - 9);
```
yhirose, line 3894: Could it be easier to understand?
```c++
auto beg = pos + 9;
boundary = content_type.substr(beg, end - beg);
```

mgaillard: Yes, totally. I agree. I will make the modification tomorrow. Thanks for the suggestion.

mgaillard: Done, and pushed on my branch.

yhirose:
@mgaillard could you please take a look at some compile errors on GitHub Actions CI (ubuntu-latest)? Thanks!

mgaillard:
I could not reproduce the problem on my side, but I think it has to do with the fact that the build is split into .h and .cpp maybe? I noticed that there was no forward declaration for `parse_multipart_boundary()` so I added it in hope that it would help.

yhirose:
@mgaillard thanks for the fine contribution!

PR #1517, opened 2023-03-08
Fix typos

jrosdahl:


yhirose:
@jrosdahl thanks for catching the embarrassing typos!

Issue #1250, opened 2022-04-16
Can I rename an image uploaded by httplib? After studying for a long time, I still don't know what to do. Looking forward to your reply


Haragarden:

yhirose:
Thanks for the feedback, but it's not clear to me...

Haragarden:
#include "paddle_api.h"
#include "httplib.h"
#include "json.h"
#include <arm_neon.h&gt;
#include <stdio.h&gt;
#include <sys/time.h&gt;
#include <unistd.h&gt;
#include <vector&gt;
#include <limits&gt;
#include <iostream&gt;
#include <fstream&gt;
#include <signal.h&gt;
#include <sys/stat.h&gt;


//using namespace std;
using namespace httplib;
using namespace paddle::lite_api;




const char* html = R"(
&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;<html&gt;
&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;<form id="formElem"&gt;
&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;<input type="file" name="image_file" accept="image/*"&gt;<br&gt;
&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;<input type="submit"&gt;
&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;</form&gt;
&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;<script&gt;
&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;formElem.onsubmit = async (e) =&gt; {
&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;e.preventDefault();
&nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp; &nbsp;let res = await fetch('/post', {
[...]

Issue #1212, opened 2022-03-15
upload unicode file name bug


eurake:
# problem
If you upload using a Unicode file name, you cannot get the correct file name and file content
# code 
```
svr.Post("/uploadFile", [&](const auto& req, auto& res) {
    std::string path = boost::dll::program_location().parent_path().string();
    auto imgFile = req.get_file_value("imgFile");

    // log  
    BOOST_LOG_TRIVIAL(fatal) << "file.filename: " << imgFile.filename;
    BOOST_LOG_TRIVIAL(fatal) << "file.content_type: " << imgFile.content_type;
    BOOST_LOG_TRIVIAL(fatal) << "file.name: " << imgFile.name;
    BOOST_LOG_TRIVIAL(fatal) << "file.content: " << imgFile.content.length();

    // generate file name
    boost::uuids::uuid uid = boost::uuids::random_generator()();
    std::string tmpUid = boost::uuids::to_string(uid);
    boost::filesystem::path imgPath(path + "\\upload\\" + tmpUid + ".jpg");

    // write to file 
    std::ofstream  fStream(imgPath.string(), std::ios::binary);
    fStream << imgFile.content;
    fStream.close();

    // return json 
    nlohmann::json jRes;
    jRes["id"] = 1;
    jRes["msg"] = "客户测试1112";
    res.set_content(jRes.dump(), "application/json");
}
```
# used postman to request 
http://127.0.0.1:12345/api/v1/uploadFile
use body => form data => key="imgFile" and Value="1test.jpg"
Can get the picture correctly in http server side

use body => form data => key="imgFile" and Value="测试test.jpg"  
Can‘t get the picture correctly  in http server side
[results are as follows]

# log 
```
[2022-03-15 08:45:50.726725][fatal]:  file.filename: 
[2022-03-15 08:45:50.726725][fatal]:  file.content_type: 
[2022-03-15 08:45:50.726725][fatal]:  file.name: 
[2022-03-15 08:45:50.726725][fatal]:  file.content: 0
```
# analysis
in cpp-httplib.h file line: 3794 
std::regex_match(header, m, re_content_disposition)
can get correctly name and filename 
```
          static const std::string header_name = "content-type:";
          const auto header = buf_head(pos);
          if (start_with_case_ignore(header, header_name)) {
            file_.content_type = trim_copy(header.substr(header_name.size()));
          } else {
            std::smatch m;
            if (std::regex_match(header, m, re_content_disposition)) {
[...]

PR #1519, opened 2023-03-10
Improvement to multipart parser

mgaillard:
I noticed the parsing of multipart forms had two issues:
1. The regex would fail if attributes are not in the following order: name, filename, filename*
2. Filename* attributes are not supported

This PR is my attempt at solving the problem. I noticed two issues are related to this piece of code: #1250 and #1212.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -4366,7 +4366,7 @@ private:
   size_t buf_epos_ = 0;
 };
 
-bool parse_multipart_response(const Response &res, MultipartFormDataItems& files)
+inline bool parse_multipart_response(const Response &res, MultipartFormDataItems& files)
 {
   if (!res.is_multipart_form_data()) {
     return false;
--- a/httplib.h
+++ b/httplib.h
@@ -1903,6 +1903,8 @@ void read_file(const std::string &path, std::string &out);
 
 std::string trim_copy(const std::string &s);
 
+std::string trim_double_quotes(const std::string &s);
+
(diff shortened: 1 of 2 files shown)


mgaillard:
I see that the build is failing because my tests are not compatible with the .h .cpp split script. I will look at it tomorrow.

mgaillard:
When running split.py the `class MultipartFormDataParser` ends up in the httplib.cc file, which is not visible in the test.cc
For writing tests it is much easier to access directly MultipartFormDataParser instead of going through a webserver, etc.
Would it be fine to split declaration and definition of MultipartFormDataParser so that it becomes public?

mgaillard:
I added:
```
bool parse_multipart_response(const Response &res, MultipartFormDataItems& files);
```
To allow testing the multipart form data parser without making it public.
It also allows the client code to parse multipart responses from the server.

mgaillard:
I fixed the code so that tests pass, could I get a review?

yhirose:
@mgaillard I fixed this issue in a simpler way. Could you try the latest httplib.h in the master branch? Thanks!

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition
https://httpwg.org/specs/rfc6266.html#header.field.definition
https://www.rfc-editor.org/rfc/rfc5987.html#section-3.2
https://stackoverflow.com/questions/93551/how-to-encode-the-filename-parameter-of-content-disposition-header-in-http

PR #1524, opened 2023-03-13
fix put interface with params

yushu20171007:
The default put interface with params sets params to body.


yhirose:
@yushu20171007 could you explain more about your pull request such as why it's needed? Also, could you add at least one unit text in `test/test.cc`? Then, I'll review this pull request. Thanks!

yushu20171007:
@yhirose Sorry, I checked it carefully again, there is a mistake here. It's my way of use (/ω＼)

yushu20171007:
08e4155c-f7aa-4dd8-a75e-ff7f961183fd

PR #1525, opened 2023-03-13
Make the certs loading from the Apple Keychain feature optional

yerseg:
@yhirose as you proposed, it's makes the feature optional with a macro

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -35,7 +35,11 @@ if openssl_dep.found()
   deps += openssl_dep
   args += '-DCPPHTTPLIB_OPENSSL_SUPPORT'
   if host_machine.system() == 'darwin'
-    deps += dependency('appleframeworks', modules: ['CoreFoundation', 'Security'])
+    macosx_keychain_dep = dependency('appleframeworks', modules: ['CoreFoundation', 'Security'], required: get_option('cpp-httplib_macosx_keychain'))
+    if macosx_keychain_dep.found()
+        deps += macosx_keychain_dep
+        args += '-DCPPHTTPLIB_USE_CERTS_FROM_MACOSX_KEYCHAIN'
+    endif  
   endif
 endif
 
--- a/meson_options.txt
+++ b/meson_options.txt
@@ -5,5 +5,6 @@
 option('cpp-httplib_openssl', type: 'feature', value: 'auto', description: 'Enable OpenSSL support')
 option('cpp-httplib_zlib',    type: 'feature', value: 'auto', description: 'Enable zlib support')
 option('cpp-httplib_brotli',  type: 'feature', value: 'auto', description: 'Enable Brotli support')
+option('cpp-httplib_macosx_keychain', type: 'feature', value: 'auto', description: 'Enable loading certs from the Keychain on Apple devices')
 option('cpp-httplib_compile', type: 'boolean', value: false,  description: 'Split the header into a compilable header & source file (requires python3)')
 option('cpp-httplib_test',    type: 'boolean', value: false,  description: 'Build tests')
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,6 +6,7 @@
 	* HTTPLIB_REQUIRE_OPENSSL (default off)
 	* HTTPLIB_REQUIRE_ZLIB (default off)
 	* HTTPLIB_USE_BROTLI_IF_AVAILABLE (default on)
+	* HTTPLIB_USE_CERTS_FROM_MACOSX_KEYCHAIN (default on)
 	* HTTPLIB_REQUIRE_BROTLI (default off)
 	* HTTPLIB_COMPILE (default off)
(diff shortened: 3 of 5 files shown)


meson.build
```diff
@@ -34,7 +34,8 @@ openssl_dep = dependency('openssl', version: '>=1.1.1', required: get_option('cp
 if openssl_dep.found()
   deps += openssl_dep
   args += '-DCPPHTTPLIB_OPENSSL_SUPPORT'
-  if host_machine.system() == 'darwin'
+  if get_option('cpp-httplib_apple_keychain') and host_machine.system() == 'darwin'
+    args += '-DCPPHTTPLIB_USE_CERTS_FROM_APPLE_KEYCHAIN'
```
Tachi107, line 38: I would keep the Meson (and CMake) code as is, as these libraries are always available on macOS (right?) and linking to them is only an issue when using the header without any build system handling the compilation

meson_options.txt
```diff
@@ -5,5 +5,6 @@
 option('cpp-httplib_openssl', type: 'feature', value: 'auto', description: 'Enable OpenSSL support')
 option('cpp-httplib_zlib',    type: 'feature', value: 'auto', description: 'Enable zlib support')
 option('cpp-httplib_brotli',  type: 'feature', value: 'auto', description: 'Enable Brotli support')
+option('cpp-httplib_apple_keychain', type: 'feature', value: false, description: 'Enable loading certs from the Keychain on Apple devices')
```
Tachi107, line 8: This option is a `boolean`, but here you're using a `feature` which has three states (enabled, disabled, and auto). If you want to keep the option, please correct the type.

yerseg: Fixed it

meson.build
yerseg: I think when user use build system it was more correct to implicitly enable this feature via build system. So when I use cmake, HTTPLIB_IS_USING_CERTS_FROM_APPLE_KEYCHAIN feature wiil be enable implicitly. 
But when using header-only without any build scripts library user can be explicitly specify that the feature is required.
Finnaly I think that build system should also provide this option anyway.

yerseg: So it is strange when linker gets additional flags without any semantic artifacts in the build script

meson.build
```diff
@@ -34,7 +34,8 @@ openssl_dep = dependency('openssl', version: '>=1.1.1', required: get_option('cp
...
     deps += dependency('appleframeworks', modules: ['CoreFoundation', 'Security'])
   endif
```
Tachi107, line 40: > So it is strange when linker gets additional flags without any semantic artifacts in the build script

I'm not sure what you mean by this, but if I understand correctly then no, it is not strange to me. It is perfectly reasonable for header-only libraries to depend on third party libraries, and in that case the build system should add the required compiler and linker flags to the interface of the header-only library, so that consuming the header from another project doesn't result in linker errors. Did I misunderstand your comment?

Anyway, thinking a bit more about this I think that it'd be better to fully make use of Meson's `feature` type just like what we already do with OpenSSL, Zlib, etc. I'd suggest something like this:

```suggestion
  if host_machine.system() == 'darwin'
    apple_keychain_dep = dependency('appleframeworks', modules: ['CoreFoundation', 'Security'], required: get_option('cpp-httplib_apple_keychain'))
    if apple_keychain_dep.found()
        deps += apple_keychain_dep
        args += '-DCPPHTTPLIB_USE_CERTS_FROM_APPLE_KEYCHAIN'
    endif  
  endif
```

yerseg: > I would keep the Meson (and CMake) code as is, as these libraries are always available on macOS (right?)

I just meant that it seems reasonable to add optionality for the feature both in the header using a macro and in build systems. Nothing else.

Tachi107:
> I just meant that it seems reasonable to add optionality for the feature both in the header using a macro and in build systems. Nothing else.

I get your point, but also consider that cpp-httplib has [a lot of options](https://github.com/yhirose/cpp-httplib/blob/8f96b69a258a227796c907c23a73b34e03d6f67c/httplib.h#L13-L168), and only a small subset of them is available in the build systems, for a few reasons. Having more options means that:

1. Someone has to maintain some more code (and since I maintain the Meson portion of the library, that someone is me)
2. The interface is larger, and users need to read more before knowing which option they need

Given the above, I tried to explain why I was against the idea of exposing the option in Meson and CMake. Its use case is a niche (only macOS users), and doesn't make much sense in a build system (since the user doesn't have to mess with build flags, and linking to an extra library which is always available does not create issues).

That being said, I think this patch looks good as is, so regardless of whether you decide to keep the option or not, this looks good to me, and I personally approve it!

Tachi107 (approved):

yhirose:
@yerseg, could you resolve conflicts caused by #1546? Thanks!

yerseg:
@yhirose done. Thanks for review!

PR #1529, opened 2023-03-16
Prevent overflow in hash function str2tag_core()

JohannesFluegelHighyag:
For performance reasons the _httplib.h_ uses switch statements instead of if-else clauses with string comparisons, if it compares strings. As switch statements for strings are not possible in C++, the strings are transformed into a whole number via a hash function. The compiler evaluates 
```
switch (...)
case "any string"_t: ...
```
at compile time (i.e. `"any string"` is transformed via a user defined literal, the operator `_t`, and the hash function `str2tag_core()` into a unique `unsigned int` . During this transformation we have the evaluation `h * 33` with `unsigned int h`. If `h` is big (this happens for strings with a size greater than 6) , this evaluation causes an overflow, which is printed as the warning `warning C4307: '*': integral constant overflow` by my VS2017 compiler. Therefore I have changed this evaluation into `0x3ffffff & h * 33`, which never overflows, because the 6 high bits of `h` are cut (provided that `sizeof(unsigned int) == 4`).

Pro:
- No warnings anymore with VS2017 compiler.

Contra:
-  The hash function `str2tag_core()` might be a little bit worse. But keep in mind, that collisions are theoretically possible even without this change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -347,6 +347,27 @@ svr.Get("/chunked", [&](const Request& req, Response& res) {
 });
 ```
 
+With trailer:
+
+```cpp
+svr.Get("/chunked", [&](const Request& req, Response& res) {
+  res.set_header("Trailer", "Dummy1, Dummy2");
+  res.set_chunked_content_provider(
+    "text/plain",
+    [](size_t offset, DataSink &sink) {
+      sink.write("123", 3);
+      sink.write("345", 3);
+      sink.write("789", 3);
+      sink.done_with_trailer({
+        {"Dummy1", "DummyVal1"},
+        {"Dummy2", "DummyVal2"}
+      });
+      return true;
+    }
+  );
+});
+```
+
 ### 'Expect: 100-continue' handler
 
 By default, the server sends a `100 Continue` response for an `Expect: 100-continue` header.
--- a/httplib.h
+++ b/httplib.h
@@ -2980,7 +2980,7 @@ inline constexpr unsigned int str2tag_core(const char *s, size_t l,
                                            unsigned int h) {
   return (l == 0) ? h
                   : str2tag_core(s + 1, l - 1,
-                                 (h * 33) ^ static_cast<unsigned char>(*s));
+                                 (0x3ffffff & h * 33) ^ static_cast<unsigned char>(*s));
 }
 
 inline unsigned int str2tag(const std::string &s) {
--- a/httplib.h
+++ b/httplib.h
@@ -2978,9 +2978,13 @@ inline void get_remote_ip_and_port(socket_t sock, std::string &ip, int &port) {
 
 inline constexpr unsigned int str2tag_core(const char *s, size_t l,
                                            unsigned int h) {
-  return (l == 0) ? h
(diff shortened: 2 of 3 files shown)


yhirose:
@JohannesFluegelHighyag, thanks for the the pull request. On 64 bit environment (in other words, `unsigned int` is 8 b bytes), should it be `0x3ffffffffffff`?

JohannesFluegelHighyag:
@yhirose No, if `sizeof(unsigned int) == 8`, it should be `0x3ffffffffffffff` (because `0x3ffffffffffffff * 33 <= numeric_limits<unsigned int>::max()` should just be fulfilled).

Please keep in mind that a 64 bit environment does not necessarily mean `sizeof(unsigned int) == 8`.  In my VS2017 environment, I always have `sizeof(unsigned int) == 4`, regardless of whether I choose _x64_ or _x86_ in my _Solution Platforms_ control.

yhirose:
@JohannesFluegelHighyag, yes it should be `0x3ffffffffffffff`. Thanks for the correction.

How about the following code which supports any compilers?

```cpp
  return (l == 0)
             ? h
             : str2tag_core(
                   s + 1, l - 1,
                   ((std::numeric_limits<unsigned int>::max() >> 6) & h * 33) ^
                       static_cast<unsigned char>(*s));
```

JohannesFluegelHighyag:
@yhirose Generally it works. But it does not work for C++ MFC applications. The compiler shows the error 
```
...\httplib.h(3015): warning C4003: not enough arguments for function-like macro invocation 'max'
...\httplib.h(3015): error C2589: '(': illegal token on right side of '::'
...\httplib.h(3015): error C2059: syntax error: ')'
...\httplib.h(3011): error C3615: constexpr function 'httplib::detail::str2tag_core' cannot result in a constant expression
```
It assumes `max()` to be 
```
#define max(a,b) (((a) > (b)) ? (a) : (b))
```
from _C:\Program Files (x86)\Windows Kits\10\Include\10.0.17763.0\shared\minwindef.h_
instead of the correct
```
_NODISCARD static constexpr unsigned int (max)() noexcept
	{	// return maximum value
	return (UINT_MAX);
	}
```
from _C:\Program Files (x86)\Microsoft Visual Studio\2017\Professional\VC\Tools\MSVC\14.16.27023\include\limits_.
This is very strange.

yhirose:
How about this?

```cpp
  return (l == 0)
             ? h
             : str2tag_core(
                   s + 1, l - 1,
                   (((std::numeric_limits<unsigned int>::max)() >> 6) & h * 33) ^
                       static_cast<unsigned char>(*s));
```

yhirose:
Also you could you put a source code comment, so that others can understand why `>> 6` is necessary?

Tachi107:
> It assumes `max()` to be
> 
> ```
> #define max(a,b) (((a) > (b)) ? (a) : (b))
> ```

What if you define `NOMINMAX`?

JohannesFluegelHighyag:
@yhirose This is a very clever solution to add additional parentheses. Comment added.

PR #1533, opened 2023-03-25
SSL access through a proxy causes SIGSEGV

yosh-matsuda:
I've found that GET requests via proxy cause SIGSEGV for SSL access.

@yhirose Could you review the following and the PR?

## How to reproduce

OS: Ubuntu 22.04.2  (x86_64)

Set up Squid.

```bash
$ docker run -d --name squid-container -e TZ=JST -p 3128:3128 ubuntu/squid:5.2-22.04_beta
```

Code to reproduce.

```c++
Client cli("https://www.yahoo.com");
cli.set_proxy("localhost", 3128);
auto res = cli.Get("/");
```

## Detail

It seems that the problem is originated in the commit c7e959a9489bd37fe7d796e679e48894f272e22a at the `ClientImpl::process_request` function. I don't understand the implementation details around that code, but it is probably due to the lack of a socket open check, and this seems to fix SIGSEGV.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -6824,7 +6824,7 @@ inline bool ClientImpl::process_request(Stream &strm, Request &req,
   if (!write_request(strm, req, close_connection, error)) { return false; }
 
 #ifdef CPPHTTPLIB_OPENSSL_SUPPORT
-  if (is_ssl()) {
+  if (is_ssl() && socket_.is_open()) {
     char buf[1];
     if (SSL_peek(socket_.ssl, buf, 1) == 0 &&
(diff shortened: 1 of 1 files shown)


yhirose:
@yosh-matsuda thanks for the bug report and pull request. I tested it on my Mac and this pull request didn't fix the crash on macOS at f4b02df. I fixed it in a different way, and confirmed that it now fixes the problem on macOS. Could you test it on your linux environment? Thanks for your help!

yosh-matsuda:
@yhirose Thank you very much. It works fine on my Ubuntu.

yhirose:
@yosh-matsuda this fix is included in v12.0.2. Thanks for your fine contribution. 👍

PR #1534, opened 2023-03-25
Fix lifetime issues when test fails

jimmy-park:
Many test cases have a lifetime issue. It is okay on test success, but something is missed on test failure. For example, following code calls `std::terminate()` because it tries to return at `ASSERT_TRUE()` without joining the listening thread.

```cpp
Server svr;
auto listen_thread = std::thread([&svr]{ svr.listen("127.0.0.1", 1234); });

Client cli("127.0.0.1", 1234);
auto res = cli.Get("/hi");
ASSERT_TRUE(res);

// Not called on failure
svr.stop();
listen_thread.join();
```

To prevent this, I implemented `JoiningThread` and `RAIIWrapper`. `JoiningThread` is basically [`std::jthread`](https://en.cppreference.com/w/cpp/thread/jthread) without [`std::stop_token`](https://en.cppreference.com/w/cpp/thread/stop_token) feature and `RAIIWrapper` calls `Server::stop()` at its destruction. We can now correct the order of destruction using these helpers.

```cpp
Server svr;
JoiningThread listen_thread;
RAIIWrapper<Server> svr_wrapper{svr};

listen_thread = JoiningThread([&svr]{ svr.listen("127.0.0.1", 1234); });

Client cli("127.0.0.1", 1234);
auto res = cli.Get("/hi");
ASSERT_TRUE(res);

// These are redundant now
svr.stop();
listen_thread.join();
```

### DIFF
--- a/test/test.cc
+++ b/test/test.cc
@@ -63,6 +63,15 @@ private:
   std::thread thread_;
 };
 
+template <typename Server> class RAIIWrapper {
+public:
+  RAIIWrapper(Server &server) : server_{server} {}
+  ~RAIIWrapper() { server_.stop(); }
+
+private:
+  Server &server_;
+};
+
 MultipartFormData &get_file_value(MultipartFormDataItems &files,
                                   const char *key) {
   auto it = std::find_if(
--- a/test/test.cc
+++ b/test/test.cc
@@ -1030,8 +1030,10 @@ TEST(ReceiveSignals, Signal) {
   };
 
   Server svr;
+  JoiningThread listen_thread;
+  RAIIWrapper<Server> svr_wrapper{svr};
   int port = 0;
-  auto thread = JoiningThread([&]() {
+  listen_thread = JoiningThread([&]() {
     setupSignalHandlers();
     port = svr.bind_to_any_port("localhost");
     svr.listen_after_bind();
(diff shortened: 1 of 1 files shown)


yhirose:
@jimmy-park thanks for showing the problem and fix. But I would not like to introduce things like `JoiningThread` and `RAIIWrapper` because this change will force other contributors to follow this new coding standard.

I came up with another solution with `detail::scope_exit` function. I know this is still not perfect but a more generic solution. I just updated the very last unit text with this approach in `test/test.cc`.

```cpp
TEST(RedirectTest, RedirectToUrlWithQueryParameters) {
  Server svr;

  svr.Get("/", [](const Request & /*req*/, Response &res) {
    res.set_redirect(R"(/hello?key=val%26key2%3Dval2)");
  });

  svr.Get("/hello", [](const Request &req, Response &res) {
    res.set_content(req.get_param_value("key"), "text/plain");
  });

  auto thread = std::thread([&]() { svr.listen(HOST, PORT); });

  auto se = detail::scope_exit([&](void) {
    svr.stop();
    thread.join();
    ASSERT_FALSE(svr.is_running());
  });

  std::this_thread::sleep_for(std::chrono::seconds(1));

  {
    Client cli(HOST, PORT);
    cli.set_follow_location(true);

    auto res = cli.Get("/");
    ASSERT_TRUE(res);
    EXPECT_EQ(200, res->status);
    EXPECT_EQ("val&key2=val2", res->body);
  }
}
```

The `scope_exit` is pretty much similar to Golang's `defer`.

jimmy-park:
@yhirose I didn't know `detail::scope_exit` was already implemented. I also think `detail::scope_exit` is more generic and appropriate in this case. I will close this pr for now and apply `detail::scope_exit` in new pr.

PR #1535, opened 2023-03-26
Fix lifetime issues in test using detail::scope_exit()

jimmy-park:
Fix issues reported in #1534.

### DIFF
--- a/test/test.cc
+++ b/test/test.cc
@@ -1003,7 +1003,7 @@ TEST(ReceiveSignals, Signal) {
   });
   auto se = detail::scope_exit([&] {
     svr.stop();
-    if (thread.joinable()) thread.join();
+    thread.join();
     ASSERT_FALSE(svr.is_running());
   });
 
@@ -1042,9 +1042,9 @@ TEST(RedirectToDifferentPort, Redirect) {
   });
   auto se = detail::scope_exit([&] {
     svr2.stop();
-    if (thread2.joinable()) thread2.join();
+    thread2.join();
     svr1.stop();
-    if (thread1.joinable()) thread1.join();
+    thread1.join();
     ASSERT_FALSE(svr2.is_running());
     ASSERT_FALSE(svr1.is_running());
   });
@@ -1080,7 +1080,7 @@ TEST(RedirectFromPageWithContent, Redirect) {
   auto th = std::thread([&]() { svr.listen("localhost", PORT); });
   auto se = detail::scope_exit([&] {
     svr.stop();
-    if (th.joinable()) th.join();
+    th.join();
     ASSERT_FALSE(svr.is_running());
   });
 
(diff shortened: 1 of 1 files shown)


yhirose:
@jimmy-park thank you for the changes. Could you explain why `if (thread.joinable())` is needed?

jimmy-park:
@yhirose I added `std::thread::joinable()` for safety reasons. If some test cases need to join the thread before destruction of `detail::scope_exit()`, this will prevent double joining. I think this is a good habit when using `std::thread`.

yhirose:
@jimmy-park I think we don't need the condition since `std::thread` objects in `test/test.cc` are managed in the default threads and are not referenced in others threads.
 
https://stackoverflow.com/questions/60339713/is-joinable-then-join-thread-safe-in-stdthread

Please let me know if my understanding isn't correct. :)

jimmy-park:
@yhirose How about this example?
https://github.com/yhirose/cpp-httplib/blob/76230db97ff8945438ec4bb5eb7fd6630d0311a7/test/test.cc#L3942-L3965

When we apply `detail::scope_exit`, `svr.stop()` and `thread.join()` (line 3959-3960) should remain for testing closed server. So I need to check inside `detail::scope_exit` because `std::thread` is already joined or maybe somebody writes `thread.join()` in future.

Actually I handled as follows though the second joinable check is redundant at this time.
https://github.com/jimmy-park/cpp-httplib/blob/6db6723115b44c750b3daae913d3f367ee51bb31/test/test.cc#L3960-L3988

yhirose:
@jimmy-park in this particular case, you can leave the test as it is without using `detail::scope_exit`. This is just a unit test. So we don't have to make everything 'perfect', and should be reasonable.

jimmy-park:
@yhirose We still need `detail::scope_exit` even though it is a unit test. If `ASSERT_TRUE(res)` (line 3956) fails, the program crashes and rest tests won't run. It is even worse if we are using `std::async`, the program does not finish on failure. This happened to me when I tested on Windows.

jimmy-park:
@yhirose Okay then I'll remove checking joinable threads.

yhirose:
@jimmy-park thanks for your help!

test/test.cc
```diff
@@ -2855,13 +2869,13 @@ TEST_F(ServerTest, GetStreamedEndless) {
...
       auto res = cli_.Get("/streamed-cancel",
                           [&](const char *, uint64_t) { return true; });
       ASSERT_TRUE(!res);
       EXPECT_TRUE(res.error() == Error::Canceled ||
                   res.error() == Error::Read || res.error() == Error::Write);
-    }));
+    });
   }
 
   std::this_thread::sleep_for(std::chrono::seconds(2));
```
jimmy-park, line 2881: @yhirose For some reason, `ServerTest.ClientStop` failed and ran indefinitely (please see [results](https://github.com/yhirose/cpp-httplib/actions/runs/4539121445/jobs/7998678443)). I'm not sure what could be causing this problem. My change to `ServerTest.ClientStop` was trivial.

yhirose:
@jimmy-park could you send another pull request?

Issue #1093, opened 2021-11-16
Remove meson-build tests from GitHubActions


yhirose:
Since the regular tests on GitHubActions are enough for testing all the httplib functionalities, and the meson-build tests costs about extra 10 minutes and the test on Windows is very unstable. I would like to leave only absolutely necessary tests in GitHub workflow.

PR #1537, opened 2023-03-26
Add more workflows (CMake/CodeQL)

jimmy-park:
I'm trying to provide comprehensive configurations for testing. I think current workflow is limited to support various configurations. So I adopted CMake in workflow.

# CMake

In GitHub Actions, a total of 64(=4x2x2x2x2) job configurations are created with the combination of following variables.

- compiler : GCC (Ubuntu), Clang (Ubuntu), AppleClang (macOS), MSVC (Windows)
- config : Debug, Release
- OpenSSL : ON, OFF
- brotli : ON, OFF
- zlib : ON, OFF

Some dependencies are already installed on latest runners. Unfortunately, Windows package manager (chocolatey) doesn't provide `brotli` and `zlib`. So they are built and installed manually inside workflows.

# CodeQL

CodeQL helps to find vulnerabilities and can be easily added using a template provided by GitHub. But, since `cpp-httplib` is a header-only library, we need an executable for CodeQL. `HTTPLIB_TEST` environment variable is added for this purpose.

### DIFF
--- a/.github/workflows/cmake.yaml
+++ b/.github/workflows/cmake.yaml
@@ -121,3 +121,52 @@ jobs:
 
       - name: Install
         run: cmake --install build --config ${{ matrix.config }}
+
+  build-mobile:
+    strategy:
+      fail-fast: false
+      matrix:
+        preset: [android, ios]
+        config: [Debug, Release]
+        include:
+          - preset: android
+            os: ubuntu-latest
+            generator: '"Ninja Multi-Config"'
+          - preset: ios
+            os: macos-latest
+            generator: Xcode
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Install dependencies on Linux
+        if: runner.os == 'Linux'
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y ninja-build
+
+      - name: Find android.toolchain.cmake
+        if: matrix.preset == 'android'
+        run: echo "set(CMAKE_TOOLCHAIN_FILE ${ANDROID_NDK}/build/cmake/android.toolchain.cmake CACHE STRING \"\")" >> cache.cmake
+
+      - name: Generate ios.toolchain.cmake
+        if: matrix.preset == 'ios'
+        run: |
+          echo 'set(CMAKE_SYSTEM_NAME       iOS)'       >> ios.toolchain.cmake
+          echo 'set(CMAKE_SYSTEM_PROCESSOR  aarch64)'   >> ios.toolchain.cmake
+          echo 'set(CMAKE_OSX_ARCHITECTURES arm64)'     >> ios.toolchain.cmake
+          echo 'set(CMAKE_OSX_SYSROOT       iphoneos)'  >> ios.toolchain.cmake
+          echo 'set(CMAKE_TOOLCHAIN_FILE    ios.toolchain.cmake CACHE STRING "")' >> cache.cmake
+
+      - name: Configure CMake
+        run: >
+          cmake -S . -B build -G ${{ matrix.generator }} -C cache.cmake
+          -DCMAKE_INSTALL_PREFIX=${{ github.workspace }}/install
+          -DHTTPLIB_COMPILE=ON
+
+      - name: Build
+        run: cmake --build build --config ${{ matrix.config }}
+
+      - name: Install
+        run: cmake --install build --config ${{ matrix.config }}
--- a/.github/workflows/cmake.yaml
+++ b/.github/workflows/cmake.yaml
@@ -1,0 +1,121 @@
+name: CMake
+
+on:
+  push:
+    paths: ["**/CMakeLists.txt"]
+  pull_request:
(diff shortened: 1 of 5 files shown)


jimmy-park:
Workflows for Android/iOS are added. I didn't include dependencies (except `zlib`) because they need to be cross-compiled. Especially building and linking `OpenSSL` in the workflow is not trivial.

.github/workflows/cmake.yaml
```diff
@@ -0,0 +1,169 @@
...
+      - name: Install dependencies on Windows (zlib)
+        if: runner.os == 'Windows' && matrix.zlib == 'ON'
+        run: |
+          cmake -S zlib -B zlib/build -G ${{ matrix.generator }} -C cache.cmake
+          cmake --build zlib/build --target install --config ${{ matrix.config }}
+          echo 'set(ZLIB_ROOT ${CMAKE_INSTALL_PREFIX} CACHE STRING "")' >> cache.cmake
+
+      - name: Configure CMake
+        run: >
+          cmake -S . -B build -G ${{ matrix.generator }} -C cache.cmake
```
Tachi107, line 93: `-S .` is implied by default, and it can be omitted

.github/workflows/cmake.yaml
```diff
@@ -0,0 +1,169 @@
...
+        run: |
+          echo 'set(CMAKE_SYSTEM_NAME       iOS)'       >> ios.toolchain.cmake
+          echo 'set(CMAKE_SYSTEM_PROCESSOR  aarch64)'   >> ios.toolchain.cmake
+          echo 'set(CMAKE_OSX_ARCHITECTURES arm64)'     >> ios.toolchain.cmake
+          echo 'set(CMAKE_OSX_SYSROOT       iphoneos)'  >> ios.toolchain.cmake
+          echo 'set(CMAKE_TOOLCHAIN_FILE    ios.toolchain.cmake CACHE STRING "")' >> cache.cmake
+
+      - name: Configure CMake
+        run: >
+          cmake -S . -B build -G ${{ matrix.generator }} -C cache.cmake
```
Tachi107, line 161: Same as above

test/test.cc
```diff
@@ -3287,7 +3287,8 @@ TEST(GzipDecompressor, ChunkedDecompression) {
 }
 
 #ifdef _WIN32
-TEST(GzipDecompressor, LargeRandomData) {
+// Disabled due to the out-of-memory problem on GitHub Actions Workflows
+TEST(GzipDecompressor, DISABLED_LargeRandomData) {
```
Tachi107, line 3291: Since this is an issue with GitHub Actions, the test should be disabled there instead of here. You can do so by passing the `--gtest_filter=-GzipDecompressor.LargeRandomData` flag to the test binary or by setting the `GTEST_FILTER` environment variable to the same value

Tachi107 (changes requested):
I'm not sure yhirose is willing to add CI jobs for build systems maintained by external contributors like Meson and CMake. Tests for Meson existed in the past, but were removed in commit 6fb5b630183b4345ff54b8fe3119788373c42db3 - see issue #1093 for the rationale.

If CMake tests are still desired, I think that launching 64 different build configurations is a bit too much. For example, I find testing with both GCC and Clang on Linux unnecessary, since they shouldn't be generating too different code, and Clang builds get already tested on macOS. It may be more significant to test with different C and C++ library implementations, and that is already the case with Linux (glibc + libstdc++), macOS (LLVM libc++) and Windows (MSVC's libraries).

By the way, one interesting additional test platform would be something like OpenBSD, which ships many kinds of hardenings by default which can uncover bugs in code.

One last additional testing setup which I'd recommend is a test matrix which builds code under different sanitizers (AddressSanitizer, ThreadSanitizer, UndefinedBehaviorSanitizer) - that is really effective at uncovering typical C and C++ issues.

Anyway, I digressed. To conclude, one CI setup that is more likely to get merged is one which doesn't touch Meson and CMake code, but that still expands coverage with sanitizers and additional OSes. Also because, if you merge this, Meson tests should be brought back too, and that would duplicate the test matrix...

Thanks for reading this long feedback!

Tachi107:
> CodeQL helps to find vulnerabilities and can be easily added using a template provided by GitHub. But, since `cpp-httplib` is a header-only library, we need an executable for CodeQL. `HTTPLIB_TEST` environment variable is added for this purpose.

You can always compile the library with `-Dcpp-httplib_compile=true` on Meson and which a similar flag on CMake (but I'm not sure this is relevant here)

yhirose:
@jimmy-park @Tachi107 thanks for this pull request extending the workflows. But personally, I am not interested in merging this pull request unless I feel this enhancement is absolutely necessary. It seems like this change will overly complicates the build process and cause the build time much longer. Thanks for your understanding!

jimmy-park:
@Tachi107 Thanks for detailed comments! I agree that workflows with sanitizers would be nice. But I didn't consider the Meson build which makes workflows more complicated than I thought. So I've decided to drop this pr for now.

@yhirose Understood. I'll close pr and use this workflow personally.

PR #1542, opened 2023-03-30
Fix `ServerTest.ClientStop` test case

jimmy-park:
`ServerTest.ClientStop` does not stop and runs forever in some situation. (please see [this](https://github.com/yhirose/cpp-httplib/actions/runs/4539121445/jobs/7998678443))
This happens very rarely but I found that the thread could be started after stopping the client. Waiting 2 seconds was not enough sometimes. This is because the GitHub Actions runner has limited hardware resources.


yhirose:
@jimmy-park thanks for the fix!

PR #1546, opened 2023-04-01
Don't loading system certs from Keychain on iOS

jimmy-park:
Changes in #1474 are not suitable for iOS. Its implementation is macOS only. So we need to distinguish macOS and iOS.
[`TargetConditionals.h`](https://github.com/xybp888/iOS-SDKs/blob/master/iPhoneOS16.4.sdk/usr/include/TargetConditionals.h) defines useful macros for Apple platforms. We can target macOS only by using `TARGET_OS_OSX` instead of `__APPLE__`.

This may conflict with #1525

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -239,10 +239,13 @@ using socket_t = int;
 #pragma comment(lib, "crypt32.lib")
 #pragma comment(lib, "cryptui.lib")
 #endif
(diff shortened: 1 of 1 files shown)


yhirose:
Thanks for the improvement!

PR #1554, opened 2023-04-09
Fix leaked handle in create_socket

trollixx:
Coverity Scan has flagged the following potential resource leak in `cpp-httplib`, which is bundled in [Zeal](https://github.com/zealdocs/zeal). It is the only "high impact" issue. There are other lower priority potential problems, which I may try to fix later.

image


yhirose:
@trollixx thanks for the fine contribution!

PR #1559, opened 2023-04-26
handle_file_request - handling of large files with ranges with less resources

stanislavpetkov:
handle_file_request rewritten,
in case of RANGE header, only the part of ranges will be loaded and transmitted. 
Before even if you need 10bytes of 2TB file it will try to load it in memory, and than will apply ranges.

file_request_handler_ call is moved before post_routing_handler_  (I don't know why file_request_handler_ is called before data delivery, but I kept it the same way)

find_content_type - 
1. changed return type to string, as it is converted to string anyway .
2. default type is changed to application/octetstring
3. added some extra types

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -547,6 +547,7 @@ struct Response {
   ContentProviderResourceReleaser content_provider_resource_releaser_;
   bool is_chunked_content_provider_ = false;
   bool content_provider_success_ = false;
+  bool is_file_delivery_ = false;
 };
 
 class Stream {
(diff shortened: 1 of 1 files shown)


yhirose:
@stanislavpetkov thanks for the pull request. Before I start reviewing it, could you do the following things?

1) Add at least one unit test to validate your change in `test/test.cc`
2) Format your changes with clang-format with options in `.clang-format` in this repo.

Thanks!

stanislavpetkov:
@yhirose 
1. _Add at least one unit test to validate your change in `test/test.cc`_ - it is covered by the range headers and static content download tests. I can't add anything more
2. _Format your changes with clang-format with options in `.clang-format` in this repo._ - Done from MSVC - it detected the clang-format

yhirose:
@stanislavpetkov I reviewed the code, but I decided implementing it in a different way. It supports not only a single range, but multiple ranges properly. Also better performance and less memory usage with std::array instead of std::vector. You can see my change at https://github.com/yhirose/cpp-httplib/commit/6bb580cda8fd3063e7141588d96db087b9e50420. Thanks!

PR #1563, opened 2023-04-29
Fix up CMake exports

rnichollx:
The current approach doesn't work on all platforms/toolchains.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -237,13 +237,6 @@ set(_TARGET_INSTALL_CMAKEDIR "${CMAKE_INSTALL_LIBDIR}/cmake/${PROJECT_NAME}")
 
 include(CMakePackageConfigHelpers)
 
-# Configures the meta-file httplibConfig.cmake.in to replace variables with paths/values/etc.
-configure_package_config_file("${PROJECT_NAME}Config.cmake.in"
-	"${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}Config.cmake"
-	INSTALL_DESTINATION "${_TARGET_INSTALL_CMAKEDIR}"
-	# Passes the includedir install path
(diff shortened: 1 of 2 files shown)


yhirose:
@sum01, @abouvier, @jimmy-park, @Tachi107, do you have any thoughts?

abouvier:
What are you trying to fix with that? Do you have a non-working example to show?
You remove the possibility of using an installed library and only allow the use of the library from the build tree by another project.

rnichollx:
> What are you trying to fix with that? Do you have a non-working example to show? You remove the possibility of using an installed library and only allow the use of the library from the build tree by another project.

That was an error, it should work with the install tree with the additional commit, I had the wrong path included.

For me it does not work using Ninja on Windows with MSVC compilers. I am not sure why but it was trying to use my home directory/include as the location for header files. (C:\Users\myname\include does not exist error)

Tachi107:
Hi all, thanks for the ping.

I'm not sure what you're trying to fix, but this is probably not the right solution.

If you're able to provide an example showing what is going wrong with the current build script we can try to find an alternative solution that doesn't cause the issues outlined by @abouvier

sum01:
Not much to say when everyone's already said it.

As far as I can tell you want it to use a different include dir than the default? If so, just pass a different one at build time, don't change/delete large swaths of code.

rnichollx:
> Not much to say when everyone's already said it.
> 
> As far as I can tell you want it to use a different include dir than the default? If so, just pass a different one at build time, don't change/delete large swaths of code.

The default include directory is wrong when used against a build tree.

yhirose:
@rnichollx since I am not an expert of CMake, I basically delegate any changes to the contributors who have been maintaining CMake related files, and trust what they suggest. So please work with them and apply their suggestion first. When all approve your change, then I'll merge it right away. Thanks for your understanding!

PR #1575, opened 2023-05-29
Add HTTPLIB_INSTALL CMake option

rotolof:
In my project I’m fetching httplib with FetchContent:

```cmake
FetchContent_Declare(
  cpp-httplib
  GIT_REPOSITORY https://github.com/yhirose/cpp-httplib
)
FetchContent_MakeAvailable(cpp-httplib)
```

But when I install my project, httplib gets installed too because the library's `install()` commands are there. This may not be ideal if, like in my case, the dependency is private.
The same thing would happen if I were to add the project with `add_subdirectory()`.

So I added an `HTTPLIB_INSTALL` option (default: `ON`) that disables the installation commands, now I can fetch the library like this:

```cmake
FetchContent_Declare(
  cpp-httplib
  GIT_REPOSITORY https://github.com/yhirose/cpp-httplib
)
set(HTTPLIB_INSTALL OFF)
FetchContent_MakeAvailable(cpp-httplib)
```

We could also add something like:

```cmake
# Checks if it is the main project or if it is being included as a subproject (using add_subdirectory/FetchContent).
set(MAIN_PROJECT OFF)
if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
    set(MAIN_PROJECT ON)
endif()

option(HTTPLIB_INSTALL "..." ${MAIN_PROJECT})
```

But it's probably too much.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -265,33 +265,33 @@ else()
 	)
 endif()
 
-if (HTTPLIB_INSTALL)
+if(HTTPLIB_INSTALL)
 	# Creates the export httplibTargets.cmake
 	# This is strictly what holds compilation requirements
 	# and linkage information (doesn't find deps though).
 	install(TARGETS ${PROJECT_NAME}
-			EXPORT httplibTargets
-			)
+		EXPORT httplibTargets
+	)
 
 	install(FILES "${_httplib_build_includedir}/httplib.h" TYPE INCLUDE)
 
 	install(FILES
-			"${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}Config.cmake"
-			"${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}ConfigVersion.cmake"
-			# Install it so it can be used later by the httplibConfig.cmake file.
-			# Put it in the same dir as our config file instead of a global path so we don't potentially stomp on other packages.
-			"${CMAKE_CURRENT_SOURCE_DIR}/cmake/FindBrotli.cmake"
-			DESTINATION ${_TARGET_INSTALL_CMAKEDIR}
-			)
+		"${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}Config.cmake"
+		"${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}ConfigVersion.cmake"
+		# Install it so it can be used later by the httplibConfig.cmake file.
+		# Put it in the same dir as our config file instead of a global path so we don't potentially stomp on other packages.
+		"${CMAKE_CURRENT_SOURCE_DIR}/cmake/FindBrotli.cmake"
+		DESTINATION ${_TARGET_INSTALL_CMAKEDIR}
+	)
 
 	# NOTE: This path changes depending on if it's on Windows or Linux
 	install(EXPORT httplibTargets
-			# Puts the targets into the httplib namespace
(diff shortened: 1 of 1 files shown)


yhirose:
@sum01, @abouvier, @jimmy-park, @Tachi107, do you have any thoughts?

CMakeLists.txt
```diff
@@ -262,31 +265,33 @@ else()
...
+			)
+
+	# NOTE: This path changes depending on if it's on Windows or Linux
+	install(EXPORT httplibTargets
+			# Puts the targets into the httplib namespace
+			# So this makes httplib::httplib linkable after doing find_package(httplib)
+			NAMESPACE ${PROJECT_NAME}::
+			DESTINATION ${_TARGET_INSTALL_CMAKEDIR}
+			)
+endif ()
```
jimmy-park, line 294: Minor fixes for formatting

```suggestion
if(HTTPLIB_INSTALL)
	# Creates the export httplibTargets.cmake
	# This is strictly what holds compilation requirements
	# and linkage information (doesn't find deps though).
	install(TARGETS ${PROJECT_NAME}
		EXPORT httplibTargets
	)

	install(FILES "${_httplib_build_includedir}/httplib.h" TYPE INCLUDE)

	install(FILES
		"${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}Config.cmake"
		"${CMAKE_CURRENT_BINARY_DIR}/${PROJECT_NAME}ConfigVersion.cmake"
		# Install it so it can be used later by the httplibConfig.cmake file.
		# Put it in the same dir as our config file instead of a global path so we don't potentially stomp on other packages.
		"${CMAKE_CURRENT_SOURCE_DIR}/cmake/FindBrotli.cmake"
		DESTINATION ${_TARGET_INSTALL_CMAKEDIR}
	)

	# NOTE: This path changes depending on if it's on Windows or Linux
	install(EXPORT httplibTargets
		# Puts the targets into the httplib namespace
		# So this makes httplib::httplib linkable after doing find_package(httplib)
		NAMESPACE ${PROJECT_NAME}::
		DESTINATION ${_TARGET_INSTALL_CMAKEDIR}
	)
endif()
```

abouvier:
Also [until `EXCLUDE_FROM_ALL` is supported](https://gitlab.kitware.com/cmake/cmake/-/issues/20167) in `FetchContent_Declare` you can use:
```
set_directory_properties(PROPERTIES EXCLUDE_FROM_ALL ON)
FetchContent_Declare(
  cpp-httplib
  GIT_REPOSITORY https://github.com/yhirose/cpp-httplib
)
FetchContent_MakeAvailable(cpp-httplib)
```

rotolof:
@abouvier thanks, I was not aware of this workaround. I still think that adding an explicit option would be more self-documenting, but otherwise it seems it's not as needed as I thought it was.

Tachi107:
Looks fine to me :)

sum01:
lgtm, just needs that [formatting fix](https://github.com/yhirose/cpp-httplib/pull/1575#discussion_r1209358622)

yhirose:
@rotolof since CMake experts approve, I'll merge it after the [formatting fix](https://github.com/yhirose/cpp-httplib/pull/1575#discussion_r1209358622) is made. Thanks!

rotolof:
@yhirose Done, thank you!

yhirose:
@rotolof thanks for your fine contribution!

PR #1576, opened 2023-06-01
Move, not copy, Logger and Handler functors

db-src:
Hi, and thanks a lot for the nice little library!

Here are a few minor improvements:

* Explicitly #include <utility> for use of std::move [rather than relying on indirect include]
* Move not copy Logger arg from Client to ClientImpl [will avoid copying heavy functors]
* Move not copy, set_error_handler Handler to lambda [ditto]
* Remove null statement in non-empty if/else block [I guess it was a relic from a time before the other statement was added.]

***

We might also consider whether to move `string()` arguments such as these from arguments into their destination - what do you think?

image

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -8751,7 +8751,9 @@ inline void Client::enable_server_certificate_verification(bool enabled) {
 }
 #endif
 
-inline void Client::set_logger(Logger logger) { cli_->set_logger(logger); }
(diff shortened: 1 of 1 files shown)


yhirose:
@db-src looks good to me. I'll merge it.  As for `string()`, I am ok with the existing code. Thanks for the improvement!

PR #1580, opened 2023-06-03
Provide a CMake option to disable C++ exceptions

petrhosek:
This allows disabling the use of C++ exceptions at CMake configure time. The value is encoded in the generated httplibTargets.cmake file and will be used by CMake projects that import it.


yhirose:
@petrhosek I am not maintaining CMake related files in this project though, your change looks good to me. Could you please add a comment right above the change as other options have? Thanks!

petrhosek:
> Could you please add a comment right above the change as other options have?

Done.

yhirose:
@petrhosek thanks for your contribution!

PR #1581, opened 2023-06-05
Add support for zOS

v1gnesh:


yhirose:
@v1gnesh thanks for the contribution, but I have no idea about zOS. I don't mind merging it, but I am not able to maintain it and might break zOS accidentally. Are you OK to maintain your changes whatever happens with it in the future? (I don't maintain CMake support either, but several contributes willingly maintain it right away when anything happens.) Thanks for your understanding!

v1gnesh:
Hi

> Are you OK to maintain your changes whatever happens with it in the future? 

Yes sure. Do tag me if something breaks because of these parts.

And of course, thank you for sharing your excellent work as open source 👍

yhirose:
Thanks for your fine help!

PR #1586, opened 2023-06-10
Use nghttp2-hosted httpbin.org

jimmy-park:
`httpbin.org` seems unavailable most of the time. We can use nghttp2-hosted version which is more stable.


yhirose:
Looks good! When the original httpbin.org becomes available again, I would like to switch back to it. Can we switch with each other easily with a macro or something in `test.cc`?

yhirose:
Thanks for your contribution!

PR #1589, opened 2023-06-12
Don't overwrite the last redirected location

jimmy-park:
Since redirection is handled recursively, multiple redirects result an invalid location which is the first redirected location. `res.location` is overwritten recursively.

Some test cases did not check the last location `https://www.yahoo.com`. They usually succeed even if they check the wrong location `https://yahoo.com/`. But sometimes they fail because `http://yahoo.com` redirects `http://www.yahoo.com`.
```
http://yahoo.com -> https://yahoo.com    -> https://www.yahoo.com
http://yahoo.com -> http://www.yahoo.com -> https://www.yahoo.com
```

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -3920,7 +3920,8 @@ inline bool redirect(T &cli, Request &req, Response &res,
   if (ret) {
     req = new_req;
     res = new_res;
-    res.location = location;
+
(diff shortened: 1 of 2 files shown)


yhirose:
Thanks for fixing the bug!

PR #1591, opened 2023-06-15
Fix -Wold-style-cast warnings

int2str:
Removed multiple old-style (C) casts and replaced them with the appropriate _cast's. This makes httplib a better citizen inside projects that are compiled with all warnings enabled.

Unfortunately one warning remains as a result of invoking an OpenSSL macro (at least on Linux), that would have to be fixed upstream.

Also fixed a few casts for invocations of setsockopt. setsockopt takes a const void* for the value pointer, not a char*.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -1675,17 +1675,17 @@ inline ssize_t Stream::write_format(const char *fmt, const Args &...args) {
 inline void default_socket_options(socket_t sock) {
   int yes = 1;
 #ifdef _WIN32
-  setsockopt(sock, SOL_SOCKET, SO_REUSEADDR, reinterpret_cast<char *>(&yes),
-             sizeof(yes));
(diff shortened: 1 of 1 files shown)


yhirose:
@int2str could you take a look at errors on 'test / build (windows-latest) (pull_request)'?

int2str:
> @int2str could you take a look at errors on 'test / build (windows-latest) (pull_request)'?

Done. Hopefully this fixes it. Don't have a Windows machine to build. Will monitor CI for any further issues.

yhirose:
There are still problems on Windows.

int2str:
Ahhhh, Windows..... :)
3rd time is hopefully the charm :D

yhirose:
@int2str thanks for your contribution!

PR #1592, opened 2023-06-16
Build examples using CMake

jimmy-park:
`example/CMakeLists.txt` is added for building examples with CMake. We can simply build them using `HTTPLIB_EXAMPLE` option.


yhirose:
@jimmy-park thanks for the pull request. But I am not going to take it, because those examples are not production quality but just my test example codes... So I would not like to include them in CMake. Thanks for your understanding!

jimmy-park:
Sure 😄

PR #1602, opened 2023-06-27
Ability to turn off linking against OpenSSL if already detected.

nmoinvaz:
We already detect OpenSSL in our tree for another project, but don't want to link httplib against OpenSSL or with SSL support.

Allow us to `set(HTTPLIB_IS_USING_OPENSSL FALSE)`.


yhirose:

PR #1312, opened 2022-06-26
restore compatibility with BoringSSL

artpaul:
* SSLSocketStream::read was restored because client can handle partial read
* SSLSocketStream::write was rewritten with loop

PR #1603, opened 2023-06-27
Fix type mismatch

artpaul:
g++12 reports the following problem:

httplib.h:7623:10: error: inconsistent deduction for ‘auto’: ‘int’ and then ‘long unsigned int’
 7623 |     for (auto first = 0, last = sk_X509_INFO_num(inf); first < last; ++first) {
      |          ^~~~

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -7620,7 +7620,8 @@ inline X509_STORE *ClientImpl::create_ca_cert_store(const char *ca_cert,
 
   auto cts = X509_STORE_new();
   if (cts) {
-    for (size_t first = 0, last = sk_X509_INFO_num(inf); first < last; ++first) {
+    for (size_t first = 0, last = sk_X509_INFO_num(inf); first < last;
+         ++first) {
       auto itmp = sk_X509_INFO_value(inf, first);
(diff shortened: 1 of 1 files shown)


yhirose:
According to the OpenSSL 3.0 source code, `sk_X509_INFO_num` returns a `int` value.

```cpp
#define sk_X509_INFO_num(sk) OPENSSL_sk_num(ossl_check_const_X509_INFO_sk_type(sk))
int OPENSSL_sk_num(const OPENSSL_STACK *);
```

So could you explain why this change is needed? Thanks!

artpaul:
I'm using your library with BoringSSL. You can remember me by this PR https://github.com/yhirose/cpp-httplib/pull/1312

And in BoringSSL the sk_X509_INFO_num function returns size_t.

I would be grateful if you will keep compatibility with the BoringSSL.

artpaul:
Another way to fix this issue it to put explicit cast to int like this:
https://cs.android.com/android/platform/superproject/+/master:external/curl/lib/vtls/openssl.c;l=2571;drc=0ee5fb7bc803ecfad90276d39f73696d024e5939

Issue #1587, opened 2023-06-10
Named path parameters


bugdea1er:
For now, the server is putting the responsibility on me to write the correct regex and duplicate it in every path. For example, it is not enough to use `(.+)` for project ID for paths 
```http
GET /projects/11/files HTTP 1.1
```

and

```http
GET /projects/11 HTTP 1.1
```

since it would also require to bind them to the server in the proper order.

I suggest to adopt some familiar syntax for parameters in paths (like, `{id}` or `:id`) and do not require to write regexes from the user:

```cpp
srv.Get("/projects/{id}/item", handler);
```

Such syntax would also allow me to get parameter values by the key instead of the order with some kind of operator[]:
```cpp
void handler(auto req, auto res) {
  auto project_id = req["id"];
}
```

yhirose:
@bugdea1er I am not planning to support `{...}` syntax because it requires a drastic change in the parser.

Is the following way enough for your case?

```cpp
srv.Get(R"(/projects/(\d+)/item)", [](auto req, auto res) {
  auto id = req.matches[1];
});
```

bugdea1er:
@yhirose It will go as a temporary solution, but as for me, it’s more correct to catch all requests that match the URL, and only on the backend side decide whether this format is suitable or not (responding with "404 Project not found" for /projects/string)

If you are ok with this, then you can leave the ticket open. Maybe someday I or someone else will throw a pull request on this topic.

yhirose:
@bugdea1er one of the possible problems to introduce the `{...}` syntax is that it conflicts with a Regex repetition operator like `[1-9][0-9]{3}` which also uses curly braces. So I don't think it's easy to seamlessly support both the named path parameters and the current regular expression match implementation in cpp-httplib... I guess we have to implement a completely different parser for the named path parameters and allow users to pick either one based on their requirements.

On my side, I am not planning to implement the feature since I am satisfied with the current regex way. Are you seriously thinking of implementing the feature in the near future?

bugdea1er:
@yhirose we are planning to modify this library in our project to support such named parameters. I think we can try to support both old and new syntax (possibly different syntax) and if we succeed we will send you a pr

PR #1608, opened 2023-07-04
Add named path parameters parsing (Implements #1587)

bgs99:
As discussed with @bugdea1er, I have implemented named path parameter parsing.
New parser is accessible by a different set of Get/Post/Put/Patch/Delete/Options methods with a suffix "Simple" (open for suggestions).

Example of usage from tests:
```cpp
  svr.GetSimple("/users/:id", [&](const Request &req, Response &) {
    EXPECT_EQ("user-id", req.path_params.at("id"));
    EXPECT_EQ("foo", req.get_param_value("param1"));
    EXPECT_EQ("bar", req.get_param_value("param2"));
  });
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,11 +94,20 @@ int main(void)
     res.set_content("Hello World!", "text/plain");
   });
 
+  // Match the request path against a regular expression
+  // and extract its captures
   svr.Get(R"(/numbers/(\d+))", [&](const Request& req, Response& res) {
     auto numbers = req.matches[1];
     res.set_content(numbers, "text/plain");
   });
 
+  // Capture the second segment of the request path as "id" path param
+  svr.Get("/users/:id", [&](const Request& req, Response& res) {
+    auto user_id = req.path_params.at("id");
+    res.set_content(user_id, "text/plain");
+  });
+
+  // Extract values from HTTP headers and URL query params
   svr.Get("/body-header-param", [](const Request& req, Response& res) {
     if (req.has_header("Content-Length")) {
       auto val = req.get_header_value("Content-Length");
--- a/httplib.h
+++ b/httplib.h
@@ -749,21 +749,6 @@ public:
   Server &Delete(const std::string &pattern, HandlerWithContentReader handler);
   Server &Options(const std::string &pattern, Handler handler);
 
-  Server &GetSimple(const std::string &pattern, Handler handler);
-  Server &PostSimple(const std::string &pattern, Handler handler);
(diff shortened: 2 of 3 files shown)


bugdea1er:
@yhirose one of the possible solutions for this naming is to hide the new functions behind an optional compilation flag like to did with ssl support. I doubt that anyone would use both options for paths in the same project

We want your opinion

yhirose:
Great pull request! Indeed, the 'Simple' suffix might cause confusion among users. Personally, I would prefer to retain both the regex matcher and the path params matcher. How about checking at runtime whether the path includes /: at least once, and based on that, determining if it's in 'path params' mode? The code could be modified as follows:

```cpp
inline Server &Server::Get(const std::string &pattern, Handler handler) {
  auto matcher = pattern.find("/:") != std::string::npos
    ? detail::make_unique<detail::PathParamsMatcher>(pattern),
    : detail::make_unique<detail::RegexMatcher>(pattern);

  get_handlers_.push_back(std::make_pair(matcher, std::move(handler)));
  return *this;
}
```

By following the above approach, we can maintain just one `Get` method with a minimal impact on performance. The only limitation of this approach is that we can no longer utilize /: in the regular expression pattern. However, I believe this constraint is reasonable and won't cause any inconvenience for existing users. What do you think?

bgs99:
Great idea!
I was thinking about something like that but was not sure how you would judge the risk of breaking some convoluted setups.
I will update the PR tomorrow, if nothing goes wrong.

yhirose:
Or something like this?

```cpp
inline std::unique_ptr<detail::MatcherBase> Server::make_matcher(
    const std::string &pattern) const {
  return pattern.find("/:") != std::string::npos
         ? detail::make_unique<detail::PathParamsMatcher>(pattern),
         : detail::make_unique<detail::RegexMatcher>(pattern);
}

inline Server &Server::Get(const std::string &pattern, Handler handler) {
  get_handlers_.push_back(
      std::make_pair(make_matcher(pattern), std::move(handler)));
  return *this;
}
```

Could you please also update README to show calling examples? Thanks a lot!

bgs99:
Done.
Used an `if` statement instead of the ternary operator though, as the latter won't compile with different argument types.

yhirose:
Thanks for the fine contribution! I believe many users will like the new feature. 👍

yhirose:
@bgs99 could you take a look at #1805? It seems like this PR creates the issue. Thanks a lot!

PR #1609, opened 2023-07-05
Result: allow default constructor

vmaffione:
Allow default constructor for `Result`, to support use cases such as the following:
````
       [...]
       httplib::Client client(host);
       httplib::Result result; // needed here

        if (method == "GET") {
                result = client.Get(path);

        } else if (method == "POST") {
                result = client.Post(path, headers, body,
                                "application/json");
         }

         if (!result) {
             std::cout << "Error";
             return -1;
         }

         // use result
         [...]
````

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -878,6 +878,7 @@ std::ostream &operator<<(std::ostream &os, const Error &obj);
 
 class Result {
 public:
+  Result() = default;
   Result(std::unique_ptr<Response> &&res, Error err,
          Headers &&request_headers = Headers{})
       : res_(std::move(res)), err_(err),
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: httplib.h [hunk 1]
```diff
@@ -878,6 +878,7 @@ std::ostream &operator<<(std::ostream &os, const Error &obj);
 
 class Result {
 public:
+  Result() {}
```
yhirose, line 881: I prefer `Result() = default;` instead.

vmaffione: ofc, thanks!

yhirose:
@vmaffione thanks for your contribution!

PR #1612, opened 2023-07-07
Fix successful decompress reported as Error::Read

bdenhollander:
Streams less than 4096 bytes are sometimes reported as failed reads because stream_.avail_in is not reduced to 0. The next iteration of the loop finds `prev_avail_in == strm_.avail_in` and return false. `ret = inflate(...)` returns `Z_STREAM_END` on the first iteration of the loop indicating that inflate is finished. This fix prevents the second iteration of the loop from failing.

Relevant zlib.h source code:
https://github.com/madler/zlib/blob/04f42ceca40f73e2978b50e93806c2a18c1281fc/zlib.h#L462-L464

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -3384,15 +3384,11 @@ inline bool gzip_decompressor::decompress(const char *data, size_t data_length,
     data += strm_.avail_in;
 
     std::array<char, CPPHTTPLIB_COMPRESSION_BUFSIZ> buff{};
-    while (strm_.avail_in > 0) {
+    while (strm_.avail_in > 0 && ret == Z_OK) {
       strm_.avail_out = static_cast<uInt>(buff.size());
       strm_.next_out = reinterpret_cast<Bytef *>(buff.data());
 
-      auto prev_avail_in = strm_.avail_in;
-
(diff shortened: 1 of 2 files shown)


yhirose:
@bdenhollander could you please include at least one unit test in test/test.cc as I typically request from contributors when they send me a pull request? Thank you!

bdenhollander:
@yhirose Writing the unit test helped narrow down the original source of the problem in my code that uses cpp-httplib, an incorrect Content-Length was including garbage after the end of the data I had compressed. Fetching the file with the extra bytes via browser fetch was successfully skipping the extra bytes but cpp-httplib couldn't handle them.

yhirose:
> but cpp-httplib couldn't handle them

Does your fix handle the above issue as well, right?

bdenhollander:
> > but cpp-httplib couldn't handle them
> 
> Does your fix handle the above issue as well, right?

Yes, that's correct. The second unit test recreates the scenario where there are extra bytes after the end of the Deflate compressed data and passes on this branch. The same test on the `master` version of `httplib.h` will fail.

yhirose:
@bdenhollander thanks for your fine contribution!

Issue #1613, opened 2023-07-07
Compiler freezes on Debian 10 (buster) with GCC 8.3.0


Ben1138:
### Environment
OS: Debian 10 (buster)
Processor Arch: ARM
Compiler: GCC 8.3.0
CMake: 3.13.4

When trying to build my CMake project, as soon as GCC compiles my (only) compilation unit that `#include "httplib.h"`, I get these warnings:
![grafik](https://github.com/yhirose/cpp-httplib/assets/8263806/d9dd4178-6861-4f82-b23d-bf026f93663d)

and GCC just freezes and is unresponsive forever. 
On Windows x64 with MSVC, it builds just fine. 
I don't know whether this is an issue within httplib or the GCC compiler itself.

yhirose:
@Ben1138 thanks for the report.

@bgs99 @bugdea1er, it seems like it is caused by #1608 at the following line. Could you please take a look at it? Thanks!

<img width="749" alt="image" src="https://github.com/yhirose/cpp-httplib/assets/357397/2da21991-aed7-47b0-9e7e-138602063dc9">

bgs99:
I've reproduced the warning locally on docker with Debian 10 + GCC 8.3 when compiling the examples, but no compiler freezes.
The warning is valid and can be fixed by replacing this line with
```c++
request.matches = std::smatch();
```

@Ben1138 , could you check if it helps with the freeze? Otherwise I will probably need the minimal reproducible example.

yhirose:
@Ben1138 please let me and @bgs99 know if it still happens.

Ben1138:
The warning is gone, but it still freezes.
Here's a minimal example with which I can reproduce the freeze: [freeze.zip](https://github.com/yhirose/cpp-httplib/files/12011604/freeze.zip)

If someone can confirm or confute the freeze in the same environment, that'd be great.

bgs99:
Checked on Docker x64 Debian 10 + GCC 8.3, it takes a couple of seconds but builds fine, if I add the missing linking to pthread in CMakeLists.txt.

bgs99:
Checked on Docker ARM too, no freezes either.

PR #1614, opened 2023-07-07
Fix explicit constructor warning

bgs99:
This fixes the `converting to ... from initializer list would use explicit constructor` warning introduced in #1608 and reported in #1613 .


yhirose:
@bgs99 thanks for the quick resonse!

Issue #1600, opened 2023-06-26
Default headers after handling?


bugdea1er:
Why set default headers before handling and not after? Setting them after would allow just a simple assignment of a request.

Consider this example: I would like to extend a `httplib::Response` structure with a custom class:
```cpp
class JsonResponse final : public httplib::Response {
public:
  JsonResponse(int status, const nlohmann::json& content) {
    this->status = status;
    this->set_content(content, "application/json");
  }
};
```

and just respond with it in a handler:
```cpp
void handler(auto request, auto response) {
  response = JsonResponse(200, {})
}
```

Also I personally would love to see a possible handler signature in a more functional way
```cpp
httplib::Response handler(httplib::Request request)
```

but I cannot do this even with some binding stub because of default headers setting

yhirose:
@bugdea1er, thank you for the feedback. The reason why the default headers are set before a handler is called is because sometimes I need to make changes to the response object based on the default values within a handler. Therefore, the default header values should be accessible within a handler.

There is also a potential issue if the default headers are set after the handler is called, as the default values may mistakenly overwrite values set by the handler. Additionally, you can observe that there are multiple exit paths in Server::process_request. So, if we were to set the default header values afterwards, we would have to do so at all the locations where write_response is called, which would unnecessarily complicate the code.

The main reason for sticking to the current interface is that I want to avoid making any breaking changes that would surprise existing users. :)

Hope it helps!

bugdea1er:
@yhirose would you consider an optional (non-default) way to enable default handlers setting after handling?

As for overwriting, server could just check if the header already there. You are already doing this:
https://github.com/yhirose/cpp-httplib/blob/3533503323b22e48ef4875ace082118e4b69949a/httplib.h#L6270

As for multiple exit paths, all of them ends with a `write_response` call, so this wouldn't be hard to implement in a proper way.

I understand that you cannot make breaking changes, and I do not suggest making this a default behaviour

PR #1615, opened 2023-07-08
Added support for CPPHTTPLIB_DEFAULT_HEADERS_AFTER_HANDLING option

bugdea1er:
@yhirose, a simple implementation of what I was bothering you with in https://github.com/yhirose/cpp-httplib/issues/1600

This shouldn't break any previous code


yhirose:
@bugdea1er thanks for the pull request. But at this point, I am not interested in merging it, since I don't find any strong benefit from this change. Also users lose the opportunity to change header values in the default headers...

PR #1618, opened 2023-07-11
build(meson): use C++14 with GTest >= 1.13.0

Tachi107:
GoogleTest, starting with vesion 1.13.0, requires C++14 to build. This patch enables C++14 if GoogleTest 1.13.0 or newer is detected when compiling the tests with Meson, making it possible to use new GTest versions.

You can find GoogleTest's release notes at <https://github.com/google/googletest/releases/tag/v1.13.0>.


yhirose:
Thanks!

PR #1623, opened 2023-07-18
Add a Bazel BUILD and WORKSPACE file.

tsanti928:


yhirose:
Thanks for the pull request. But it doesn't look like it has the same quality as CMake and Meson supports... Could you take a look at what they are doing? Also would you willingly accept the responsibility to maintain this Bazel build support continuously from now on? If you agree these, then I'll consider merging it. Thanks for your consideration!

tsanti928:
That's true. I'll take a look at the requirements for feature parity. For now, I'll close this and may submit a new pull request in the future. Thanks!

PR #1624, opened 2023-07-20
export httplib::detail::status_message()

vmaffione:
This function can be useful to clients for error logging and reporting.


yhirose:
I made it public.

vmaffione:
great, thanks

PR #1630, opened 2023-07-26
Fix #1590

### LINKED ISSUE #1590: HTTP Client POST operation hangs in case of proxy error "Proxy Authentication Required"
I'm using httplib version 0.12.5

I configured a proxy with basic authentication and I try to use the httplib::Client::Post method with content provider.
It looks like `socket_mutex_` is locked when processing the request which causes the operation to hang, in case of invalid proxy credentials.

I think the issue is that the mutex is locked in `send_` method, but when calling `process_request` as a callback function from  `process_client_socket` we are still in the original `send_` scope so the mutex is still locked.

call stack: 
httplib::ClientImpl::process_request httplib.h:6922 ***
<lambda#1>::operator()(httplib::Stream &) const httplib.h:8009
std::__invoke_impl<…>(std::__invoke_other, <lambda#1> &, httplib::Stream &) invoke.h:61
std::__invoke_r<…>(<lambda#1> &, httplib::Stream &) invoke.h:114
std::_Function_handler::_M_invoke(const std::_Any_data &, httplib::Stream &) std_function.h:290
std::function::operator()(httplib::Stream &) const std_function.h:590
httplib::detail::process_client_socket(int, long, long, long, long, std::function<…>) httplib.h:2630 ***
httplib::SSLClient::connect_with_proxy httplib.h:8003 ***
httplib::ClientImpl::send_ httplib.h:6402 ***
httplib::ClientImpl::send httplib.h:6362
httplib::ClientImpl::send_with_content_provider(httplib::Request &, const char *, unsigned long, std::function<…>, std::function<…>, const std::string &, httplib::Error &) httplib.h:6809
httplib::ClientImpl::send_with_content_provider(const std::string &, const std::string &, const std::multimap<…> &, const char *, unsigned long, std::function<…>, std::function<…>, const std::string &) httplib.h:6826
httplib::ClientImpl::Post(const std::string &, const std::multimap<…> &, unsigned long, std::function<…>, const std::string &)
httplib::Client::Post(const std::string &, const std::multimap<…> &, unsigned long, std::function<…>, const std::string &)

Please have a look on this, thank you!

akinelden:
1. Moved `Connection: close` header control block from `process_request` to `handle_request` since the `process_request` is also called from `connect_with_proxy` method which causes a deadlock state.
2. Some improvements for proxy error handling:
   - Proxy connect response status control block added to stop the execution of the request when the proxy response is not successful. 
   - `ProxyConnection` error type added to distinguish proxy related unsuccessful response from standard response.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -6643,8 +6643,7 @@ inline bool ClientImpl::send_(Request &req, Response &res, Error &error) {
     }
 
     if (socket_should_be_closed_when_request_is_done_ || close_connection ||
-        !ret || res.get_header_value("Connection") == "close" ||
-        (res.version == "HTTP/1.0" && res.reason != "Connection established")) {
+        !ret) {
       shutdown_ssl(socket_, true);
       shutdown_socket(socket_);
       close_socket(socket_);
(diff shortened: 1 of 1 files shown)


akinelden:
@yhirose fixed the issue #1590, can you review the PR? Thanks.

yhirose:
@pappa5 could you try this pull request to see if it fixes your problem?

yhirose:
@akinelden I reviewed the code and it looks good to me. I confirmed that it fixes the issue #1590. Thanks for the fine contribution!

PR #1634, opened 2023-08-01
Fix CPPHTTPLIB_ALLOW_LF_AS_LINE_TERMINATOR

mrexodia:
In https://github.com/yhirose/cpp-httplib/commit/e5cacb465d3086c7d9be1d87dfffedbafeb02ce8 this issue was supposedly fixed, but in fact it introduced a bug and implemented it the wrong way.

The `\r?` is the default code path instead of specialized with `CPPHTTPLIB_ALLOW_LF_AS_LINE_TERMINATOR` as it should be.

In my opinion this behavior should be opt-out instead of opt-in (to match curl, which doesn't require any special options either) but that's probably something to discuss another time.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -6723,9 +6723,9 @@ inline bool ClientImpl::read_response_line(Stream &strm, const Request &req,
   if (!line_reader.getline()) { return false; }
 
 #ifdef CPPHTTPLIB_ALLOW_LF_AS_LINE_TERMINATOR
-  const static std::regex re("(HTTP/1\\.[01]) (\\d{3})(?: (.*?))?\r\n");
-#else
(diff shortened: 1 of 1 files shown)


yhirose:
@mrexodia thanks for find the bug! Could you leave `#ifdef` and swap the lines instead of using `#ifndef`? That's more readable to me. Thanks!

yhirose:
@mrexodia if you are busy to make the above change, I can handle it directly in the master branch. Please let me know . Thanks!

mrexodia:
Force pushed the change as you suggested just now. Thanks for the reminder!

mrexodia:
And thanks again for making this library. I use it quite often!

yhirose:
Thanks for your contribution!

PR #1639, opened 2023-08-14
Avoid a -Warray-bounds false positive in GCC 13.

svenpanne:
The exact circumstances when this false positive is triggered are quite tricky to reproduce, but it happened reproducibly with g++ 13.1 and 13.2 in a close-source SW I'm working on.  The fix even improves performance by a very tiny bit: There is no need to copy the std::smatch, having a const reference is enough.

Just as a side note: -Warray-bounds seems to cause trouble in other projects, too, so e.g. the Linux kernel has disabled since June 2022.


yhirose:
@svenpanne Thanks for the contribution!

PR #1643, opened 2023-08-19
Accept AbsoluteURI in Request Lines

hokacci:
RFC2616 says

> all HTTP/1.1 servers MUST accept the absoluteURI form in requests

https://datatracker.ietf.org/doc/html/rfc2616#section-5.1.2

I found the current implementation does not accept this form and fixed this issue by just ignoring unrelated parts (scheme, host, and etc) and extracting path part.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -5781,6 +5781,27 @@ inline bool Server::parse_request_line(const char *s, Request &req) {
   if (req.version != "HTTP/1.1" && req.version != "HTTP/1.0") { return false; }
 
   {
+    // Find path in URL
+    bool path_found = false;
+    for (size_t i = 0; i < req.target.size(); i++) {
+      if (req.target[i] == '/') {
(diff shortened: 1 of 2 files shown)


yhirose:
@hokacci thank you for the pull request. However, I am not accepting it.

Please see the context of the sentence.

> all HTTP/1.1 servers MUST accept the absoluteURI form in requests, even though **HTTP/1.1 clients will only generate them in requests to proxies**

I actually handle it when a request is made to a proxy at the following place:
https://github.com/yhirose/cpp-httplib/blob/afe627e7af13f78225076af1639c9ea40a807921/httplib.h#L6878-L6879

The path parameter in request methods such as GET and POST is specifically intended for representing path and fragments, not for indicating the scheme, host, and port. The scheme, host, and port should be defined within the client constructor.. 

In your implementation, it's possible to specify an unsupported scheme or specify a different host and port from the ones defined in the client. It doesn't make sense...

But thank you for your effort trying to make this library better!

PR #1644, opened 2023-08-20
Add move optimizations

trollixx:
A few move optimizations suggested by Coverity. There are a couple other places where lambda parameters can be moved, but IIRC that's only supported since C++14.


yhirose:
@trollixx thanks for the pull request, but it doesn't work on the Github Actions CI workflow.

trollixx:
@yhirose I've fixed the issue in the branch, but since the PR is closed it doesn't update.

PR #1648, opened 2023-08-23
use look table  in encode_query_param instead branching

SeanTolstoyevski:
this change implements the use of a look table instead of branching in the function.
this results in a noticeable performance improvement.

the following program can be used to test whether the table is created correctly.

```cpp
#include <array>
#include <iostream>

int main() {
static const std::array<bool, 256> lookTable{
        false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, true, false, false, false, false, false, true, true, true, true, false, false, true, true, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, true, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, true, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false, false
};

std::cout << lookTable['-'] << " " << lookTable['_']  << " " << lookTable['.'] << " " << lookTable['!'] << " " << lookTable['~'] << " " << lookTable['*'] <<" " <<  lookTable['\''] << " "  << lookTable['('] << " " << lookTable[')'] << std::endl;
return 0;
}
```

output:

`1 1 1 1 1 1 1 1 1`

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -2327,11 +2327,37 @@ inline std::string encode_query_param(const std::string &value) {
   std::ostringstream escaped;
   escaped.fill('0');
   escaped << std::hex;
+  static const std::array<bool, 256> lookTable{
+      false, false, false, false, false, false, false, false, false, false,
+      false, false, false, false, false, false, false, false, false, false,
+      false, false, false, false, false, false, false, false, false, false,
+      false, false, false, true,  false, false, false, false, false, true,
+      true,  true,  true,  false, false, true,  true,  false, false, false,
+      false, false, false, false, false, false, false, false, false, false,
+      false, false, false, false, false, false, false, false, false, false,
+      false, false, false, false, false, false, false, false, false, false,
+      false, false, false, false, false, false, false, false, false, false,
+      false, false, false, false, false, true,  false, false, false, false,
+      false, false, false, false, false, false, false, false, false, false,
+      false, false, false, false, false, false, false, false, false, false,
+      false, false, false, false, false, false, true,  false, false, false,
+      false, false, false, false, false, false, false, false, false, false,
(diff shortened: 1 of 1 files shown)


yhirose:
@SeanTolstoyevski thanks for the pull request. I still prefer the original code since it's more readable than this lookup table. I would like to know how much performance improvement you achieved with this change compared to the original one. Could you show the actual statistic report on your machine? Thanks!

> this results in a noticeable performance improvement.

SeanTolstoyevski:
@yhirose hi,

If you wish, I can comment the indexes that are true to make it more understandable (know which character has an index).

I will prepare performance data and share it as soon as possible.

SeanTolstoyevski:
hi,

as a result of my analysis, the result was not as good as i expected.
that would be too much of a change for a performance boost.

i will close this pr.

PR #1681, opened 2023-09-28
fix make_content_range_header_field start&end default value & fix str…

sinnren:
in make_content_range_header_field ,if range.second==-1, return value will be "bytes 0-/1024"(content-length is 1024),mp4 stream in chrome not played, change to "bytes 0-1023/1024"(content-length - 1) worked.

when range.first==-1,range start will be 0,so the fix the default value to (0, content - 1).

some string param fix....

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -3310,7 +3310,7 @@ find_content_type(const std::string &path,
   auto ext = file_extension(path);
 
   auto it = user_data.find(ext);
-  if (it != user_data.end()) { return it->second.c_str(); }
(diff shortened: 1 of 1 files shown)


yhirose:
@sinnren thank you for the pull request. Could you please add some unit tests in `test/test.cc`?

yhirose:
@sinnren what happens?

sinnren:
I read the http protocol again, this pr is something wrong, some test failed. I'll fix this and pass all test . then->re pr.
sorry my fault.

PR #1683, opened 2023-09-30
Add optional user defined header writer

PabloMK7:
Adds the ability to redefine the `write_header` function to implement custom header writers.

Context: 
httplib is used in the Nintendo 3DS [Citra Emulator](https://github.com/citra-emu/citra). One of the functionalities of the Nintendo 3DS is a built-in http "system module" as part of the OS which allows games to perform http request without having to implement all the http stack themselves.

This "system module" has a very specific behaviour regarding header ordering. Also, httplib is adding some headers by default (such as `Connection` and `Content-type`) which the 3DS "system module" doesn't do. In order to properly emulate the behaviour of the Nintendo 3DS, we need to be able to control header ordering and generation, and having a custom writer function feels like the best option.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -737,6 +737,8 @@ private:
   std::regex regex_;
 };
 
+ssize_t write_headers(Stream &strm, const Headers &headers);
+
 } // namespace detail
 
 class Server {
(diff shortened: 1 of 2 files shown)


yhirose:
@PabloMK7 it sounds good to me, and the code looks good as well. Could you add at least one unit test for the `Server::set_header_writer`? Then, I'll merge it. Thanks!

PabloMK7:
Good that I implemented the test, because there were some compilation issues that I didn't expect. I have implemented a test and it is passing!

yhirose:
@PabloMK7 thanks for your contribution! v0.14.1 has been released.

yhirose:
@PabloMK7 I made a mistake when releasing v0.14.1, and did it again.

PR #1694, opened 2023-10-18
check request range and fix response Content-Range.

sinnren:
Adjusted content 

The return value of the "apply_ranges" in the "write_response_core" is adjusted to bool. In the "apply_ranges", the validity of the requested range field is verified and saved to res.ranges_offset_length_ (new field) (using the original get_range_offset_and_length logic). Caching this fields are also needed for subsequent code and do not need to be processed again.

If "apply_ranges" returns false, it means that the range is wrong. Return 416 directly and clear the content.
The subsequent code only removes the offset verification, because all offsets and lengths have been verified in "apply_ranges".

Adjusted content in "make_content_range_header_field"
In the Content-Range field of the return value, if the request range-start or range-end is empty(-1),the offset and length of the actual returned content must be filled in, otherwise the browser will not work. (Use the Chrome view mp4 file, and the request is Range:1024-. If the return field does not fill in the range-end, such as "bytes 1024-/3124123", it cannot be played normally)

get_range_offset_and_length method is no longer needed, so removed

Several range processing logics are:
Assuming the content length is 10,
range:0-9 -> bytes 0-9/10, content-length:10
range:-3 -> bytes 7-9/10, content-length:3 (last 3 bytes)
range:-99 -> bytes 0-9/10, content-length:10 (fix length to 10, last 10 bytes)
range:0- -> bytes 0-9/10, content-length:10
range:0-99 -> bytes 0-9/10, content-length:10 (fix length to 10)
range:10- -> error:416 (range error)
range:10-99 -> error:416 (range error)

fix some string param(without .c_str())

ps.translated by google.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -556,6 +556,7 @@ struct Response {
   ContentProviderResourceReleaser content_provider_resource_releaser_;
   bool is_chunked_content_provider_ = false;
   bool content_provider_success_ = false;
+  std::vector<std::pair<size_t, size_t>> ranges_offset_length_;
 };
 
 class Stream {
@@ -876,7 +877,7 @@ private:
                                       const HandlersForContentReader &handlers);
 
   bool parse_request_line(const char *s, Request &req);
-  void apply_ranges(const Request &req, Response &res,
+  bool apply_ranges(const Request &req, Response &res,
                     std::string &content_type, std::string &boundary);
   bool write_response(Stream &strm, bool close_connection, const Request &req,
                       Response &res);
@@ -4593,45 +4594,25 @@ serialize_multipart_formdata(const MultipartFormDataItems &items,
   return body;
 }
 
-inline std::pair<size_t, size_t>
-get_range_offset_and_length(const Request &req, size_t content_length,
-                            size_t index) {
-  auto r = req.ranges[index];
-
-  if (r.first == -1 && r.second == -1) {
-    return std::make_pair(0, content_length);
-  }
-
-  auto slen = static_cast<ssize_t>(content_length);
-
-  if (r.first == -1) {
-    r.first = (std::max)(static_cast<ssize_t>(0), slen - r.second);
-    r.second = slen - 1;
-  }
-
-  if (r.second == -1) { r.second = slen - 1; }
-  return std::make_pair(r.first, static_cast<size_t>(r.second - r.first) + 1);
-}
-
 inline std::string
-make_content_range_header_field(const std::pair<ssize_t, ssize_t> &range,
+make_content_range_header_field(const std::pair<size_t, size_t> &offset_length,
                                 size_t content_length) {
   std::string field = "bytes ";
-  if (range.first != -1) { field += std::to_string(range.first); }
+  field += std::to_string(offset_length.first);
   field += "-";
-  if (range.second != -1) { field += std::to_string(range.second); }
+  field += std::to_string(offset_length.first + offset_length.second - 1);
   field += "/";
   field += std::to_string(content_length);
   return field;
 }
 
 template <typename SToken, typename CToken, typename Content>
-bool process_multipart_ranges_data(const Request &req, Response &res,
+bool process_multipart_ranges_data(const Request &/*req*/, Response &res,
                                    const std::string &boundary,
                                    const std::string &content_type,
                                    SToken stoken, CToken ctoken,
                                    Content content) {
-  for (size_t i = 0; i < req.ranges.size(); i++) {
+  for (size_t i = 0; i < res.ranges_offset_length_.size(); i++) {
     ctoken("--");
     stoken(boundary);
     ctoken("\r\n");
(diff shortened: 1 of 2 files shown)


yhirose:
@sinnren thanks for the pull request. Since it's a pretty big change, could you add enough unit tests in `test/test.cc`?

sinnren:
> @sinnren thanks for the pull request. Since it's a pretty big change, could you add enough unit tests in `test/test.cc`?

16 lines added in test.cc, need more?

yhirose:
@sinnren what kind of changes did you make in `test/test.cc`? It says 'Binary file'...

image

> 16 lines added in test.cc, need more?

I can't see your test code due to the above 'binary' file problem. Anyway, could you cover all the normal cases and edge cases?

sinnren:
My mistake was that the encoding of the test.cc file was changed and turned into a binary file. It has been adjusted.

httplib.h
```diff
@@ -454,11 +454,50 @@ class ContentReader {
...
+    return std::string();
+  }
+}
+
 using Range = std::pair<ssize_t, ssize_t>;
 using Ranges = std::vector<Range>;
 
 struct Request {
-  std::string method;
+  RequestMethod method = UNKNOWN;
```
yhirose, line 500: This is a breaking change and will cause build errors of existing users's code using `send` method. Could you change back `method` to `std::string` to `RequestMethod` please?

sinnren: oh ， I'll revert this change.

sinnren: reverted the breaking chage.

httplib.h
```diff
@@ -524,8 +524,8 @@ struct Response {
   void set_header(const std::string &key, const std::string &val);
 
   void set_redirect(const std::string &url, int status = 302);
-  void set_content(const char *s, size_t n, const std::string &content_type);
-  void set_content(const std::string &s, const std::string &content_type);
+  void set_content(const char *s, size_t n, const std::string &content_type = {});
```
yhirose, line 527: Could you remove the default parameter `= {}` please?

httplib.h
```diff
@@ -524,8 +524,8 @@ struct Response {
...
+  void set_content(const std::string &s, const std::string &content_type = {});
```
yhirose, line 528: Could you remove the default parameter `= {}` as well please?

httplib.h
```diff
@@ -786,6 +787,7 @@ class Server {
   Server &set_file_extension_and_mimetype_mapping(const std::string &ext,
                                                   const std::string &mime);
   Server &set_default_file_mimetype(const std::string &mime);
+  Server &set_default_content_mimetype(const std::string &mime);
```
yhirose, line 790: Could you remove `set_default_content_mimetype` please? You can see the default with `set_default_headers`.

httplib.h
yhirose: You don't have to use `set_content` method at all if you want to set the content text only. You can simply do it by `res.body = s` `with std::string` or `res.body.assign(s, n)` with `const char*`.

https://github.com/yhirose/cpp-httplib/blob/20a7f088cea4df658ed66b17153148b75e64881e/httplib.h#L5220

sinnren:
copy that.

yhirose:
@sinnren I made changes to remove unnecessary `.c_str()` calls on the master branch. So could you please rebase your code? Then, I think this pull request will end up including only changes directly related to the Content-Range. After that, I'll start review your pull request.

sinnren:
> @sinnren I made changes to remove unnecessary `.c_str()` calls on the master branch. So could you please rebase your code? Then, I think this pull request will end up including only changes directly related to the Content-Range. After that, I'll start review your pull request.


ok.

test/test.cc
```diff
@@ -2549,7 +2550,7 @@ TEST_F(ServerTest, StaticFileRanges) {
           .find(
               "multipart/byteranges; boundary=--cpp-httplib-multipart-data-") ==
       0);
-  EXPECT_EQ("265", res->get_header_value("Content-Length"));
+  EXPECT_EQ("266", res->get_header_value("Content-Length"));
```
yhirose, line 2553: Why did you change the value to '266'?

sinnren: the request is cli_.Get("/dir/test.abcde", {{make_range_header({{1, 2}, {4, -1}})}});
this range {4, -1}

before:
ret range is "bytes 4-/5" , it was wrong because the range-end is empty(in method make_content_range_header_field).


after:
ret range is "bytes4-4/5",fix the range-end.(the purpose of the PR ) 

so.add 1 byte.

full body is:

`
----cpp-httplib-multipart-data-REkZlTo5uvw80k3K
Content-Type: text/abcde
Content-Range: bytes 1-2/5

bc
----cpp-httplib-multipart-data-REkZlTo5uvw80k3K
Content-Type: text/abcde
Content-Range: bytes 4-4/5

e
----cpp-httplib-multipart-data-REkZlTo5uvw80k3K--
`

sinnren:
Any further adjustments needed?

yhirose:
@sinnren thanks for your responses. I reviewed the changes, however, I am not comfortable with `ranges_offset_length_` in `Response`. So I resolved this issue in a simpler way at 420c975. Thank you anyway for finding this bug.

PR #1695, opened 2023-10-20
build(meson): copy 1MB.txt test file

Tachi107:
Since tests are run in the build directory, the 1MB.txt file has to be copied there.


yhirose:
Thanks!

PR #1700, opened 2023-10-27
Include missing stdint.h on fuzz test

jimmy-park:
Building fuzz test fails on Ubuntu workflow due to missing `stdint.h` and `std::`.

fuzz


yhirose:
@jimmy-park could you try without adding `std::` for `uint8_t`?

jimmy-park:
@yhirose It seems to work without `std::`.

yhirose:
Thanks!

PR #1701, opened 2023-10-27
Fix warnings from MSVC code analysis

jimmy-park:
[Warning C6001](https://learn.microsoft.com/en-us/cpp/code-quality/c6001) : Using uninitialized memory
[Warning C26444](https://learn.microsoft.com/en-us/cpp/code-quality/c26444) : Don't try to declare a local variable with no name


yhirose:
Could you rebase your pull request and remove the change for the exception? It was my mistake, and std::runtime_error should be removed. Thanks.

yhirose:
Thanks!

PR #1707, opened 2023-11-09
Run fuzz test in CTest

jimmy-park:
Add missing fuzz tests when building with CMake.


yhirose:
Thanks!

Fix3dP0int:
It seems that the sanitizer flags were lost. The compiling process isn't same as the compiling in oss-fuzz using `Makefile`.

PR #1709, opened 2023-11-11
Minor fixes on test cases

jimmy-park:
- `SSLClientServerTest.SSLConnectTimeout` has a data race when stopping the server. (detected by ThreadSanitizer)
- After https://github.com/yhirose/cpp-httplib/commit/7aba2938d3e7efb36aed8eba606200e949c20f03, we don't need to wait the server using `std::this_thread::sleep_for()`. Just use `Server::wait_until_ready()`.


yhirose:
Thanks for simplifying the code.

Issue #1703, opened 2023-11-01
Mingw cross-compilation -lcryptui not found.


montarcannon:
I'm trying to cross-compile small project that uses cpp-httplib with openssl using cmake and mingw. 
Cmake runs fine but make fails with following output:
`
[ 50%] Linking CXX executable cr_persist.exe
/usr/bin/i686-w64-mingw32-ld: cannot find -lcryptui: No such file or directory
collect2: error: ld returned 1 exit status
make[2]: *** [CMakeFiles/cr_persist.dir/build.make:106: cr_persist.exe] Error 1
make[1]: *** [CMakeFiles/Makefile2:100: CMakeFiles/cr_persist.dir/all] Error 2
make: *** [Makefile:136: all] Error 2
`
probably i'm missing some library but i don't know which.

yhirose:
@sum01, @jimmy-park  do you have any clue?

jimmy-park:
I have successfully built with MinGW-w64 on my [CMake workflow](https://github.com/jimmy-park/cpp-httplib/actions/runs/6723104501/job/18272437145).
It seems like `cryptui.dll` is missing on your machine.

jimmy-park:
Do we really need to link the `cryptui` library on Windows? I couldn't find any method provided by `cryptui`.
And removing `cryptui` doesn't affect on building with `httplib.h`.

yhirose:
@jimmy-park, thanks for the answers. @montarcannon, hope the above inform helps.

montarcannon:
> I have successfully built with MinGW-w64 but failed with MinGW32. Please see my [CMake workflow](https://github.com/jimmy-park/cpp-httplib/actions/runs/6723743785/job/18274474044). It seems like `cryptui.dll` is missing on MinGW32.

you ran into same problems as i did

PR #1710, opened 2023-11-11
Remove `cryptui` on Windows

jimmy-park:
`cryptui` library is linked when building on Windows, but isn't actually used. If `cryptui.lib` doesn't exist on some Windows platforms (especially MinGW32), building with `httplib.h` will fail.
I have tested without linking `cryptui` on the various Windows environments (MSVC, MinGW, Cygwin), and it works fine.

Related to #1703.


yhirose:
Thanks!

PR #1713, opened 2023-11-16
Fix: Query parameter including query delimiter ('?') not being parsed properly

davidalo:
When query delimiter `?` is contained inside the query parameters, split function is cutting it and query parameters and not being parsed properly.

As an example:

```
/regex-with-delimiter?key=^(?.*(value))
```

This PR fixes it by adding `m` (maximum) parameter to `details::split`. This parameter indicates the maximum number of parts which will be got when splitting: `-1` (infinite, no max), >= 0.

In `Server::parse_request_line` we use it as m=2 so we only get two parts: url and query parameters. By this way, we avoid splitting query parameters if they contain `?`.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -1988,8 +1988,9 @@ void read_file(const std::string &path, std::string &out);
 
 std::string trim_copy(const std::string &s);
 
-void split(const char *b, const char *e, char d,
+void split(const char *b, const char *e, char d, int m,
            std::function<void(const char *, const char *)> fn);
+
 
 bool process_client_socket(socket_t sock, time_t read_timeout_sec,
                            time_t read_timeout_usec, time_t write_timeout_sec,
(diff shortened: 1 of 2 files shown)


yhirose:
@davidalo thank you very for your pull request. The change looks good to me. Could you make a couple of minor adjustments to make the code more readable?

1.  Change `int` for `m` to `size_t`
2. Change `-1` to `std::numeric_limits<size_t>::max()`
3.  Make the following helper function to pass `std::numeric_limits<size_t>::max()`
```
inline void split(const char *b, const char *e, char d,
                  std::function<void(const char *, const char *)> fn) {
  split(b, e, d, std::numeric_limits<size_t>::max(), fn);
}
```
4. Change all the callers back to originals
5. Change the line `if (b[i] == d && (m < 0 || count < m)) {` to `if (b[i] == d && count < m) {`

yhirose:
@davidalo thanks a lot for the fine contribution!

PR #1715, opened 2023-11-20
Fix performance-noexcept-move-constructor

jfsimoneau:
Fix clang-tidy warning for rule performance-noexcept-move-constructor


yhirose:
Thanks!

PR #1716, opened 2023-11-20
Fix performance-inefficient-string-concatenation

jfsimoneau:


httplib.h
```diff
@@ -5434,8 +5434,15 @@ inline PathParamsMatcher::PathParamsMatcher(const std::string &pattern) {
 
 #ifndef CPPHTTPLIB_NO_EXCEPTIONS
     if (param_name_set.find(param_name) != param_name_set.cend()) {
-      std::string msg = "Encountered path parameter '" + param_name +
```
yhirose, line 5437: I prefer the original one line code. It's much easier to read, and it gives very little performance gain.

PR #1717, opened 2023-11-20
Fix performance-unnecessary-value-param

jfsimoneau:
Also fixes related performance-move-const-arg

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -789,7 +789,7 @@ public:
   Server &set_file_request_handler(Handler handler);
 
   Server &set_error_handler(HandlerWithResponse handler);
-  Server &set_error_handler(Handler handler);
+  Server &set_error_handler(const Handler &handler);
   Server &set_exception_handler(ExceptionHandler handler);
   Server &set_pre_routing_handler(HandlerWithResponse handler);
   Server &set_post_routing_handler(Handler handler);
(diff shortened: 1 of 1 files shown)


httplib.h
```diff
@@ -788,7 +788,7 @@ class Server {
   Server &set_file_request_handler(Handler handler);
 
   Server &set_error_handler(HandlerWithResponse handler);
-  Server &set_error_handler(Handler handler);
+  Server &set_error_handler(const Handler &handler);
```
yhirose, line 791: I am not comfortable with changes like this. I intentionally made the parameter as a value as other set handler methods are also taking value params. It creates inconsistency and causes a confusion...

jfsimoneau: I was hesitant about this change too. I feel all should be const references or all should always be moved. Personally, I prefer const references for their low overhead, but since the other calls are perfect as they are, I don't want to change them. I will revert this one.

jfsimoneau: I made a compile-time conditional check to use C++14 init captures if available. This silences the warning on C++14 by moving the handler value into the lambda.

jfsimoneau:
Hey @yhirose, is there anything else I can do to move this forward? Thanks!

yhirose:
@jfsimoneau I really appreciate your changes though, I am not very interested in them and I still prefer the existing code. I'll most likely close this pull request, but keep it open for while just in case I'll change my mind. Thanks for your understanding.

PR #1722, opened 2023-11-21
Fix readability warnings

jfsimoneau:
Did not fix readbility-qualified-auto, will do a separate pull request

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -4727,10 +4727,12 @@ get_range_offset_and_length(const Request &req, const Response &res,
 }
 
 inline bool expect_content(const Request &req) {
+  if (req.method == "POST" || req.method == "PUT" || req.method == "PATCH" ||
+      req.method == "PRI" || req.method == "DELETE") {
+    return true;
+  }
   // TODO: check if Content-Length is set
-  return (req.method == "POST" || req.method == "PUT" ||
-          req.method == "PATCH" || req.method == "PRI" ||
-          req.method == "DELETE");
+  return false;
 }
 
 inline bool has_crlf(const std::string &s) {
@@ -8357,7 +8359,7 @@ inline SSLClient::SSLClient(const std::string &host, int port,
     : ClientImpl(host, port, client_cert_path, client_key_path) {
   ctx_ = SSL_CTX_new(TLS_client_method());
 
-  detail::split(host_.data(), &host_[host_.size()], '.',
+  detail::split(&host_[0], &host_[host_.size()], '.',
                 [&](const char *b, const char *e) {
                   host_components_.emplace_back(b, e);
                 });
--- a/httplib.h
+++ b/httplib.h
@@ -868,15 +868,16 @@ private:
   bool routing(Request &req, Response &res, Stream &strm);
   bool handle_file_request(const Request &req, Response &res,
                            bool head = false);
-  bool dispatch_request(Request &req, Response &res, const Handlers &handlers);
-  bool
+  static bool dispatch_request(Request &req, Response &res,
+                               const Handlers &handlers);
+  static bool
   dispatch_request_for_content_reader(Request &req, Response &res,
                                       ContentReader content_reader,
                                       const HandlersForContentReader &handlers);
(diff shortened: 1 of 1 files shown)


httplib.h
```diff
@@ -4724,12 +4727,10 @@ get_range_offset_and_length(const Request &req, const Response &res,
 }
 
 inline bool expect_content(const Request &req) {
-  if (req.method == "POST" || req.method == "PUT" || req.method == "PATCH" ||
-      req.method == "PRI" || req.method == "DELETE") {
-    return true;
-  }
   // TODO: check if Content-Length is set
-  return false;
+  return (req.method == "POST" || req.method == "PUT" ||
```
yhirose, line 4731: This is incorrect. I intentionally put the if statement before the TODO comment.

httplib.h
```diff
@@ -868,15 +868,16 @@ class Server {
   bool routing(Request &req, Response &res, Stream &strm);
   bool handle_file_request(const Request &req, Response &res,
                            bool head = false);
-  bool dispatch_request(Request &req, Response &res, const Handlers &handlers);
-  bool
+  static bool dispatch_request(Request &req, Response &res,
```
yhirose, line 871: I don't like to put 'static'. It's less readable.

httplib.h
jfsimoneau: Ah, I see. Something will come after the first check. I will revert this part.

httplib.h
```diff
@@ -8361,7 +8363,7 @@ inline SSLClient::SSLClient(const std::string &host, int port,
     : ClientImpl(host, port, client_cert_path, client_key_path) {
   ctx_ = SSL_CTX_new(TLS_client_method());
 
-  detail::split(&host_[0], &host_[host_.size()], '.',
+  detail::split(host_.data(), &host_[host_.size()], '.',
```
yhirose, line 8366: If we do `host.data()`, should we do the same to `&host[host.size()]` like `host_data() + host.size()` This change actually creates inconsistency and made the code less readable...

yhirose:
@jfsimoneau I basically don't agree half of changes like adding `static`...

httplib.h
jfsimoneau: I understand your feeling, but having a static method means the `this` pointer is not passed as an automatic parameter. This will reduce the call overhead a little bit and ensure the calling instance won't affect the outcome or be affected by any side-effect.

httplib.h
jfsimoneau: Good point, I overlooked this case.

yhirose: I am ok with the notation like `&a[0]`. IMHO, most of C/C++ programmers know what it means, and it's a common notation.

jfsimoneau: I agree with you, the meaning is very well-known. In this particular case, it even makes more sense to use this notation because of the begin/end parameters of `detail::split()` for iteration. In the other cases, the meaning really is to put data into a vector's buffer.

yhirose:
Could you leave only changes for adding missing `const`, removing unnecessary `const, and adding missing curly braces? As for `static`, it still make me confused when reading the class. I know what you mentioned, but I would like to keep the code as it is. Thanks for your understanding!

jfsimoneau:
Sure, on it!

jfsimoneau:
Replaced all `static` methods with `const` methods to emphasize the absence of side-effects. Also reverted the `.data()` accessors to leave `&var[0]` so it's consistent everywhere.

I did notice most `static` methods could also be hidden internal functions or helper functions instead of class methods, except some internal class member access like `private` classes. Worth noting, but I won't touch that; if it's not broken, don't fix it.

yhirose:
Thanks for all the improvements.

> I did notice most static methods could also be hidden internal functions or helper functions instead of class methods, except some internal class member access like private classes. Worth noting, but I won't touch that; if it's not broken, don't fix it.

When I have time, I'll take a look at the code and thing about what I can do. Thanks again for your fine contribution!

PR #1725, opened 2023-11-23
Fix readbility-qualified-auto

jfsimoneau:


httplib.h
```diff
@@ -2216,7 +2216,7 @@ inline bool from_hex_to_i(const std::string &s, size_t i, size_t cnt,
 }
 
 inline std::string from_i_to_hex(size_t n) {
-  static const auto charset = "0123456789abcdef";
+  static const auto *const charset = "0123456789abcdef";
```
yhirose, line 2219: Adding `const` is ok, but I don't like to leave `*`, since I do the same for shared pointers as below.

```cpp
auto mm = std::make_shared<detail::mmap>(path.c_str());
```

jfsimoneau: In the second case. `mm` is not a raw pointer, though. Here, I could change the pointer to either a C-style array or an `std::array`. I'll try a few things to see which one would be more elegant and still be compliant.

yhirose: My saying is that I would like to keep the code as always using just `auto` regardless of regular value, raw pointer or managed pointer (std::shared_ptr, std::unique_ptr). Thanks for your effort, but I'l close it. :)

PR #1726, opened 2023-11-27
include string_view if compiler supports it

artpaul:
With this feature enabled the headers can be queried with key of type std::string_view without need to convert it to std::string.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -113,6 +113,10 @@
 #define CPPHTTPLIB_LISTEN_BACKLOG 5
 #endif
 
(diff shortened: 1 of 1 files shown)


artpaul:
@yhirose That do you think about to enable string_view support?

https://en.cppreference.com/w/cpp/container/multimap/find

yhirose:
@artpaul thanks for the pull request. But  if we do it, we have to do it at all the places where std::string is used as the code consistency and it's too much. I would like to stick to C++11 at this point. Thank you for your understanding.

artpaul:
OK. Got it.

PR #1739, opened 2023-12-18
Add StatusCode enum

bugdea1er:
Hi there!

C++ Core Guidelines recommend avoiding "magic constants" ([link](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es45-avoid-magic-constants-use-symbolic-constants)) and clang-tidy enforces this. I added a enum with Response Status Codes to use them instead of plain numbers in dependent code, please let me know if you would like to use them in httplib too.

~~Also I changed some of status messages based on [RFC 9110](https://httpwg.org/specs/rfc9110.html#overview.of.status.codes)~~ 
Moved RFC 9110 changes to a separate pr https://github.com/yhirose/cpp-httplib/pull/1740

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -1791,6 +1791,81 @@ inline void default_socket_options(socket_t sock) {
 #endif
 }
 
+enum StatusCode {
+  // Information responses
+  Continue = 100,                      ///< 100 Continue
+  SwitchingProtocol = 101,             ///< 101 Switching Protocols
+  Processing = 102,                    ///< 102 Processing
+  EarlyHints = 103,                    ///< 103 Early Hints
+
+  // Successful responses
+  OK = 200,                            ///< 200 OK
+  Created = 201,                       ///< 201 Created
+  Accepted = 202,                      ///< 202 Accepted
+  NonAuthoritativeInformation = 203,   ///< 203 Non-Authoritative Information
+  NoContent = 204,                     ///< 204 No Content
+  ResetContent = 205,                  ///< 205 Reset Content
+  PartialContent = 206,                ///< 206 Partial Content
+  MultiStatus = 207,                   ///< 207 Multi-Status
+  AlreadyReported = 208,               ///< 208 Already Reported
+  IMUsed = 226,                        ///< 226 IM Used
+
+  // Redirection messages
+  MultipleChoices = 300,               ///< 300 Multiple Choices
+  MovedPermanently = 301,              ///< 301 Moved Permanently
+  Found = 302,                         ///< 302 Found
+  SeeOther = 303,                      ///< 303 See Other
+  NotModified = 304,                   ///< 304 Not Modified
+  UseProxy = 305,                      ///< 305 Use Proxy
+  unused = 306,                        ///< 306 unused
+  TemporaryRedirect = 307,             ///< 307 Temporary Redirect
+  PermanentRedirect = 308,             ///< 308 Permanent Redirect
+
+  // Client error responses
+  BadRequest = 400,                    ///< 400 Bad Request
+  Unauthorized = 401,                  ///< 401 Unauthorized
+  PaymentRequired = 402,               ///< 402 Payment Required
+  Forbidden = 403,                     ///< 403 Forbidden
+  NotFound = 404,                      ///< 404 Not Found
+  MethodNotAllowed = 405,              ///< 405 Method Not Allowed
+  NotAcceptable = 406,                 ///< 406 Not Acceptable
+  ProxyAuthenticationRequired = 407,   ///< 407 Proxy Authentication Required
+  RequestTimeout = 408,                ///< 408 Request Timeout
+  Conflict = 409,                      ///< 409 Conflict
+  Gone = 410,                          ///< 410 Gone
+  LengthRequired = 411,                ///< 411 Length Required
+  PreconditionFailed = 412,            ///< 412 Precondition Failed
+  PayloadTooLarge = 413,               ///< 413 Payload Too Large
+  UriTooLong = 414,                    ///< 414 URI Too Long
+  UnsupportedMediaType = 415,          ///< 415 Unsupported Media Type
+  RangeNotSatisfiable = 416,           ///< 416 Range Not Satisfiable
+  ExpectationFailed = 417,             ///< 417 Expectation Failed
+  ImATeapot = 418,                     ///< 418 I'm a teapot
+  MisdirectedRequest = 421,            ///< 421 Misdirected Request
+  UnprocessableContent = 422,          ///< 422 Unprocessable Content
+  Locked = 423,                        ///< 423 Locked
+  FailedDependency = 424,              ///< 424 Failed Dependency
+  TooEarly = 425,                      ///< 425 Too Early
+  UpgradeRequired = 426,               ///< 426 Upgrade Required
+  PreconditionRequired = 428,          ///< 428 Precondition Required
+  TooManyRequests = 429,               ///< 429 Too Many Requests
+  RequestHeaderFieldsTooLarge = 431,   ///< 431 Request Header Fields Too Large
+  UnavailableForLegalReasons = 451,    ///< 451 Unavailable For Legal Reasons
+
+  // Server error responses
+  InternalServerError = 500,           ///< 500 Internal Server Error
+  NotImplemented = 501,                ///< 501 Not Implemented
+  BadGateway = 502,                    ///< 502 Bad Gateway
+  ServiceUnavailable = 503,            ///< 503 Service Unavailable
+  GatewayTimeout = 504,                ///< 504 Gateway Timeout
+  HttpVersionNotSupported = 505,       ///< 505 HTTP Version Not Supported
+  VariantAlsoNegotiates = 506,         ///< 506 Variant Also Negotiates
+  InsufficientStorage = 507,           ///< 507 Insufficient Storage
+  LoopDetected = 508,                  ///< 508 Loop Detected
+  NotExtended = 510,                   ///< 510 Not Extended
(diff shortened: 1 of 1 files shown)


httplib.h
```diff
@@ -1807,7 +1882,7 @@ inline const char *status_message(int status) {
   case 207: return "Multi-Status";
   case 208: return "Already Reported";
   case 226: return "IM Used";
-  case 300: return "Multiple Choice";
+  case 300: return "Multiple Choices";
```
bugdea1er, line 1885: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/300

httplib.h
```diff
@@ -1836,7 +1911,7 @@ inline const char *status_message(int status) {
   case 417: return "Expectation Failed";
   case 418: return "I'm a teapot";
   case 421: return "Misdirected Request";
-  case 422: return "Unprocessable Entity";
+  case 422: return "Unprocessable Content";
```
bugdea1er, line 1914: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422

httplib.h
bugdea1er, line 1885: https://httpwg.org/specs/rfc9110.html#status.300

httplib.h
bugdea1er, line 1914: https://httpwg.org/specs/rfc9110.html#status.422

bugdea1er:
Also I would suggest to use this new type for `Response::status` field, but I don't know if it would be a breaking change

yhirose:
> Also I changed some of status messages based on [RFC 9110](https://httpwg.org/specs/rfc9110.html#overview.of.status.codes)

Could you move the change to a separate pull request?

> I would suggest to make this enum class and use this new type for Response::status field, but I don't know if it would be a breaking change

It's a absolutely big breaking change. I understand it's a good practice to enum values instead of hardcoded numbers though, I'll keep the current code as it is since I would not like existing users to be surprised and modify their existing code. Thanks for your understanding.

bugdea1er:
> Could you move the change to a separate pull request? I'll merge it right away!

https://github.com/yhirose/cpp-httplib/pull/1740

> It's a absolutely big breaking change. I understand it's a good practice to enum values instead of hardcoded numbers though, I'll keep the current code as it is since I would not like to break the existing users' code at all. Thanks for your understanding.

So a plane enum (like in this pr) is okay with you? All the types stays `int` and users can use this constats without breaking their code?

yhirose:
> Btw, I checked with our existing code, compiles fine before and after using this enum in Request::status assignments. C++ allows implicit enum to int conversion.

Here are compile errors which shows the breaking change when compiling 'test/test.cc'.

```cpp
  ...
  LoopDetected = 508,                  ///< 508 Loop Detected
  NotExtended = 510,                   ///< 510 Not Extended
  NetworkAuthenticationRequired = 511, ///< 511 Network Authentication Required
};

struct Response {
  std::string version;
  // int status = -1;
  StatusCode status = -1;
  std::string reason;
  Headers headers;
  ...
```

```
~/Projects/cpp-httplib/test$ make
clang++ -o test -I.. -g -std=c++11 -I. -Wall -Wextra -Wtype-limits -Wconversion -Wshadow  test.cc include_httplib.cc gtest/gtest-all.cc gtest/gtest_main.cc -DCPPHTTPLIB_OPENSSL_SUPPORT -I/usr/local/opt/openssl@1.1/include -L/usr/local/opt/openssl@1.1/lib -lssl -lcrypto -DCPPHTTPLIB_USE_CERTS_FROM_MACOSX_KEYCHAIN -framework CoreFoundation -framework Security -DCPPHTTPLIB_ZLIB_SUPPORT -lz -DCPPHTTPLIB_BROTLI_SUPPORT -I/usr/local/opt/brotli/include -L/usr/local/opt/brotli/lib -lbrotlicommon -lbrotlienc -lbrotlidec -pthread
In file included from test.cc:1:
../httplib.h:590:23: error: cannot initialize a member subobject of type 'StatusCode' with an rvalue of type 'int'
  StatusCode status = -1;
                      ^~
../httplib.h:4247:9: error: no viable overloaded '='
    res = new_res;
    ~~~ ^ ~~~~~~~
../httplib.h:7087:20: note: in instantiation of function template specialization 'httplib::detail::redirect<httplib::ClientImpl>' requested here
    return detail::redirect(*this, req, res, path, location, error);
                   ^
../httplib.h:622:13: note: candidate function not viable: expects an rvalue for 1st argument
  Response &operator=(Response &&) = default;
            ^
../httplib.h:4247:9: error: no viable overloaded '='
    res = new_res;
    ~~~ ^ ~~~~~~~
../httplib.h:7094:22: note: in instantiation of function template specialization 'httplib::detail::redirect<httplib::SSLClient>' requested here
      return detail::redirect(cli, req, res, path, location, error);
                     ^
../httplib.h:622:13: note: candidate function not viable: expects an rvalue for 1st argument
  Response &operator=(Response &&) = default;
            ^
3 errors generated.
In file included from include_httplib.cc:5:
../httplib.h:590:23: error: cannot initialize a member subobject of type 'StatusCode' with an rvalue of type 'int'
  StatusCode status = -1;
                      ^~
../httplib.h:4247:9: error: no viable overloaded '='
    res = new_res;
    ~~~ ^ ~~~~~~~
...
```

yhirose:
Even if I added `NotSet = -1`, we get lots of compile errors in 'test/test.cc'.

```cpp
  ...
  NetworkAuthenticationRequired = 511, ///< 511 Network Authentication Required

  NotSet = -1;
};

struct Response {
  std::string version;
  // int status = -1;
  StatusCode status = StatusCode::NotSet;
  std::string reason;
  ...
```

```
~/Projects/cpp-httplib/test$ make
clang++ -o test -I.. -g -std=c++11 -I. -Wall -Wextra -Wtype-limits -Wconversion -Wshadow  test.cc include_httplib.cc gtest/gtest-all.cc gtest/gtest_main.cc -DCPPHTTPLIB_OPENSSL_SUPPORT -I/usr/local/opt/openssl@1.1/include -L/usr/local/opt/openssl@1.1/lib -lssl -lcrypto -DCPPHTTPLIB_USE_CERTS_FROM_MACOSX_KEYCHAIN -framework CoreFoundation -framework Security -DCPPHTTPLIB_ZLIB_SUPPORT -lz -DCPPHTTPLIB_BROTLI_SUPPORT -I/usr/local/opt/brotli/include -L/usr/local/opt/brotli/lib -lbrotlicommon -lbrotlienc -lbrotlidec -pthread
In file included from test.cc:1:
../httplib.h:586:14: error: expected '}' or ','
  NotSet = -1;
             ^
../httplib.h:5315:22: error: assigning to 'StatusCode' from incompatible type 'int'
      this->status = stat;
                     ^~~~
../httplib.h:5317:22: error: assigning to 'StatusCode' from incompatible type 'int'
      this->status = 302;
                     ^~~
../httplib.h:6096:22: error: assigning to 'StatusCode' from incompatible type 'int'
        res.status = 413; // NOTE: should be 414?
                     ^~~
../httplib.h:6127:20: error: assigning to 'StatusCode' from incompatible type 'int'
      res.status = 400;
                   ^~~
../httplib.h:6156:8: error: no matching function for call to 'read_content'
  if (!detail::read_content(strm, req, payload_max_length_, res.status, nullptr,
       ^~~~~~~~~~~~~~~~~~~~
../httplib.h:3982:6: note: candidate function template not viable: no known conversion from 'StatusCode' to 'int &' for 4th argument
bool read_content(Stream &strm, T &x, size_t payload_max_length, int &status,
     ^
../httplib.h:6163:20: error: assigning to 'StatusCode' from incompatible type 'int'
      res.status = 400;
                   ^~~
../httplib.h:6405:16: error: assigning to 'StatusCode' from incompatible type 'int'
  res.status = 400;
               ^~~
../httplib.h:6488:22: error: assigning to 'StatusCode' from incompatible type 'int'
        res.status = 416;
                     ^~~
...
```

bugdea1er:
@yhirose I'm not suggesting replacing the `Response::status` type to `enum class StatusCode` __in this pull-request__

I'm just suggesting an enum with constants like `StatusCode::Forbidden`, which the user can then use with the `Response` struct unchanged

I don't get any errors here:
```cpp
struct Response {
  std::string version;
  int status = -1;
  std::string reason;
  ...

response.status = httplib::StatusCode::Forbidden
```

yhirose:
@bugdea1er thanks for the clarification. I know understand what you meant. :)

Still I am not very comfortable with the enum though. It's because the enum is lacking the category information. The first digit of each status code gives me very valuable information. As soon as I see '2', I can quickly recognize the message is a 'Success' code, and '4' makes me understand it's a request problem. '5' gives me information that something bad happening in the server, and '3' is a redirect code.

So if the enum looks like the following, it might be acceptable...

```cpp
enum StatusCode {
  ...
  NotFound_404 = 404,
  ...
};

int status = StatusCode::NotFound_404;
```

bugdea1er:
@yhirose I added number suffixes to enum constants

yhirose:
@bugdea1er also we don't need the comments like `///< 100 Continue`.

bugdea1er:
@yhirose ok, I removed them

yhirose:
@bugdea1er thanks for your quick modification!

bugdea1er:
@yhirose I can create a follow-up pr where I replace every number status code with an enum constant in cpp-httplib code

Should I?

yhirose:
@bugdea1er could you make two separate pull requests, the first one is for 'httplib.h' and the other one is for files in 'test' and 'example' directories? Thanks!

bugdea1er:
@yhirose Sure thing! The first one for httplib.h is already here: https://github.com/yhirose/cpp-httplib/pull/1742

The second one will appear in a few days

PR #1741, opened 2023-12-19
ThreadPool: optional limit for jobs queue

vmaffione:
For very busy servers, the internal jobs queue where accepted sockets are enqueued can grow without limit.
This is a problem for two reasons:
 - queueing too much work causes the server to respond with huge latency, resulting in repetead timeouts on the clients; it is definitely better to reject the connection early, so that the client receives the backpressure signal as soon as the queue is becoming too large
 - the jobs list can eventually cause an out of memory condition

### DIFF
--- a/README.md
+++ b/README.md
@@ -433,6 +433,17 @@ If you want to set the thread count at runtime, there is no convenient way... Bu
 svr.new_task_queue = [] { return new ThreadPool(12); };
 ```
 
+You can also provide an optional parameter to limit the maximum number
+of pending requests, i.e. requests `accept()`ed by the listener but
+still waiting to be serviced by worker threads.
+
+```cpp
+svr.new_task_queue = [] { return new ThreadPool(/*num_threads=*/12, /*max_queued_requests=*/18); };
+```
+
+Default limit is 0 (unlimited). Once the limit is reached, the listener
+will shutdown the client connection.
+
 ### Override the default thread pool with yours
 
 You can supply your own thread pool implementation according to your need.
@@ -444,8 +455,10 @@ public:
     pool_.start_with_thread_count(n);
   }
 
-  virtual void enqueue(std::function<void()> fn) override {
-    pool_.enqueue(fn);
+  virtual bool enqueue(std::function<void()> fn) override {
+    /* Return true if the task was actually enqueued, or false
+     * if the caller must drop the corresponding connection. */
+    return pool_.enqueue(fn);
   }
 
   virtual void shutdown() override {
--- a/test/test.cc
+++ b/test/test.cc
@@ -6506,18 +6506,103 @@ TEST(SocketStream, is_writable_INET) {
 #endif // #ifndef _WIN32
 
 TEST(TaskQueueTest, IncreaseAtomicInteger) {
-  static constexpr unsigned int number_of_task{1000000};
+  static constexpr unsigned int number_of_tasks{1000000};
   std::atomic_uint count{0};
   std::unique_ptr<TaskQueue> task_queue{
       new ThreadPool{CPPHTTPLIB_THREAD_POOL_COUNT}};
 
-  for (unsigned int i = 0; i < number_of_task; ++i) {
-    task_queue->enqueue(
+  for (unsigned int i = 0; i < number_of_tasks; ++i) {
+    auto queued = task_queue->enqueue(
         [&count] { count.fetch_add(1, std::memory_order_relaxed); });
+    EXPECT_TRUE(queued);
   }
 
   EXPECT_NO_THROW(task_queue->shutdown());
-  EXPECT_EQ(number_of_task, count.load());
+  EXPECT_EQ(number_of_tasks, count.load());
+}
+
+TEST(TaskQueueTest, IncreaseAtomicIntegerWithQueueLimit) {
+  static constexpr unsigned int number_of_tasks{1000000};
+  static constexpr unsigned int qlimit{2};
+  unsigned int queued_count{0};
+  std::atomic_uint count{0};
+  std::unique_ptr<TaskQueue> task_queue{
+      new ThreadPool{/*num_threads=*/1, qlimit}};
+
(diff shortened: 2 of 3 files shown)


yhirose:
@vmaffione thanks for the pull request. Could you update README to explain the return value of `enqueue`
 method and add a unit test in test/test.cc right after `TaskQueueTest.IncreaseAtomicInteger`?

https://github.com/yhirose/cpp-httplib?tab=readme-ov-file#override-the-default-thread-pool-with-yours

Since it's a breaking change, it will be merged in 0.15.0 after the pending minor release (0.14.x) are published.

vmaffione:
Done.

test/test.cc
```diff
@@ -6512,14 +6512,33 @@ TEST(TaskQueueTest, IncreaseAtomicInteger) {
...
+  for (unsigned int i = 0; i < number_of_task; ++i) {
+    if (task_queue->enqueue(
+        [&count] { count.fetch_add(1, std::memory_order_relaxed); })) {
+      queued_count++;
+    }
+  }
+
+  EXPECT_NO_THROW(task_queue->shutdown());
+  EXPECT_EQ(queued_count, count.load());
+}
```
yhirose, line 6540: Could you insert `EXPECT_TRUE(queued_count < number_of_task);` at line 6540?

README.md
```diff
@@ -433,6 +433,17 @@ If you want to set the thread count at runtime, there is no convenient way... Bu
 svr.new_task_queue = [] { return new ThreadPool(12); };
 ```
 
+You can also provide an optional parameter to limit the maximum number
+of pending requests, i.e. requests `accept()`ed by the listener but
+still waiting to be serviced by worker threads.
+
+```cpp
+svr.new_task_queue = [] { return new ThreadPool(/*num_threads=*/12, /*max_pending=*/18); };
```
yhirose, line 441: Could you replace `max_pending` to `max_queued_requests` for consistency?

test/test.cc
```diff
@@ -6512,14 +6512,33 @@ TEST(TaskQueueTest, IncreaseAtomicInteger) {
       new ThreadPool{CPPHTTPLIB_THREAD_POOL_COUNT}};
 
   for (unsigned int i = 0; i < number_of_task; ++i) {
-    task_queue->enqueue(
+    bool queued = task_queue->enqueue(
```
yhirose, line 6515: Could you use `auto` instead of `bool`?

vmaffione: Ok, but doesn't that make the code less readable?

yhirose: It's consistency. I try to use `auto` at all the places where we can apply it.

test/test.cc
yhirose: Thanks for the adjustment. But I suggested `<` instead of `<=`, so that we can confirm that `task_queue->enqueue` fails at least once.

vmaffione: Although extremely unlikely, it is still possible that `enqueue` never fails.
The worker thread loop may be faster than the enqueue loop, in such a way that the latter will never find the queue full.

You can reproduce this condition by adding something like `std::this_thread::sleep_for(std::chrono::microseconds(100));` in the enqueue loop.

vmaffione: What we can add, instead, is `EXPECT_TRUE(queued_count >= 2)`, because even in the worst case (worker thread starts after the enqueue loop) two enqueue operations will always succeed.

yhirose: To be honest, I would like to see something simpler which clearly shows that this edge case is properly handled. The original IncreaseAtomicInteger test isn't intended for this purpose, and I don't think it's a good idea to reuse it as base...

vmaffione: Do you have any additional ideas?

To me this looks a very good test, because it will, in practice, heavily exercise both the cases (enqueue successful or not): on my machine I see about 15% of successful enqueue and 85% of failures. And it checks that the counter increments only when the enqueue is successful, thus validating the return value of enqueue.

Since the workers start on ThreadPool constructor (rather than with a separate `start()` method), in order to force an enqueue to fail I can only think of even more complex tests...

yhirose: I came up with the following simple edge case test.

```cpp
TEST(TaskQueueTest, MaxQueuedRequests) {
  std::unique_ptr<TaskQueue> task_queue{new ThreadPool{1, 2}};

  auto count = 8;
  while (count-- > 0) {
    task_queue->enqueue(
        [] { std::this_thread::sleep_for(std::chrono::milliseconds(50)); });
  }

  auto ret = task_queue->enqueue([] {});
  EXPECT_FALSE(ret);

  std::this_thread::sleep_for(std::chrono::milliseconds(100));

  ret = task_queue->enqueue([] {});
  EXPECT_TRUE(ret);

  EXPECT_NO_THROW(task_queue->shutdown());
}
```

I tested it 100 times with the following command, and confirmed that it worked successfully. Of course, there is a slight change to see it fails, but I think it's really rare and ok to put it in the unit test.

```
~/Projects/cpp-httplib/test$ make test && for i in {0..99}; do ./test --gtest_filter="*MaxQueued*"; done
```

Could you add the above the unit test case right below your `IncreaseAtomicIntegerWithQueueLimit`?

Also please run the clang-format on 'httplib.h' and 'test/test.cc'. Thanks for your help!

vmaffione: clang-format done.

What about the same test idea you proposed, but with proper synchronization?

yhirose: Looks much nicer!

yhirose: Could take a look at the failure in 'build (macOS-latest)' on Github Actions?
https://github.com/yhirose/cpp-httplib/actions/runs/7309110277/job/19916387047?pr=1741

vmaffione: There was a logic bug in the test. Now is fixed.

vmaffione: Now I've run the test 10000 times and it always succeded.

yhirose:
@vmaffione thanks for your fine contribution!

PR #1742, opened 2023-12-20
Use StatusCode in httplib code

bugdea1er:
@yhirose Replaced number status codes with enum constants. Also had to move `enum StatusCode` up.

Question: I've come across number in comparisons like this:
```cpp
if (300 < res.status && res.status < 400) ...
```

Should I replace these numbers with constants here too?

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -353,6 +353,81 @@ private:
 
 } // namespace detail
 
+enum StatusCode {
+  // Information responses
+  Continue_100 = 100,
+  SwitchingProtocol_101 = 101,
+  Processing_102 = 102,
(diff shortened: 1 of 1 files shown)


yhirose:
@bugdea1er good questions. 🤔

In such a situation, it's confusing to me if 300 is replaced with MultipleChoices_300. Could you leave them as they are for now?

bugdea1er:
@yhirose I thought you wouldn't like it, so I didn't replace them

The pr is ready

yhirose:
@bugdea1er thanks again for your fine contribution!

PR #1744, opened 2023-12-22
add missing #include for strcasecmp

divinity76:
Posix specifications say you need strings.h to use strcasecmp

- it happens to work on most Linux systems without it, but that's basically just a compiler include bug (i think?)
- Cygwin gcc actually requires strings.h, compiling on Cygwin is how I noticed strings.h was missing


yhirose:
@divinity76 thanks for the report. I resolved this issue with removing the 'strcasecmp' dependency.

PR #1750, opened 2023-12-30
Fixes Windows std::max macro problem

TheOnlyJoey:
Prevents the need for NOMINMAX macro on Windows compilation.


yhirose:
@TheOnlyJoey thank you for your contribution!

PR #1754, opened 2024-01-04
Minor cmake fix & cleanup

sum01:
Most of this is just formatting that changes nothing but comments and rearranges some things.

The only important fix is b5941d40b04d5f57374525ffaa57ec2b1f2a6a9e as it prevents it from accidentally using a dependency when the user didn't set that option to ON but did use it in their own project.


yhirose:
@sum01, Thanks always!

PR #1755, opened 2024-01-11
Add a getter for a bearer token from a request

bugdea1er:
Hi there!

This pr adds a simple getter for bearer token from a request

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -561,6 +561,8 @@ struct Request {
   size_t get_header_value_count(const std::string &key) const;
   void set_header(const std::string &key, const std::string &val);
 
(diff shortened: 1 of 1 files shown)


yhirose:
@bugdea1er thanks for the pull request. However, I would like to keep `Request` simple and just include things absolutely necessary.

I am not against adding helper functions for `Request` as below.

```cpp
std::string get_bearer_token_auth(const Request &req);
```

bugdea1er:
@yhirose got it, replaced the method with a helper function

yhirose:
@bugdea1er thanks for the update!

PR #1757, opened 2024-01-12
Fix select() return code for fd >= 1024

jfsimoneau:
I noticed all callers look for errors in negative numbers, but these code paths returned 1 instead of -1.


yhirose:
@jfsimoneau thanks for catching the bug!

PR #1764, opened 2024-01-27
Support move semantics for `Response::set_content()`

jimmy-park:
When we respond with a pre-allocated/temporary `std::string` variable, this method avoids an unnecessary memory copy.
It helps us to transfer large data without using content providers.


yhirose:
Thanks!

PR #1769, opened 2024-02-02
Prevent test txt & html files being converted to CRLF line ending.

chunhualiu:
Test will fail if the test files were converted to CRLF line ending on Windows.


yhirose:
@chunhualiu I am not super interested in this pull request, since I don't have any problem without this file on both my Windows and GitHub Action workflow. Anyway, thanks for your effort!

PR #1770, opened 2024-02-04
Fix typo in gtest-all.cc

eltociear:
synthetic -> synthetic


yhirose:
@eltociear thanks for the report. It is actually not my code, but Google's. So could you also report it to the Google Test project? Thanks!

PR #1775, opened 2024-02-06
Implement file mapping for UWP apps

### LINKED ISSUE #1773: Breaking change on WinRT/UWP
This change https://github.com/yhirose/cpp-httplib/pull/1632 breaks WinRT/UWP builds - there is no [`CreateFileA()`](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilea#requirements) available

sergio-nsk:
Fixes #1773.

@yhirose I have not checked it - perhaps, the new code also can replace Windows Desktop code, but it needs a real UTF-8 to UTF-16 conversion.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -2686,21 +2686,29 @@ inline bool mmap::open(const char *path) {
   close();
 
 #if defined(_WIN32)
-  hFile_ = ::CreateFileA(path, GENERIC_READ, FILE_SHARE_READ, NULL,
-                         OPEN_EXISTING, FILE_ATTRIBUTE_NORMAL, NULL);
+  std::wstring wpath;
+  for (size_t i = 0; i < strlen(path); i++) {
+    wpath += path[i];
+  }
+
+  hFile_ = ::CreateFile2(wpath.c_str(), GENERIC_READ, FILE_SHARE_READ,
+                         OPEN_EXISTING, NULL);
 
   if (hFile_ == INVALID_HANDLE_VALUE) { return false; }
 
-  size_ = ::GetFileSize(hFile_, NULL);
+  LARGE_INTEGER size{};
+  if (!::GetFileSizeEx(hFile_, &size)) { return false; }
+  size_ = static_cast<size_t>(size.QuadPart);
 
-  hMapping_ = ::CreateFileMapping(hFile_, NULL, PAGE_READONLY, 0, 0, NULL);
+  hMapping_ = ::CreateFileMappingFromApp(hFile_, NULL, PAGE_READONLY, size_,
+                                         NULL);
 
   if (hMapping_ == NULL) {
     close();
     return false;
   }
 
-  addr_ = ::MapViewOfFile(hMapping_, FILE_MAP_READ, 0, 0, 0);
+  addr_ = ::MapViewOfFileFromApp(hMapping_, FILE_MAP_READ, 0, 0);
 #else
   fd_ = ::open(path, O_RDONLY);
   if (fd_ == -1) { return false; }


yhirose:
@sergio-nsk What does "I have not checked it" mean? At least, I would like you to confirm this change works on our UWP apps, and also confirm that it doesn't break anything on Windows.

sergio-nsk:
I mean, I have not checked if the new code can work on Windows Desktop and the inner #if #else #endif can be eliminated. It works in UWP.

yhirose:
@sergio-nsk thanks for the reply.

According to your comment in #1773, the only thing we have to do is to change `CreateFileA` to something else.

If `CreateFile2` works on WinRT/UWP, I think the following small change is enough for both the regular Windows and WinRT/UWP. I confirmed that all the unit tests (test/test.cc) related to the code change passed successfully.

```cpp
diff --git a/httplib.h b/httplib.h
index dfdd260..8ef4d01 100644
--- a/httplib.h
+++ b/httplib.h
@@ -2686,8 +2686,13 @@ inline bool mmap::open(const char *path) {
   close();

 #if defined(_WIN32)
-  hFile_ = ::CreateFileA(path, GENERIC_READ, FILE_SHARE_READ, NULL,
-                         OPEN_EXISTING, FILE_ATTRIBUTE_NORMAL, NULL);
+  std::wstring wpath;
+  for (size_t i = 0; i < strlen(path); i++) {
+     wpath += path[i];
+  }
+
+  hFile_ = ::CreateFile2(wpath.c_str(), GENERIC_READ, FILE_SHARE_READ,
+                         OPEN_EXISTING, NULL);

   if (hFile_ == INVALID_HANDLE_VALUE) { return false; }
```

Could you test the above code on your UWP environment? Thanks a lot!

sergio-nsk:
Honestly, I talked only about the first compiler error. Other functions [`GetFileSize()`](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getfilesize#requirements), [`CreateFileMapping()`](https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-createfilemappinga#requirements), [`MapViewOfFile()`](https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-mapviewoffile#requirements) are also unavailable in UWP apps. [`UnmapViewOfFile()`](https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-unmapviewoffile#requirements) is only available. That's why I suggest that you test the complete new code block.
```diff
- #if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP)
+ #if 0 && WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP)
```

sergio-nsk:
Sorry, I just noticed that it will not compile on Windows 7. Are you sure you want to break that although you don't support it?
> NOTE: Windows 8 or lower, Visual Studio 2013 or lower, and Cygwin and MSYS2 including MinGW are neither supported nor tested.

yhirose:
I think that's fine, since the support for Windows 7 has already ended on January 14, 2020.

sergio-nsk:
Done, with support VS 2015.

CMakeLists.txt
```diff
@@ -227,6 +227,8 @@ target_link_libraries(${PROJECT_NAME} ${_INTERFACE_OR_PUBLIC}
 
 # Set the definitions to enable optional features
 target_compile_definitions(${PROJECT_NAME} ${_INTERFACE_OR_PUBLIC}
+	# Require Windows 8
+	$<$<PLATFORM_ID:Windows>:_WIN32_WINNT=0x0602>
```
yhirose, line 231: @sum01 @jimmy-park could you review the change in CMakeLists.txt, since I am not an expert of CMake? Thanks a lot.

jimmy-park: Won't this affect other CMake projects that use httplib?
I think the macro should remain in `httplib.h`.

yhirose: @jimmy-park thank you for the comment.
@sergio-nsk could you remove this change?

sergio-nsk: > Won't this affect other CMake projects that use httplib? I think the macro should remain in `httplib.h`.

It will affect user projects if `HTTPLIB_COMPILE` is `ON`.

yhirose:
@sergio-nsk thanks for your contribution!

rakakarakaka:
@sergio-nsk this is fucking shit!!! you broke support Windows 7, Vista, and even XP with this pull request!!!

May you be cursed for the rest of your life for this!

PR #1776, opened 2024-02-07
Reduce object copy further

jimmy-park:
I found more object copies that are not detected by static analyzers after submitting #1767

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -1956,7 +1956,7 @@ inline std::string get_bearer_token_auth(const Request &req) {
     return req.get_header_value("Authorization")
         .substr(BearerHeaderPrefix.length());
   }
-  return "";
+  return {};
 }
 
 template <class Rep, class Period>
@@ -4272,8 +4272,8 @@ inline bool redirect(T &cli, Request &req, Response &res,
 
   auto ret = cli.send(new_req, new_res, error);
   if (ret) {
-    req = new_req;
-    res = new_res;
+    req = std::move(new_req);
+    res = std::move(new_res);
 
     if (res.location.empty()) { res.location = location; }
   }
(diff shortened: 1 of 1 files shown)


yhirose:
@jimmy-park thanks for the further modification. But I have actually been thinking of removing those changes using `std::move` unless they are absolutely necessary, since it degrades code readability in my opinion.

Also I found there are some cases that I don't think we should use `std::move`.

For instance, I found the following code recently, and it looks dangerous because `pop_front()` now accesses the moved object.

https://github.com/yhirose/cpp-httplib/blob/5c00bbf36ba8ff47b4fb97712fc38cb2884e5b98/httplib.h#L722-L723

Another example is the following.

https://github.com/yhirose/cpp-httplib/blob/5c00bbf36ba8ff47b4fb97712fc38cb2884e5b98/httplib.h#L3802
https://github.com/yhirose/cpp-httplib/blob/5c00bbf36ba8ff47b4fb97712fc38cb2884e5b98/httplib.h#L3839-L3842
https://github.com/yhirose/cpp-httplib/blob/5c00bbf36ba8ff47b4fb97712fc38cb2884e5b98/httplib.h#L3940-L3943

`parse_header` now passes moved objects to a functor, and a user who supplies a functor needs to know parameters need to be received as `&&`. I would like to avoid such a hidden dependency that makes the code harder to understand.

Using `std::move` may give us performance benefit though, I think it's pretty small performance gain or even unnoticeable. I am not more concerned with the code readability as the httplib.h grows, and would like to keep the code simple. Also any performance improvements should be measured to see if they are really necessary.

I'll keep this pull request for now, but I may close it later because of the above reason.
Thanks for your understanding in this matter.

PR #1779, opened 2024-02-17
Use `final` keyword for devirtualization

jimmy-park:
Adding the `final` specifier may help compilers to optimize code by inlining. ([This article](https://devblogs.microsoft.com/cppblog/the-performance-benefits-of-final-classes/) explains devirtualization very well)
`SSLServer` could also be final, but the test case `SSLClientServerTest.SSLConnectTimeout` should be modified.

This change will prevent users from inheriting internal classes.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -462,7 +462,7 @@ public:
   std::ostream os;
 
 private:
-  class data_sink_streambuf : public std::streambuf {
+  class data_sink_streambuf final : public std::streambuf {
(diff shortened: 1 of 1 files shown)


yhirose:
Thanks for the improvement!

Agritite:
What? Since when httplib::SSLClient is an internal class (it's literally used in the examples)? What if I want to inherit from it?

PR #1781, opened 2024-02-23

ArnaudBienner:
When doing some profiling on a client application doing a lot of downloads, I realized my application spent ~13% of its time doing `std::string::append` (compiled with gcc and -03, even more when compiling in debug mode with -O1: ~30%).

I realized this could be improved by reserving the output buffer (`res.body`) since we know the output size if `Content-Length` header is present.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -6065,6 +6065,21 @@ inline bool Server::write_response_core(Stream &strm, bool close_connection,
     detail::write_data(strm, data.data(), data.size());
   }
 
+  // If we know the content length, reserve the expected size in the output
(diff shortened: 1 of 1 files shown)


ArnaudBienner:
I've implemented an alternative in https://github.com/yhirose/cpp-httplib/pull/1782.

yhirose:
@ArnaudBienner I don't think `std::string::append` gets called after your modification, because the rest of code after the line  6069 in `Server::write_response_core` is only for writing `res.body` or data from a content provider. Could you confirm that your code really works? Thanks!

PR #1782, opened 2024-02-23

ArnaudBienner:
This is an alternative approach to https://github.com/yhirose/cpp-httplib/pull/1781.

It is simpler: only one line, the checks on length have already been made.

But I wasn't sure if it was OK or not, since this implies `x` has a `body` member, so it makes this template function less generic, which might defeat the purpose of having a template here...

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -4030,6 +4030,7 @@ bool read_content(Stream &strm, T &x, size_t payload_max_length, int &status,
             skip_content_with_length(strm, len);
             ret = false;
           } else if (len > 0) {
+            x.body.reserve(len);
             ret = read_content_with_length(strm, len, std::move(progress), out);
           }
         }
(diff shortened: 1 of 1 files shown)


httplib.h
```diff
@@ -4030,6 +4030,7 @@ bool read_content(Stream &strm, T &x, size_t payload_max_length, int &status,
             skip_content_with_length(strm, len);
             ret = false;
           } else if (len > 0) {
+            x.body.reserve(len);
```
ArnaudBienner, line 4033: CI is failing probably because there is no bound checking here.
I believe the following would be better:
```suggestion
            if (len > 0 && len < payload_max_length) {
              x.body.reserve(len);
            }
```

ArnaudBienner:
@yhirose I'll have a look. I suspect this is due to the max limit being too high resulting in memory allocation failures.

Apart from that, do you think this is the good way to improve the performance I had, or is #1781 is better? If you think the other PR is preferable, I can simply close this PR without further investigation.

yhirose:
Yes, you can just close it.

PR #1785, opened 2024-02-27
Add description & URL to Cmake

sum01:
Doesn't do much, but some packages/builders might find a use for these vars it provides.


yhirose:
Thanks!

PR #1788, opened 2024-02-29

ArnaudBienner:
The append I'm trying to avoid is the one at line 7507:
`res.body.append(buf, n);`

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -7461,6 +7461,21 @@ inline bool ClientImpl::process_request(Stream &strm, Request &req,
     error = Error::Read;
     return false;
   }
+  
+  // If we know the content length, reserve the expected size in the output
+  // buffer to avoid re-allocations and copies when appending received data
+  auto it = res.headers.find("Content-Length");
+  if (it != res.headers.end()) {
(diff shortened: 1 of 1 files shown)


yhirose:
@ArnaudBienner same as https://github.com/yhirose/cpp-httplib/pull/1781#issuecomment-1962254305, I don't think it improve performance since `std::string::append` won't be called after your code.

ArnaudBienner:
> @ArnaudBienner same as [#1781 (comment)](https://github.com/yhirose/cpp-httplib/pull/1781#issuecomment-1962254305), I don't think it improve performance since `std::string::append` won't be called after your code.

Isn't it called as part of `out` which is passed to `detail::read_content` along with `res` and used to read the data?

yhirose:
@ArnaudBienner sorry for the late reply. You are right. `res.body.append` gets called after your code. But the code shouldn't be called when a content receiver is provided, because `res.body` should be empty in this case.

```cpp
    auto res = cli.Get("/hi", [&](const char *data, size_t data_length) {
      ...
      return true;
    });
```

PR #1792, opened 2024-03-07
fix ambiguous for HandlerWithResponse and Handler in set_error_handler

HerrCai0907:
using lambda expression as Handler in set_error_handler will cause ambiguous.
Template forwarding can forward HandlerWithResponse to the correct overloading function

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -231,6 +231,7 @@ using socket_t = int;
 #include <string>
 #include <sys/stat.h>
 #include <thread>
+#include <type_traits>
 #include <unordered_map>
 #include <unordered_set>
 #include <utility>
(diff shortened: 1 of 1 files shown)


HerrCai0907:
> @HerrCai0907 what actual ambiguous situation is the current cpp-httplib causing? Compile warning?

It is a compiler error. here is a small example https://godbolt.org/z/P8G45vc54

yhirose:
Thanks. Could you also show me your code snippet with `set_error_handler` which caused the compile error?

HerrCai0907:
```c++
inline Server &Server::set_error_handler(HandlerWithResponse handler) {
  error_handler_ = std::move(handler);
  return *this;
}

inline Server &Server::set_error_handler(Handler handler) {
  error_handler_ = [handler](const Request &req, Response &res) {
    handler(req, res);
    return HandlerResponse::Handled;
  };
  return *this;
}
```

yhirose:
@HerrCai0907 sorry, I was asking for 'your' calling example of the set_error_handler.

HerrCai0907:
source code:
```c++
#include "httplib.h"
int main() {
  httplib::Server server{};
  server.set_error_handler([](const httplib::Request &, httplib::Response &) {
    return httplib::Server::HandlerResponse::Unhandled;
  });
}
```

compiler error:
```
a.cpp:5:10: error: call to member function 'set_error_handler' is ambiguous
    5 |   server.set_error_handler([](const httplib::Request &, httplib::Response &) {
      |   ~~~~~~~^~~~~~~~~~~~~~~~~
./httplib.h:5802:24: note: candidate function
 5802 | inline Server &Server::set_error_handler(HandlerWithResponse handler) {
      |                        ^
./httplib.h:5807:24: note: candidate function
 5807 | inline Server &Server::set_error_handler(Handler handler) {
      |                        ^
1 error generated.
```

clang++ -v
```
Homebrew clang version 17.0.6
Target: arm64-apple-darwin23.2.0
Thread model: posix
InstalledDir: /opt/homebrew/opt/llvm/bin
```

yhirose:
@HerrCai0907 thanks for your contribution!

PR #1804, opened 2024-03-20
Added file upload along the link path

sirotenkodev:
Added the ability to create files of the type about.html so that the folder is not read when downloading index.html and the file name corresponded to the path name

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -6251,7 +6251,15 @@ inline bool Server::handle_file_request(const Request &req, Response &res,
       std::string sub_path = "/" + req.path.substr(entry.mount_point.size());
       if (detail::is_valid_path(sub_path)) {
         auto path = entry.base_dir + sub_path;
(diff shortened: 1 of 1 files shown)


sirotenkodev:
In the example, you can load a page from a file index.html , and by the name of the folder path, in the example about
https://github.com/sirotenkodev/ServerApp/tree/main/public

yhirose:
For now, I don't merge this pull request unless it's a common functionality among popular HTTP libraries.

sirotenkodev:
@yhirose 
For example, the Drogon library allows you to create your own views and load the necessary files by name https://github.com/drogonframework/drogon/blob/master/examples/login_session/main.cc
Also, C# asp.net allows you to load all the necessary files by name by view  https://learn.microsoft.com/en-us/aspnet/core/mvc/views/overview?view=aspnetcore-8.0

PR #1806, opened 2024-03-21
Apply range header base on response status code

xyhuang7215:
I've recently been using cpp-httplib to handle range requests for streaming video. 
However, when I try to use JSON as the response to provide additional error information,  response body get truncated and content-type might be override by 'multipart/byteranges' due to the range header of request. 
This PR employs a naive approach to address this issue. 
If there's a better implementation or other methods to achieve my goal, please let me know. 
Thank you.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -590,7 +590,6 @@ struct Response {
   Headers headers;
   std::string body;
   std::string location; // Redirect location
-  bool ignore_range = false;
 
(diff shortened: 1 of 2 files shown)


xyhuang7215:
Update: A better approach should be leveraging response status code.

yhirose:
@xyhuang7215 thanks for your contribution!

PR #1811, opened 2024-04-05
Generate missing PEMs for CTest

jimmy-park:
New PEM files are introduced in 548dfff, so CMake needs to follow this.


yhirose:
@jimmy-park Thanks!

PR #1812, opened 2024-04-06
Fix range parser when parsing too many ranges

jimmy-park:
Stack overflow is detected by AddressSanitizer on MSVC when parsing too many ranges using `std::regex`.
`ServerTest.GetStreamedWithTooManyRanges` has failed due to this reason.
I would like to suggest a range parser with plain string manipulations rather than relying on `std::regex`.
And some test cases are added for invalid ranges not covered by the current implementation.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -4365,35 +4365,44 @@ inline bool parse_range_header(const std::string &s, Ranges &ranges) {
 #else
 inline bool parse_range_header(const std::string &s, Ranges &ranges) try {
 #endif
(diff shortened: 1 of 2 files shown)


yhirose:
@jimmy-park thanks for the fix. Could you check the performance difference between the regex implementation and the new implementation? I believe the new one should outperform the regex one, but I would like to confirm it before merge the pull request. Thanks!

jimmy-park:
Here is a simple benchmark result.
https://quick-bench.com/q/baC_7m3nbxFySBpi5kTcGYfr2j0

yhirose:
@jimmy-park great job!

PR #1814, opened 2024-04-07
fix: increase default receive buffer to 16kb

rustyconover:
Since TLS packets have a maximum size of 16kb it makes sense to fully accommodate them on reads.


yhirose:
@rustyconover thanks for the contribution!

PR #1820, opened 2024-04-18
Highlight notes using GitHub Markdown feature

jimmy-park:
GitHub has a Markdown feature that is useful for highlighting information to users.
There are five types of highlights and I changed `README.md` a little bit based on following contexts.

> [!NOTE]  
> Highlights information that users should take into account, even when skimming.

> [!TIP]
> Optional information to help a user be more successful.

> [!IMPORTANT]  
> Crucial information necessary for users to succeed.

> [!WARNING]  
> Critical content demanding immediate user attention due to potential risks.

> [!CAUTION]
> Negative potential consequences of an action.

Here is a preview!
https://github.com/jimmy-park/cpp-httplib/blob/highlight-notes/README.md

More information about this feature
https://github.com/orgs/community/discussions/16925

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,8 @@ A C++11 single-file header-only cross platform HTTP/HTTPS library.
 
 It's extremely easy to setup. Just include the **httplib.h** file in your code!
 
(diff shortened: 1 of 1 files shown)


yhirose:
@jimmy-park thanks! It now looks fancier.

PR #1821, opened 2024-04-18
Added progress to POST, PUT, PATCH and DELETE requests

### LINKED ISSUE #1330: httplib::Progress for Post requests?
Hello,
have you thought about adding a httplib::Progress for the "Post" requests ?

ty

rndm13:
Added multiple methods to Client and ClientImpl classes that accept progress callback for POST, PUT, PATCH and DELETE.
Needed to have a functionality for user to cancel their requests and these changes add that.
Main change is in send_with_content_provider which now also accepts Progress. Other changes are mostly boilerplate functions additions.
Fixes #1330 issue as well

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -8205,7 +8205,7 @@ inline Result ClientImpl::Delete(const std::string &path,
                                  const Headers &headers, const char *body,
                                  size_t content_length,
                                  const std::string &content_type) {
-  return Delete(path, headers, body, content_length, content_type);
+  return Delete(path, headers, body, content_length, content_type, nullptr);
 }
 
 inline Result ClientImpl::Delete(const std::string &path,
--- a/httplib.h
+++ b/httplib.h
@@ -1148,10 +1148,18 @@ public:
               const std::string &content_type);
   Result Post(const std::string &path, const Headers &headers, const char *body,
               size_t content_length, const std::string &content_type);
+  Result Post(const std::string &path, const Headers &headers, const char *body,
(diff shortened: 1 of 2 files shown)


yhirose:
@rndm13, thanks for the pull request. Could you please add unit tests for the new methods?

yhirose:
@rndm13, thanks for the contribution!

kunlinyu:
@rndm13 HI, I am using this feature for "PUT"
```cpp
httplib::Client cli(domain);
auto res = cli.Put(
    path_with_params, {}, buffer.data(), bytes_read,
    "application/octet-stream", [&](uint64_t current, uint64_t total) {
      std::cout << "current/total: " << current << "/" << total << std::endl;
      return true;
    });
```

But even if I put a big file more then 10MB, the progress will always be "current/total: 16/16".
Then I check the code. I notice that the "read_content_with_length()" will call the progress lambda.

So I want to ask, is it about the progress of the file I put to the server, or the progress about the response?

rndm13:
@kunlinyu Hello, from what I understand the progress does indeed give you information about server response, not about your client request.
Maybe it would be best to rename Progress to ResponseProgress to avoid this kind of confusion?

kunlinyu:
@rndm13 Thank you! I think you are right. 
Accually the first time when I saw the name "progress", I thought it is response progress for GET and request progress for PUT and POST. It is very good to rename it ResponseProgress. And furthur more, I think we need both RequestProgress and ResponseProgress.

PR #1822, opened 2024-04-19
Fix query parsing when value has `=` characters

jimmy-park:
The current query parser handles `"?a=b=c&..."` as `("a", "c")` which should be `("a", "b=c")`.
We should not use `split()` when parsing query strings.
So I need to implement a string divider which just splits in half.
Here are possible outcomes of the string divider.

```cpp
divide("", '=')      -> ("", "")
divide("=", '=')     -> ("", "")
divide(" ", '=')     -> (" ", "")
divide("a", '=')     -> ("a", "")
divide("a=", '=')    -> ("a", "")
divide("=b", '=')    -> ("", "b")
divide("a=b", '=')   -> ("a", "b")
divide("a=b=", '=')  -> ("a", "b=")
divide("a=b=c", '=') -> ("a", "b=c")
```

> [!NOTE]
> Whitespaces aren't trimmed!

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -2604,11 +2604,11 @@ divide(const char *data, std::size_t size, char d,
        std::function<void(const char *, std::size_t, const char *, std::size_t)>
            fn) {
   const auto it = std::find(data, data + size, d);
-  const auto found = static_cast<int>(it != data + size);
+  const auto found = static_cast<std::size_t>(it != data + size);
   const auto lhs_data = data;
   const auto lhs_size = static_cast<std::size_t>(it - data);
   const auto rhs_data = it + found;
-  const auto rhs_size = static_cast<std::size_t>(size - lhs_size - found);
+  const auto rhs_size = size - lhs_size - found;
 
   fn(lhs_data, lhs_size, rhs_data, rhs_size);
 }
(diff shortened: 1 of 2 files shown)


yhirose:
@jimmy-park thanks for the pull request and the code looks good to me. Where do you find the information of `"?a=b=c&..."` in the HTML spec or RFCs?

jimmy-park:
@yhirose RFC does not specify the format of query. Any string between `'?'` and `'#'` could be a query . I think `key=value` pair is just a well known convention that people use.
https://datatracker.ietf.org/doc/html/rfc3986#section-3.4

And I've seen some queries with `'='` in the value actually being used at work.

yhirose:
@jimmy-park, ok. Could you please check how other HTTP libraries handle these queries? If they do the same, I have no issue with this pull request.

jimmy-park:
@yhirose I found one.
https://github.com/drogonframework/drogon/blob/96919df488e0ebaa0ed304bbd76bba33508df3cc/lib/src/HttpRequestImpl.cc#L145-L163

yhirose:
@jimmy-park thanks for the improvement!

PR #1826, opened 2024-04-21
Move httplibConf.cmake.in & install readme/license

sum01:
Moving the file to the cmake dir just seems to be what I should've done before since it has nothing to do with anything else.

Installing the readme & license seems like it makes sense to me. I think MIT licensing says you're supposed to anyways.


yhirose:
@sum01 thanks for the contribution!

PR #1827, opened 2024-04-25
New function SSLServer::update_certs. Allows to update certificates while server is running

RainerSchielke:
We want to update the ssl certificates without stopping the httplib server. The new function does this in a safe way, because it uses the ctx_mutex_

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -1819,6 +1819,9 @@ public:
   bool is_valid() const override;
 
   SSL_CTX *ssl_context() const;
+  
+  void update_certs (X509 *cert, EVP_PKEY *private_key,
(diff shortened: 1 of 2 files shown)


yhirose:
@RainerSchielke thank you for the pull request. Could you please add a unit test for this change?

RainerSchielke:
> @RainerSchielke thank you for the pull request. Could you please add a unit test for this change?

Ok I willt try it, but I cannot find the certificates used in the test eg. "./cert.pem"

yhirose:
@RainerSchielke sorry for the late reply. Thanks for your contribution!

PR #1828, opened 2024-05-01
Thread safe cloexec

kdombroski:
Thread safe socket creation, using:

- SOCK_CLOEXEC on Linux (and others where available)
- accept4 on Linux
- WSAAccept on Windows

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -6484,7 +6484,15 @@ inline bool Server::listen_internal() {
 #ifndef _WIN32
       }
 #endif
+
+#if defined _WIN32
+      // sockets conneced via WASAccept inherit flags NO_HANDLE_INHERIT, OVERLAPPED
+      socket_t sock = WSAAccept(svr_sock_, nullptr, nullptr, nullptr, 0);
+#elif defined __linux__
+      socket_t sock = accept4(svr_sock_, nullptr, nullptr, SOCK_CLOEXEC);
+#else
       socket_t sock = accept(svr_sock_, nullptr, nullptr);
+#endif
 
       if (sock == INVALID_SOCKET) {
         if (errno == EMFILE) {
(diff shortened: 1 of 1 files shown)


yhirose:
@kdombroski thanks for your pull request, but I am not sure if it is really needed. Could you explain why you need the changes and benefits? Thanks.

kdombroski:
There is a brief window of time between creating a socket and calling fcntl with FD_CLOEXEC where another thread could create a new process inheriting the file descriptor.
Adding the SOCK_CLOEXEC flag eliminates this race condition.

yhirose:
How about the followings?

* accept4 on Linux
* WSAAccept on Windows

kdombroski:
accept4 on Linux and WSAAccept on Windows is great.
But - for true fork() safety sockets need to be constructed with the SOCK_CLOEXEC flag.

From the open manpage:
"Note  that  the use of this flag is essential in some multithreaded programs, because using a separate fcntl(2) F_SETFD operation to set the FD_CLOEXEC flag does not suffice to avoid race conditions where one thread opens a file descriptor and attempts to set its close-on-exec flag using fcntl(2) at the same time as another thread does a fork(2) plus execve(2)."

SOCK_CLOEXEC is supported on Linux and BSD.
The main platform without it is Apple - I have not found an alternative there.

httplib.h
```diff
@@ -6470,7 +6484,15 @@ inline bool Server::listen_internal() {
 #ifndef _WIN32
       }
 #endif
+
+#if defined _WIN32
+      // sockets conneced via WASAccept inherit flags NO_HANDLE_INHERIT, OVERLAPPED
+      socket_t sock = WSAAccept(svr_sock_, nullptr, nullptr, nullptr, 0);
+#elif defined __linux__
```
yhirose, line 6491: I am not comfortable with using `__linux__`. If it possible to use `SOCK_CLOEXEC` instead?

PR #1841, opened 2024-05-20
Constantify lambda variables

GermanAizek:
Fixes from here: https://github.com/ggerganov/llama.cpp/pull/5511


GermanAizek:
Im sorry, we do not have an up-to-date version httplib. Im closed PR.

PR #1843, opened 2024-05-22
Tweak CI & fix macOS prefix

jajik:
Hello,

I separated each os in the workflow file to make it cleaner. Additionally, now, in case of one OS workflow failure, the others don't get cancelled (although this could be achieved with `fail-fast: false` in the current setting). That seems more reasonable, because if there are issues, you want to see them all at once.

Another part is the fix for macOS and its homebrew prefix.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -3,39 +3,41 @@ name: test
 on: [push, pull_request]
 
 jobs:
-  build:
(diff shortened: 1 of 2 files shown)


yhirose:
@jajik thanks for the contribution!

jajik:
Thanks for the merge.

Issue #1851, opened 2024-06-08
Different shared library (.so) names for CMake and Meson builds


jrosdahl:
Hi,

I use cpp-httplib in [ccache](https://github.com/ccache/ccache). Thanks for writing and maintaining cpp-httplib!

I noticed that the CMake and Meson build configurations have different opinions on what to call the shared library: CMake produces `libhttplib.so` but Meson produces `libcpp-httplib.so`. Consequently, different Linux distributions now use different library names depending on whether they chose CMake or Meson to build, which means that software that (at least optionally) can use cpp-httplib in its shared library form need to search for both `libhttplib.so` and `libcpp-httplib.so` to stay portable.

So, I just wanted to spread knowledge about this, as I assume it is not intentional to have different names. Perhaps the build configurations should be changed to use the same shared library name?

yhirose:
@jrosdahl thanks for the report.

@TheMaverickProgrammer @jimmy-park @Tachi107 It would be better to have the same shared library name on both CMake and Meson as @jrosdahl mentioned. What do you think?

Tachi107:
Hi @jrosdahl, thanks for noticing! It'd be better to have the same name everywhere of course. I'd personally prefer the library to be called libcpp-httplib, as it is less ambiguous, and the pkg-config file is named like this too.

yhirose:
@Tachi107 @jimmy-park is it possible to change the library name from `libhttplib.so` to `libcpp-httplib.so` on the CMake side as well?

jimmy-park:
@yhirose We can use [OUTPUT_NAME](https://cmake.org/cmake/help/latest/prop_tgt/OUTPUT_NAME.html) CMake property for that purpose.
For example,

```cmake
set_target_properties(httplib PROPERTIES
    OUTPUT_NAME cpp-httplib
)
```

yhirose:
@jimmy-park could you make a pull request including your suggestion?

yhirose:
@jrosdahl this issue has been handled at #1854 by @jimmy-park.
Thanks all!

PR #1854, opened 2024-06-10
Change library name to cpp-httplib

jimmy-park:
Change the library name from `httplib` to `cpp-httplib`.

Refers #1851


yhirose:
@jimmy-park thanks for the quick response!

PR #1857, opened 2024-06-13
avoid memory leaks if linked with static openssl libs

RainerSchielke:


yhirose:
@RainerSchielke could you explain more about this pull request?

RainerSchielke:
We at *Heidelberger Druckmashinen AG* use the httplib with statically linked openssl libs on windows. Without calling OPENSSL_thread_stop we had many memory leaks. I found this solution in the openssl documentation at 

https://www.openssl.org/docs/man3.3/man3/OPENSSL_thread_stop.html
 
**NOTES**
Resources local to a thread are deallocated automatically when the thread exits (e.g. in a pthreads environment, when pthread_exit() is called). On Windows platforms this is done in response to a DLL_THREAD_DETACH message being sent to the libcrypto32.dll entry point. Some windows functions may cause threads to exit without sending this message (for example ExitProcess()). If the application uses such functions, then the application must free up OpenSSL resources directly via a call to OPENSSL_thread_stop() on each thread. **Similarly this message will also not be sent if OpenSSL is linked statically, and therefore applications using static linking should also call OPENSSL_thread_stop() on each thread**. Additionally if OpenSSL is loaded dynamically via LoadLibrary() and the threads are not destroyed until after FreeLibrary() is called then each thread should call OPENSSL_thread_stop() prior to the FreeLibrary() call.

yhirose:
@RainerSchielke thank you for the additional information. I looked at the change, and it looks `OPENSSL_thread_stop` gets called at the end of each `worker`. In other words, `OPENSSL_thread_stop` will be called multiple times if we have more than one workers. Is it OK to call the API more than once? If it's ok, I'll merge the pull request right away.

RainerSchielke:
In my experience, it doesn't matter if **OPENSSL_thread_stop()** is called multiple times. It deallocates all resources and does nothing if the resources are already deallocated in a prior call.

yhirose:
Sounds good. I have just merged it. Thanks for your contribution!

PR #1859, opened 2024-06-17
Allow hex for ipv6 literal addr in redirect

zhlinh:
Fix the support for ipv6 literal address in redirect.

image


yhirose:
@zhlinh thanks for your contribution!

PR #1865, opened 2024-06-20
Fix build on Windows with no WINAPI_PARTITION_APP support

code-disaster:
I'm trying to compile in a Windows environment which doesn't have access to the `*FromApp()` API. Tried to keep the change as unintrusive as possible, so it shouldn't interfere with any desktop or UWP target.


yhirose:
@code-disaster thanks for your contribution!

Issue #1795, opened 2024-03-10
`GetRangeWithMaxLongLength` failing on 32-bit


Tachi107:
Hi, it seems that commit fceada9ef42ce1af64a1d5941d8009dd0520901a broke the `GetRangeWithMaxLongLength` test on 32-bit machines - I've tested in a Linux i686 chroot.

    Running main() from ./googletest/src/gtest_main.cc
    Note: Google Test filter = ServerTest.GetRangeWithMaxLongLength
    [==========] Running 1 test from 1 test suite.
    [----------] Global test environment set-up.
    [----------] 1 test from ServerTest
    [ RUN      ] ServerTest.GetRangeWithMaxLongLength
    ../test/test.cc:2982: Failure
    Expected equality of these values:
      StatusCode::RangeNotSatisfiable_416
        Which is: 416
      res->status
        Which is: 206
    
    ../test/test.cc:2983: Failure
    Expected equality of these values:
      "0"
      res->get_header_value("Content-Length")
        Which is: "7"
    
    ../test/test.cc:2984: Failure
    Expected equality of these values:
      false
      res->has_header("Content-Range")
        Which is: true
    
    ../test/test.cc:2985: Failure
    Expected equality of these values:
      0
      res->body.size()
        Which is: 7
    
    [  FAILED  ] ServerTest.GetRangeWithMaxLongLength (18 ms)
    [----------] 1 test from ServerTest (18 ms total)
    
    [----------] Global test environment tear-down
    [==========] 1 test from 1 test suite ran. (18 ms total)
    [  PASSED  ] 0 tests.
    [  FAILED  ] 1 test, listed below:
    [  FAILED  ] ServerTest.GetRangeWithMaxLongLength
    
     1 FAILED TEST

As shown above, the test response returns a status code of `206` an a `Content-Lenght` of `7`, just like it did before the commit. It's almost like the change was ineffective on 32-bit machines.

I'm pretty sure there's an overflow somewhere in the new code which is only triggered on machines with smaller integer types, but I can't tell where it is.

What do you think is the cause of the problem?

Thanks!

q2a3z:
Hi @Tachi107 
I thinked this is why.
Maybe your 32bit-machines was not returned the Error  416 (Range Not Satisfiable) in here.
```c++:httplib.h
inline bool range_error(Request &req, Response &res) {
  if (!req.ranges.empty() && 200 <= res.status && res.status < 300) {
[...]

PR #1867, opened 2024-06-23
test: fix GetRangeWithMaxLongLength on 32 bit machines

Tachi107:
The test used the hardcoded long value for 64 bit machines even on 32 bit ones, leading to test failures. With this patch the max long length is obtained using `std::numeric_limits<long>::max()`. Thanks to @q2a3z for the hint!

Fixes: https://github.com/yhirose/cpp-httplib/issues/1795

### DIFF
--- a/test/test.cc
+++ b/test/test.cc
@@ -6,6 +6,7 @@
 #include <atomic>
 #include <chrono>
 #include <future>
+#include <limits>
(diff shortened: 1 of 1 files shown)


yhirose:
@Tachi107 thanks a lot for fixing this issue!

PR #1869, opened 2024-06-27
Fix build when targeting Windows 7 as platform.

mol123:
This change makes more of the code introduced in
https://github.com/yhirose/cpp-httplib/pull/1775
conditional on feature macros.

`CreateFile2`, `CreateFileMappingFromApp` and `MapViewOfFileFromApp` are available only starting from Windows 8.

 * https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfile2
 * https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-createfilemappingfromapp
 * https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-mapviewoffilefromapp

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -2823,8 +2823,13 @@ inline bool mmap::open(const char *path) {
     wpath += path[i];
   }
 
+#if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_APP) && (_WIN32_WINNT >= _WIN32_WINNT_WIN8)
   hFile_ = ::CreateFile2(wpath.c_str(), GENERIC_READ, FILE_SHARE_READ,
                          OPEN_EXISTING, NULL);
+#else
+  hFile_ = ::CreateFileW(wpath.c_str(), GENERIC_READ, FILE_SHARE_READ, NULL,
+                         OPEN_EXISTING, FILE_ATTRIBUTE_NORMAL, NULL);
+#endif
 
   if (hFile_ == INVALID_HANDLE_VALUE) { return false; }
 
@@ -2832,7 +2837,7 @@ inline bool mmap::open(const char *path) {
   if (!::GetFileSizeEx(hFile_, &size)) { return false; }
   size_ = static_cast<size_t>(size.QuadPart);
 
-#if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_APP)
+#if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_APP) && (_WIN32_WINNT >= _WIN32_WINNT_WIN8)
   hMapping_ =
       ::CreateFileMappingFromApp(hFile_, NULL, PAGE_READONLY, size_, NULL);
 #else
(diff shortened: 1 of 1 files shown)


Zor-X-L:
@yhirose win7 boxes are still widely available, great downstream projects like llama.cpp are using this great library, I hope it can support win7. ggerganov/llama.cpp#8208

@mol123 I think your code can work but for the actual condition according to win11 sdk header file:

- CreateFile2: `(_WIN32_WINNT >= _WIN32_WINNT_WIN8) && WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_APP | WINAPI_PARTITION_SYSTEM | WINAPI_PARTITION_GAMES)`
- CreateFileMappingFromApp: `(_WIN32_WINNT >= _WIN32_WINNT_WIN8) && WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_APP | WINAPI_PARTITION_SYSTEM)`
- MapViewOfFileFromApp: `(_WIN32_WINNT >= _WIN32_WINNT_WIN8) && WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_APP | WINAPI_PARTITION_SYSTEM)`
- GetFileSizeEx: `WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_APP | WINAPI_PARTITION_SYSTEM | WINAPI_PARTITION_GAMES)`
  - So if only WINAPI_PARTITION_DESKTOP is available, in theory we should use GetFileSize, which has condition `WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP | WINAPI_PARTITION_SYSTEM | WINAPI_PARTITION_GAMES)`

mol123:
* I've updated the conditions used as suggested by @Zor-X-L 
* `GetFileSizeEx` is now used conditionally as well.

yhirose:
@mol123 thanks for the pull request. Could you confirm that Win8 or later use the new APIs and Win7 or lower uses the old APIs? Once you test and confirm it, please let me know.

mol123:
I've tested the code on both Windows 7 and Windows 10 (>= Windows8) and can confirm that it works.

Details:

I've created the following test program:

```
#include <iostream>
#include <string>

#include "httplib.h"

int main() {
  httplib::detail::mmap m("c:\\Windows\\System.ini");
  std::string s(m.data(), m.size());
  std::cout << m.size() << std::endl;
  std::cout << s << std::endl;
}
```

I've compiled it:

 * With `-D_WIN32_WINNT=_WIN32_WINNT_WIN8 -DWINVER=_WIN32_WINNT_WIN8` (targeting Win8 as minimum version) -> `httplibtest8.exe`
 * With `-D_WIN32_WINNT=_WIN32_WINNT_WIN7 -DWINVER=_WIN32_WINNT_WIN7` (targeting Win7 as minimum version) -> `httplibtest7.exe`

On Windows 10, both `httplibtest8.exe` and `httplibtest7.exe` work.

On Windows 7, `httplibtest7.exe` works. `httplibtest8.exe`  fails with the error "The procedure entry point CreateFile2 could not be located in the dynamic link library KERNEL32.dll "

yhirose:
@mol123 @Zor-X-L thanks for your contribution!

qwe857359351a:
@yhirose

yhirose:
@mol123 @Zor-X-L, could you answer the question from @qwe857359351a, since I didn't work on this? Thanks.

mol123:
@qwe857359351a: the code has to be compiled targeting Windows 7 as minimal version, specifically setting `-D_WIN32_WINNT=_WIN32_WINNT_WIN7 -DWINVER=_WIN32_WINNT_WIN7` during compilation.

See:
* https://learn.microsoft.com/en-us/cpp/porting/modifying-winver-and-win32-winnt?view=msvc-170
* https://stackoverflow.com/questions/63874422/how-to-use-visual-studio-2019-to-target-windows-7

qwe857359351a:
> _WIN32_WINNT=_WIN32_WINNT_WIN7 -DWINVER=_WIN32_WINNT_WIN7

Thinks, already able to run.
@yhirose

PR #1873, opened 2024-06-30
Added move assignment operator to Client class.

Hlado:
Fixes issue similar to #1050, solution is based on #1051 approach. After brief inspection i haven't found any drawbacks of making Client class move assignable in addition to move construction.


yhirose:
Thanks!

PR #1886, opened 2024-07-29
fix a bug that cause split.py cannot work properly

Ericple:


yhirose:
Your change causes a build problem...

```
~/cpp-httplib/test$ make test_split
python3 ../split.py -o .
Traceback (most recent call last):
  File "/Users/hirose/Projects/cpp-httplib/test/../split.py", line 42, in <module>
    with open('.' + in_file) as f:
         ^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '.../httplib.h'
make: *** [httplib.cc] Error 1
```

yhirose:
Yes, I am working on MacBook Pro.

PR #1889, opened 2024-07-30
Require the server to use a minimum of TLS 1.2

hanslivingstone:
TLS 1. is deprecated: https://www.ietf.org/rfc/rfc8996.html

This change is only for the server. It will ensure the server isn't subject to a TLS downgrade attack.

Thanks!


yhirose:

PR #1890, opened 2024-07-31
best method for SSE

yz-java:
Implementing SSE without affecting the HTTP processing.

在不影响HTTP处理的情况下实现SSE


yhirose:
@yz-java thank you for your contribution. However, I am not comfortable to merge it. One of the reasons is exposing `sock_fd` from the socket `Stream` class. Also I would like to keep see functionality completely outside of cpp-httplib.

PR #1892, opened 2024-08-07
BoringSSL compatibility fixes

markmentovai:
I’m not sure if BoringSSL compatibility is still desired, but there’s an explicit callout to BoringSSL support at [httplib.h:242](https://github.com/yhirose/cpp-httplib/blob/ae63b89cbf70481ae60515dfd95467e91eecd992/httplib.h#L242), added in 4a7a81e039c216426b22d4787fc9c87528b2ca46. I don’t know if eba980846be1711796106b3c0bb7100bce2394c7, estabilshing the OpenSSL 3 minimum, changes anything. While it’s true that OpenSSL 1 has reached end-of-life, BoringSSL is still supported and largely implements the OpenSSL 1 API. It’s worth trying to upstream this patch, so that I don’t need to maintain it downstream.

This patch is necessary to build cpp-httplib in Crashpad, itself in Chromium, using BoringSSL. [Details here](https://chromium.googlesource.com/crashpad/crashpad/+/1a62a0182557c89494676c06611f1ca731bcb2db).

The fixes include:
 - Library version check: tolerate BoringSSL as an alternative to OpenSSL 3.
 - Don’t call `OPENSSL_thread_stop`, which is not in BoringSSL.
 - Use `SSL_get_peer_certificate` (deprecated in OpenSSL 3), the old name for `SSL_get1_peer_certificate`, because the new name is not in BoringSSL.
 - Call `SSL_set_tlsext_host_name` directly instead of making an `SSL_ctrl` call that BoringSSL does not support. The feared -Wold-style-cast warning that occurs when buidling with OpenSSL is not triggered in BoringSSL.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -269,7 +269,12 @@ using socket_t = int;
 #include <iostream>
 #include <sstream>
 
-#if OPENSSL_VERSION_NUMBER < 0x30000000L
+#if defined(OPENSSL_IS_BORINGSSL)
+#if OPENSSL_VERSION_NUMBER < 0x1010107f
+#error Please use OpenSSL or a current version of BoringSSL
+#endif
+#define SSL_get1_peer_certificate SSL_get_peer_certificate
+#elif OPENSSL_VERSION_NUMBER < 0x30000000L
 #error Sorry, OpenSSL versions prior to 3.0.0 are not supported
 #endif
 
@@ -727,7 +732,7 @@ private:
         fn();
       }
 
-#ifdef CPPHTTPLIB_OPENSSL_SUPPORT
+#if defined(CPPHTTPLIB_OPENSSL_SUPPORT) && !defined(OPENSSL_IS_BORINGSSL)
       OPENSSL_thread_stop();
 #endif
     }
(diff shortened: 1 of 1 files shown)


yhirose:
@markmentovai thanks for the pull request. Since I am not planning to use BoringSSL, I won't be able to maintain this code changes. So if you can accept a responsibility to maintain this BoringSSL related code, I don't mind merging the pull request. (I actually do the same to the CMake build code. Some contributors maintain it for me.) 

As for your changes, I just wonder if you can disallow BoringSSL versions which are no longer supported. (In the case of OpenSSL, I disallow all the "end of life" versions for safety by checking with `OPENSSL_VERSION_NUMBER`.)

markmentovai:
https://github.com/yhirose/cpp-httplib/pull/1892#issuecomment-2274797001:
> @markmentovai thanks for the pull request. Since I am not planning to use BoringSSL, I won't be able to maintain this code changes. So if you can accept a responsibility to maintain this BoringSSL related code, I don't mind merging the pull request. (I actually do the same to the CMake build code. Some contributors maintain it for me.)

I completely understand.

My team and I can certainly contribute patches to maintain BoringSSL support in cpp-httplib as necessary for our use. We wouldn’t be able to commit to continuing ongoing support outside of when we sync to a new version of cpp-httplib. But certainly as we pull new versions of cpp-httplib, we’d need to write the patches to make it work with BoringSSL anyway, and I’m happy to (and in fact, prefer) send those patches back here.

> As for your changes, I just wonder if you can disallow BoringSSL versions which are no longer supported. (In the case of OpenSSL, I disallow all the "end of life" versions for safety by checking with `OPENSSL_VERSION_NUMBER`.)

BoringSSL doesn’t really version itself in the same way that OpenSSL does. In fact, API stability is explicitly listed as not a goal (see its [README.md](https://boringssl.googlesource.com/boringssl/+/2bc9245eb4b50bd9e6b8bd0aa4ac94b91eda997a/README.md)). The current version does identify as OpenSSL 1.1.1, although this is more for API purposes than anything else—and it doesn’t really help with much, because [it’s been at that version](https://boringssl.googlesource.com/boringssl/+blame/68c6fd8943ffba4e5054ff3a9befa8882b6b226a/include/openssl/base.h#101) for [4 years](https://boringssl.googlesource.com/boringssl/+/49e9f67d8b7cbeb3953b5548ad1009d15947a523).

I’ve pinned it to that at your request, I’m just mentioning it so that you know what you’re getting.

yhirose:
@markmentovai thanks for the clear explanation. I have just merged this pr and bumped cpp-httplib version to v0.16.2. Thanks for your fine contribution!

PR #1894, opened 2024-08-09
Post method with encoded URL

hahnnicholas:
This PR adds a Post with URL encoded params at same time that sends the body content-type.
This new api is intended to be used in rest servers that requires specific data in the body while also needs the session id to be sent in the URL.
Tested with a CROW rest server.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -1186,6 +1186,8 @@ public:
               const Params &params);
   Result Post(const std::string &path, const Headers &headers,
               const Params &params, Progress progress);
+  Result Post(const std::string &path, const Params &params,
+              const std::string &body, const std::string &content_type);
   Result Post(const std::string &path, const MultipartFormDataItems &items);
   Result Post(const std::string &path, const Headers &headers,
               const MultipartFormDataItems &items);
@@ -1611,6 +1613,8 @@ public:
               const Params &params);
   Result Post(const std::string &path, const Headers &headers,
               const Params &params, Progress progress);
+  Result Post(const std::string &path, const Params &params,
+              const std::string &body, const std::string &content_type);
   Result Post(const std::string &path, const MultipartFormDataItems &items);
   Result Post(const std::string &path, const Headers &headers,
               const MultipartFormDataItems &items);
@@ -8007,6 +8011,13 @@ inline Result ClientImpl::Post(const std::string &path, const Headers &headers,
               progress);
 }
 
+inline Result ClientImpl::Post(const std::string &path, const Params &params,
+                               const std::string &body,
+                               const std::string &content_type) {
+  std::string path_with_query = append_query_params(path, params);
+  return Post(path_with_query, Headers(), body, content_type);
+}
+
 inline Result ClientImpl::Post(const std::string &path,
                                const MultipartFormDataItems &items) {
   return Post(path, Headers(), items);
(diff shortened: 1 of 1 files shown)


yhirose:
@hahnnicholas thanks for the pull request. But as you can see the signature pattern of Post methods, we have to also add versions which take `Headers`, `Progress`, `ContentProvider` and so on for consistency. Also same Put methods need to be added. In addition to these, enough unit tests need to be added for the new methods...

In fact, this new Post method is a bit confusing to me, since we usually expect that `const Params &params` will be treated in content body as 'application/x-www-form-urlencoded'. But only this particular situation will handle it as query parameters on the path. At least, it should be clearly mentioned in README, otherwise users will get confused like me.

Anyway, I understand what you would like to do and it's a valid request. I'll think over this for a while from the consistent API design standpoint.

For now, you can just call `append_query_params` outside the Post method as workaround. Actually, the one of Get methods does that as below.

https://github.com/yhirose/cpp-httplib/blob/390f2c41f62bf10f41deff1722374826a1b011d8/httplib.h#L7860-L7866

hahnnicholas:
Sorry, I should've write a better explanation about what I'm trying to do. I just made a workarround to my software and would like to send it as a suggestion for furter implementation. Maybe I can work in it latter and help in a better way.

In a short resume, I need to send a post request with the body as 'application/json' fullfiled with a specific object and also put in the request URL some parameters like session ID and etc.

the code bellow exemplifies how I intent to use it:

```C++
httplib::Params params;
params.emplace("session", session);

auto res2 = cli.Post("/load_objects", params, "{\"object\": \"users\"}", "application/json");
if (!res2 || res2->status != httplib::StatusCode::OK_200) {
    auto err = res2.error();
    std::cout << "HTTP error: " << httplib::to_string(err) << std::endl;
    return;
}
std::cout << res2->body << std::endl;
```
Yes, I saw this code you pointed me and was based on that I've made my workarround.
But yes you are right, deal with the same variable, for the same method, in different ways is kind of confusing.
Maybe we need a way to inform how the Params should be send.

Please, let me know if you have any idea of how it should be implemented in a proper way.
Maybe I could help in this task.

yhirose:
@hahnnicholas thank for explaining more details. The workaround that I suggest is as follows:

```cpp
std::string path = httplib::append_query_params(/load_objects, params); 
auto res2 = cli.Post(path, "{\"object\": \"users\"}", "application/json");
```

I know it requires one extra step and does look beautiful, I think it's not too bad.
For now, I don't find any easy and better solution, yet. I'll keep is as 'Open' with 'enhancement'.
Please let me know if you come up with a good idea. Thanks!

hahnnicholas:
@yhirose, yes I have one idea but i don't know.. it will require a lot of changes. Let me explain it for you.

Now the request seems to be fullfiled inside the client object.
For exemple in the function send_with_content_provider:
https://github.com/yhirose/cpp-httplib/blob/6c3e8482f7b4e3b307bb42afbb85fd8771da86b8/httplib.h#L7612-L7625 

Maybe we can move this request object out of this client class and let to the user the reponsability to fullfill this request.
An option is to provide in the Request Class some set methods chained.

``` C++
httplib::Request req =
      httplib::Request()
          .SetCustomHeader(headers)
          .SetUrlParams(urlParams)
          .timeout(10)
          .whatever(2)
          .build();
cli.Post(req);
```

To not change to much the APIs already used, we can provide some alias functions to mostly used API's where the request will be built inside this functions, but more advanced constructions (like this mine) should be made through the Request method.
kind of simple usage and advanced usage.

It will also reduce the number of different API's to achive each combination of parameters.

What do you think?

yhirose:
@hahnnicholas sorry for the late replay. The fluent API that you suggests looks better, but it requires a lot of changes and it has no compatibility to the original API. So I am not planning to implement the fluent style. Anyway, thanks for bringing up the nice idea!

PR #1895, opened 2024-08-10
Fix `KeepAliveTest.SSLClientReconnectionPost`

jimmy-park:
The listening thread could hang because `std::future::wait()` is not called on test failures.

PS. Not related to this issue, but `KeepAliveTest.SSLClientReconnection` and  `KeepAliveTest.SSLClientReconnectionPost` always fail on Windows machines when httplib is built with OpenSSL.


yhirose:
Thanks!

Issue #1901, opened 2024-08-26
LARGE_INTEGER (windows 7,8 not supported?)


UnrealKaraulov:
![image](https://github.com/user-attachments/assets/ed5a850d-98d6-4387-b3d0-bb22c312cacb)

UnrealKaraulov:
```
httplib.h(2862): error C2228: left of '.HighPart' must have class/struct/union 
```

UnrealKaraulov:
Also I has randomly strange warnings

```
                     D:\a\testProjectWin7\testProjectWin7\testProjectWin7\src\httplib.h(2832): warning C4067: unexpected tokens following preprocessor directive - expected a newline [D:\a\testProjectWin7\testProjectWin7\testProjectWin7\testProjectWin7.vcxproj]
                     D:\a\testProjectWin7\testProjectWin7\testProjectWin7\src\httplib.h(2844): warning C4067: unexpected tokens following preprocessor directive - expected a newline [D:\a\testProjectWin7\testProjectWin7\testProjectWin7\testProjectWin7.vcxproj]
                     D:\a\testProjectWin7\testProjectWin7\testProjectWin7\src\httplib.h(2854): warning C4293: '<<': shift count negative or too big, undefined behavior [D:\a\testProjectWin7\testProjectWin7\testProjectWin7\testProjectWin7.vcxproj]
                     D:\a\testProjectWin7\testProjectWin7\testProjectWin7\src\httplib.h(2857): warning C4067: unexpected tokens following preprocessor directive - expected a newline [D:\a\testProjectWin7\testProjectWin7\testProjectWin7\testProjectWin7.vcxproj]
                     D:\a\testProjectWin7\testProjectWin7\testProjectWin7\src\httplib.h(2871): warning C4067: unexpected tokens following preprocessor directive - expected a newline [D:\a\testProjectWin7\testProjectWin7\testProjectWin7\testProjectWin7.vcxproj]
                     ```

UnrealKaraulov:
If Microsoft stopped support Windows XP, 7, 8, 9, that doesn't mean that users don't use these versions of the OS!" Users use Windows 7,8 massive, even WinXP sometimes))

UnrealKaraulov:
Possible invalid ```#if...#else``` order.

yhirose:
@UnrealKaraulov please keep in mind that cpp-httplib doesn't officially support Windows 8 or lower.
[...]

PR #1902, opened 2024-08-27
Fix call to CreateFileMappingW.

mol123:
The call shouldn't rely on the `size` variable, since it is available only conditionally, based on preprocessor macros.

https://github.com/yhirose/cpp-httplib/issues/1901

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -2859,8 +2859,7 @@ inline bool mmap::open(const char *path) {
   hMapping_ =
       ::CreateFileMappingFromApp(hFile_, NULL, PAGE_READONLY, size_, NULL);
 #else
-  hMapping_ = ::CreateFileMappingW(hFile_, NULL, PAGE_READONLY, size.HighPart,
-                                   size.LowPart, NULL);
+  hMapping_ = ::CreateFileMappingW(hFile_, NULL, PAGE_READONLY, 0, 0, NULL);
 #endif
 
   if (hMapping_ == NULL) {
--- a/httplib.h
+++ b/httplib.h
@@ -2851,7 +2851,13 @@ inline bool mmap::open(const char *path) {
   DWORD sizeLow;
   sizeLow = ::GetFileSize(hFile_, &sizeHigh);
   if (sizeLow == INVALID_FILE_SIZE) { return false; }
-  size_ = (static_cast<size_t>(sizeHigh) << (sizeof(DWORD) * 8)) | sizeLow;
+  size_ = sizeLow;
+#if defined(_WIN64)
+  size_ |= (static_cast<size_t>(sizeHigh) << (sizeof(DWORD) * 8));
+#else
+  // The size of the file is too large to fit into `size_`, which is 32 bit.
+  if (sizeHigh != 0) { return false; }
+#endif
 #endif
 
 #if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_APP | WINAPI_PARTITION_SYSTEM) && \


httplib.h
```diff
@@ -2859,8 +2859,7 @@ inline bool mmap::open(const char *path) {
   hMapping_ =
       ::CreateFileMappingFromApp(hFile_, NULL, PAGE_READONLY, size_, NULL);
```
yhirose, line 2860: For consistency, `size_` can be 0 as well.

yhirose:
Could you confirm that passing 0 to the create file mapping functions dons't make any performance regression? (I guess passing length of the file explicitly performs better.)

Also I don't see enough examples which pass 0 length to the APIs...
https://github.com/search?q=CreateFileMappingFromApp+language%3AC%2B%2B&type=code&l=C%2B%2B

httplib.h
mol123: For `CreateFileMappingFromApp`, the documentation doesn't state that size can be 0, so I would continue passing `size_` to stay aligned with the documentation : https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-createfilemappingfromapp

For `CreateFileMappingW`, the documentation explictly states "If this parameter and dwMaximumSizeHigh are 0 (zero), the maximum size of the file mapping object is equal to the current size of the file that hFile identifies." https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-createfilemappingw

yhirose: I got it, thanks!

mol123:
I've run a benchmark of just creating an `httplib::detail::mmap` object of a small file, reading the data, and throwing the mmap away on Windows 7. There is no significant change in performance.

With size set to zero: 10011 ns with 56000 iterations.

With size set explicitly: 10529 ns with 74667 iterations.

UnrealKaraulov:
```
#if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_APP | WINAPI_PARTITION_SYSTEM |   \
                            WINAPI_PARTITION_GAMES)
  LARGE_INTEGER size{};
  if (!::GetFileSizeEx(hFile_, &size)) { return false; }
  size_ = static_cast<size_t>(size.QuadPart);
#else
  DWORD sizeHigh;
  DWORD sizeLow;
  sizeLow = ::GetFileSize(hFile_, &sizeHigh);
  if (sizeLow == INVALID_FILE_SIZE) { return false; }
  size_ = (static_cast<size_t>(sizeHigh) << (sizeof(DWORD) * 8)) | sizeLow;
#endif

#if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_APP | WINAPI_PARTITION_SYSTEM) && \
    (_WIN32_WINNT >= _WIN32_WINNT_WIN8)
  hMapping_ =
      ::CreateFileMappingFromApp(hFile_, NULL, PAGE_READONLY, size_, NULL);
#else
  hMapping_ = ::CreateFileMappingW(hFile_, NULL, PAGE_READONLY, size.HighPart,
                                   size.LowPart, NULL);
#endif

```

```LARGE_INTEGER size``` can missing if ```WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_APP | WINAPI_PARTITION_SYSTEM | WINAPI_PARTITION_GAMES)``` not executed

Maybe need just 

Change
```
#if WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_APP | WINAPI_PARTITION_SYSTEM |   \
                            WINAPI_PARTITION_GAMES)
  LARGE_INTEGER size{};
  if (!::GetFileSizeEx(hFile_, &size)) { return false; }
  size_ = static_cast<size_t>(size.QuadPart);
#else
  DWORD sizeHigh;
  DWORD sizeLow;
  sizeLow = ::GetFileSize(hFile_, &sizeHigh);
  if (sizeLow == INVALID_FILE_SIZE) { return false; }
  size_ = (static_cast<size_t>(sizeHigh) << (sizeof(DWORD) * 8)) | sizeLow;
#endif
```
to 
```
LARGE_INTEGER size{};
if (!::GetFileSizeEx(hFile_, &size)) { return false; }
size_ = static_cast<size_t>(size.QuadPart);
```

If ```GetFileSizeEx``` has starting from WinXP...

File: httplib.h (hunk 1 above)
Comment (sergio-nsk) [reply httplib.h:2866]: How do you come to this code https://github.com/yhirose/cpp-httplib/pull/1902/files#diff-2324f73f4a4958b756eb44d0607268befc5de62db01164ccd73e154da808d8f3R2849-R2861? I think this blocked is never reached by a compiler if Windows SDK is updated (latest one w/ Windows 7).

I would remove this block completely and deal with bit shifts at `LARGE_INTEGER::HighPart`. `LARGE_INTEGER` is supported on Windows XP, `LARGE_INTEGER::QuadPart` is not supported.

Comment (mol123) [reply httplib.h:2866]: According to my SDK's `fileapi.h`:

`GetFileSize` is available if `WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_DESKTOP | WINAPI_PARTITION_SYSTEM | WINAPI_PARTITION_GAMES)`

`GetFileSizeEx` is available if `WINAPI_FAMILY_PARTITION(WINAPI_PARTITION_APP | WINAPI_PARTITION_SYSTEM | WINAPI_PARTITION_GAMES)`

So the else block would be used if only `WINAPI_PARTITION_DESKTOP` would be set.

**Comment by yhirose** [comment]
@mol123 thank you for this pull request. But, I am not comfortable with the changes, since it makes the current code more complicated... I am more interested in completely rewriting the complicated preprocessor conditions to support various Windows environments. (#1903)

**Comment by UnrealKaraulov** [comment]
@yhirose _WIN32_WINNT_WIN8 not found if compile with Windows 7 SDK. This cause issue.

**Comment by mol123** [comment]
As a workaroud, the following can be added before including the header:

```
#ifndef _WIN32_WINNT_WIN8
 #define _WIN32_WINNT_WIN8 0x0602
#endif
```

Issue #1848, opened 2024-06-07
Incorrect usage of urls ending without '/' in static file server (Windows only)


IBEX-BrianSmith:
Found in `cpp-httplib` v0.15.2 running on Windows 10 with Visual Studio 2022 (v143). Note: this behaviour has not been observed on Linux.

If a root mount point '/' is set before other mount points, then this prevents urls addressing the other mount points without a '/' at the end. For example:

```cpp 
#include <iostream>
#include "cpp-httplib/httplib.h"

using namespace httplib;

int main()
{
    Server svr;

    svr.set_mount_point("/", "C:\\dev");
    svr.set_mount_point("/test", "C:\\dev\\test");

    std::jthread lThread([&]() {
        svr.listen("0.0.0.0", 80); });
    std::cin.get();
    svr.stop();
}
```
`curl -v 127.0.0.1/test` results in a `404 Not Found`.
`curl -v 127.0.0.1/test/` results in a `200 OK`.

The workaround is to set the root mount point as the last entry:

```cpp 
...
    svr.set_mount_point("/test", "C:\\dev\\test");
    svr.set_mount_point("/", "C:\\dev");
...
```
`curl -v 127.0.0.1/test` results in a `200 OK`.
`curl -v 127.0.0.1/test/` results in a `200 OK`.

The problem lies in `Server::handle_file_request` where `detail::is_file(path)` is used. This uses [_access_s](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/access-s-waccess-s?view=msvc-170) for Windows which is for files or directories and is incorrect for this use(`detail::is_dir()` does not use this approach). The directory is then opened as a file which fails and hence the `404` response.

This may be linked to #1389.

yhirose:
@IBEX-BrianSmith thanks for the report!

PR #1907, opened 2024-09-02
Automatically redirect http://host/folder to http://host/folder/

### LINKED ISSUE #1389: Accessing Directory
Hello,

I was testing you library and I had a problem to access a directory http://localhost/dir  
Looking into the code I noticed that you check for / at the end so http://localhost/dir/ did work.

I would suggest to change the code to handle this like:

```
inline bool Server::handle_file_request(const Request &req, Response &res,
                                        bool head) {
  for (const auto &entry : base_dirs_) {
    // Prefix match
    if (!req.path.compare(0, entry.mount_point.size(), entry.mount_point)) {
      std::string sub_path = "/" + req.path.substr(entry.mount_point.size());
      if (detail::is_valid_path(sub_path)) {
        auto path = entry.base_dir + sub_path;
        if (path.back() == '/') { path += "index.html"; }

        else if (detail::is_dir(path)) { path += "/index.html"; }
```
Only the last line was added.

Best Regadrs,
Thomas

paulharris:
This will then automatically look for the /index.html file.

This fixes #1389

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -6474,6 +6474,11 @@ inline bool Server::handle_file_request(const Request &req, Response &res,
         auto path = entry.base_dir + sub_path;
         if (path.back() == '/') { path += "index.html"; }
 
+        if (detail::is_dir(path)) {
+           res.set_redirect(sub_path + "/");
+           return true;
+        }
+
         if (detail::is_file(path)) {
           for (const auto &kv : entry.headers) {
             res.set_header(kv.first, kv.second);


paulharris:
I wonder if #1848 will also be fixed with this?

httplib.h
```diff
@@ -6474,6 +6474,11 @@ inline bool Server::handle_file_request(const Request &req, Response &res,
         auto path = entry.base_dir + sub_path;
         if (path.back() == '/') { path += "index.html"; }
 
+        if (detail::is_dir(path)) {
```
sergio-nsk, line 6477: Then it should also have the opposite part
```suggestion
        if (path.back() == '/') {
          path.pop_back();
          if (detail::is_file(path)) {
             <redirect>
             return true;
          }
          path += "/index.html"; }

        if (detail::is_dir(path)) {
```

yhirose:
> I wonder if https://github.com/yhirose/cpp-httplib/issues/1848 will also be fixed with this?

The pull request may fix #1848 though, the real cause of the issue is a problem with `detail::is_file` on Windows.
I have just fixed it.

yhirose:
@paulharris could you please provide at least one unit test case in `test/test.cc`? Thanks!

httplib.h
yhirose: I don't think it's correct.

image

sergio-nsk: example.com does that.

yhirose: NGINX doesn't redirect `file` to `file/` either, but returns as 404.

image

yhirose:
@paulharris I added some test cases and adjust your code to return 301 instead of 302. Thanks for your contribution!

paulharris:
My apologies for the late reply, it was circling back to this but I'm very happy that you liked it and cleaned it up.

re status 302, why not 307 ?  It appears 303 and 307 were introduced to deal with inconsistent implementations.  In this case we would want the client to repeat the POST if that was the original request.
If it should change to GET, then return 303, right?

paulharris:
My apologies for the late reply, it was circling back to this but I'm very happy that you liked it and cleaned it up.

PR #1911, opened 2024-09-04
Fix incorrect handling of Expect: 100-continue

### LINKED ISSUE #1808: Incorrect handling of `Expect: 100-continue` during large file uploads
### Description

When using `curl` to upload large files (>1M) to a server written with cpp-httplib, `curl` adds the `Expect: 100-continue` header. After the server rejects the request by returning a non-100 status code and some data, `curl` blocks for 1 second after receiving the data and then fails with a broken pipe error.

### Steps to Reproduce

1. Set up a server using cpp-httplib.
2. Use `curl` to upload a file larger than 1M to the server.
3. Observe the `--trace` output from `curl`, which shows that it still sends data after handling the `100-continue` response.

    ```
    == Info: Done waiting for 100-continue
    => Send data, 65536 bytes (0x10000)
    ```

### Expected Behavior

According to [RFC 7231 section 5.1.1](https://datatracker.ietf.org/doc/html/rfc7231#section-5.1.1):

```
A server that responds with a final status code before reading the
entire message body SHOULD indicate in that response whether it
intends to close the connection or continue reading and discarding
the request message (see Section 6.6 of [RFC7230]).
```

The expected behavior is for the server to either:

1. Read and discard the remaining request data, or
2. Close the connection immediately with a `Connection: close` header.

### Current Behavior

Currently, cpp-httplib does not handle this scenario correctly. It neither reads and discards the remaining request data nor closes the connection immediately. Instead, it leaves the connection as-is and continues to respond with the `Keep-Alive` header for `CPPHTTPLIB_KEEPALIVE_MAX_COUNT` times if the client does not close the connection in the request header.

### Attempted Solution

The following modification to the code at https://github.com/yhirose/cpp-httplib/blob/548dfff0aef25e36e971af96b49ce7fbb72d840e/httplib.h#L6706 resolves the issue for `curl`:

```c++
default:
  connection_closed = true;
  return write_response(strm, true, req, res);
```

With this change, curl exits without the broken pipe error.

Additional Context
* Link to the relevant curl documentation: https://everything.curl.dev/http/post/expect100.html

solarispika:
Fix #1808

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -6956,7 +6956,9 @@ Server::process_request(Stream &strm, bool close_connection,
       strm.write_format("HTTP/1.1 %d %s\r\n\r\n", status,
                         status_message(status));
       break;
-    default: return write_response(strm, close_connection, req, res);
+    default:
+      connection_closed = true;
+      return write_response(strm, true, req, res);
     }
   }
 
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -24,9 +24,11 @@ else()
     FetchContent_MakeAvailable(gtest)
 endif()
 
+find_package(curl REQUIRED)
+
 add_executable(httplib-test test.cc)
 target_compile_options(httplib-test PRIVATE "$<$<CXX_COMPILER_ID:MSVC>:/utf-8;/bigobj>")
-target_link_libraries(httplib-test PRIVATE httplib GTest::gtest_main)
+target_link_libraries(httplib-test PRIVATE httplib GTest::gtest_main CURL::libcurl)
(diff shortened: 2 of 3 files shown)


yhirose:
@Tachi107 the unit test now depends on libcurl. Could you take care of it for the meson build? Thanks!

Tachi107:
> @Tachi107 the unit test now depends on libcurl. Could you take care of 
> it for the meson build? Thanks!

Sure, thanks for the ping! Just submitted 
<https://github.com/yhirose/cpp-httplib/pull/1914>.

PR #1914, opened 2024-09-04
build(meson): add libcurl test dependency

Tachi107:
Prompted by PR #1911


yhirose:
Thanks for the quick response!

Issue #1917, opened 2024-09-06
Compilation error on gcc 11 or  older


toge:
I met compilation error 0.17.2 on gcc 11 or older. (gcc 13 and 14 work fine)

Error messages are following.
```
include/httplib/httplib.h: In member function 'constexpr size_t httplib::detail::case_ignore_hash::operator()(const string&) const':
include/httplib/httplib.h:359:30: error: call to non-'constexpr' function 'const _CharT* std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::data() const [with _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>]'
  359 |     return hash_core(key.data(), key.size(), 0);
      |                      ~~~~~~~~^~
In file included from /usr/local/include/c++/11.1.0/string:55,
                 from /usr/local/include/c++/11.1.0/bits/locale_classes.h:40,
                 from /usr/local/include/c++/11.1.0/bits/ios_base.h:41,
                 from /usr/local/include/c++/11.1.0/streambuf:41,
                 from /usr/local/include/c++/11.1.0/bits/streambuf_iterator.h:35,
                 from /usr/local/include/c++/11.1.0/iterator:66,
                 from /usr/local/include/c++/11.1.0/bits/ranges_algobase.h:36,
                 from /usr/local/include/c++/11.1.0/bits/ranges_algo.h:35,
                 from /usr/local/include/c++/11.1.0/algorithm:64,
                 from include/httplib/httplib.h:212,
                 from /work/test_package/test_package.cpp:1:
/usr/local/include/c++/11.1.0/bits/basic_string.h:2318:7: note: 'const _CharT* std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::data() const [with _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>]' declared here
 2318 |       data() const _GLIBCXX_NOEXCEPT
      |       ^~~~
```

The constexpr version of `std::string::data()` has been introduced since C++20.
I think removing constexpr from `httplib::detail::case_ignore_hash::operator()` solve this error.

yhirose:
@toge thanks for catching that! Could you send a pull request?

toge:
@yhirose 
There is already PR.
Please check it.
https://github.com/yhirose/cpp-httplib/pull/1918

Thanks @laowai9189!

yhirose:
Thanks a lot!

toge:
[...]

PR #1918, opened 2024-09-06
‘constexpr’ error

laowai9189:
httplib.h: In member function ‘constexpr size_t httplib::detail::case_ignore_hash::operator()(const string&) const’: httplib.h:359:30: error: call to non-‘constexpr’ function ‘const _CharT* std::__cxx11::basic_string<_CharT, _Traits, _Alloc>::data() const [with _CharT = char; _Traits = std::char_traits<char>; _Alloc = std::allocator<char>]’
  359 |     return hash_core(key.data(), key.size(), 0);

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -355,7 +355,7 @@ struct case_ignore_equal {
 };
 
 struct case_ignore_hash {
-  constexpr size_t operator()(const std::string &key) const {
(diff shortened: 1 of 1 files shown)


yhirose:
Same as #1917.

yhirose:
Thanks a lot!

Issue #1805, opened 2024-03-21
about the `:` in path param 


BiuXWT:
A url like this 
```
/usr/:name/number/:id/do:work
```
the `:work` is not a param , the whole `do:work` string is a phrase ,
but now , the framework will consider `work` as a param,could you please fix it

yhirose:
@BiuXWT thanks for the report!

yhirose:
@bgs99 do you have any thoughts on this issue?

bgs99:
> @bgs99 do you have any thoughts on this issue?

I guess we could limit path parameters capture to whole path segments, so `/:param` would be a capture, but `/not:param` would be a literal segment with `:` in the middle.

It might break code for the users that are already depend on capturing segment parts. But on the other hand, it seems that path parameter capture on segment parts is neither fully supported nor documented: it uses `/:` to select between named params capture and regex capture and only full segment capture is shown as examples in the docs.

Alternatively, if we want maximum flexibility, it should be possible to provide some kind of escaping mechanism, like writing `/not\:param` to mean `/not:param` literal.

yhirose:
@bgs99 I like "limit path parameters capture to whole path segments". Could you take care of it and make a pull request, since I don't know much about the `PathParamsMatcher` code. Thanks!

PR #1919, opened 2024-09-06
Only match path params that span full path segment

bgs99:
As discussed in #1805, I have made a change in the path params matcher to only treat `:` as a placeholder if it is in the start of the segment

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -840,7 +840,7 @@ public:
   bool match(Request &request) const override;
 
 private:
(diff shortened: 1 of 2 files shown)


bgs99:
CI errors that are left seem to be irrelevant to the changes - I guess the tests themselves are unstable.

yhirose:
@bgs99 The error on ubuntu is a different issue, and you don't mind. Thanks a lot!

Issue #1906, opened 2024-09-02
Nice way to call "handle_file_request" from user code


paulharris:
Hi, this was partially addressed in #235 ,
I think it would be nice if we could write code along the lines of:
```
srv.Get("/favicon.ico", [=](const httplib::Request &, httplib::Response &res) {
    srv.send_file( location_of_favicon, res );
}
```

or something similar.
The point is `Server::handle_file_request` calls find_content_type() and does the neat mmap async method now, it would be nice to simply be able to ask httplib to send a specific file when desired.

favicon.ico is a simplistic example, but there are other cases where you compute a filename and want to send that.

yhirose:
@paulharris thanks for the idea. There is no need for it in any my projects for now. But a pull request is always welcome!

sonavalepratham:
@paulharris I am beginner in open source & would like to work on this. can you please explain me what needs to be done here?

yhirose:
@paulharris I implemented it. (https://github.com/yhirose/cpp-httplib?tab=readme-ov-file#send-file-content)

```cpp
svr.Get("/content", [&](const Request &req, Response &res) {
  res.set_file_content("./path/to/conent.html");
});

svr.Get("/content", [&](const Request &req, Response &res) {
  res.set_file_content("./path/to/conent", "text/html");
});
```

PR #1927, opened 2024-09-09
Server::respond_with_static_file() for better control over static files

paulharris:
Alternative approach to #1906 
This approach allows library user to still get the benefits of httplib, in particular the valid_path() check and the directory redirect.
The current fix `set_file_content()` is good but skips those things.
Plus, it segfaults when I ask it to `set_file_content(a directory)`.

-----

Can now handle static file requests the same as any Get() request, meaning the handler's priority can be controlled by the developer.

ie
svr.Get("/something", [](auto req, auto res) {
  dont return the file "something";
  instead, return something else;
});

svr.GET("/(.*)", [](auto req, auto res) {
  // chance to further filter or adjust request to suit
  auto result = svr.respond_with_static_file(
                        req, res, req.matches[1], "base_dir");
  // optional chance to override response from the default
  switch (result) {
     SFR_DirRedirect:  break;
     SFR_File:         break; // chance to set additional headers
     SFR_CouldNotOpen: break; // chance to respond nicely
     SFR_NotFound:     break; // chance to respond nicely
  }
});

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -1007,6 +1007,10 @@ public:
   void stop();
   void decommission();
 
+  // helpers, member functions to get access to mime file map and other things
+   enum StaticFileResponse { SFR_DirRedirect, SFR_File, SFR_CouldNotOpen, SFR_NotFound };
+   inline StaticFileResponse respond_with_static_file(const Request &req, Response &res, std::string sub_path, std::string disk_base_dir) const;
+
   std::function<TaskQueue *(void)> new_task_queue;
 
 protected:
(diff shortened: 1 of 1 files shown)


paulharris:
Oh I forgot to mention my solution requires svr, captured in the lambda.
So I don't like the solution, I think yours is better, it just needs to add more smarts.

I've made another PR to consider.

In the end, with mmap() problems, I've now switched to doing it all in the client code and using fread.

yhirose:
Same #1932. Thanks for your understanding!

PR #1931, opened 2024-09-09
Test reading empty zero-length file

paulharris:
I've also discovered mmap() will not map a zero byte file.
So current httplib cannot serve completely empty files.

I created a test case.

### DIFF
--- a/test/test.cc
+++ b/test/test.cc
@@ -2916,6 +2916,15 @@ TEST_F(ServerTest, GetMethod200) {
   EXPECT_EQ("Hello World!", res->body);
 }
 
+TEST_F(ServerTest, GetEmptyFile) {
+  auto res = cli_.Get("/empty_file");
+  ASSERT_TRUE(res);
+  EXPECT_EQ(StatusCode::OK_200, res->status);
+  EXPECT_EQ("text/plain", res->get_header_value("Content-Type"));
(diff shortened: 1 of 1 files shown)


yhirose:
Thanks!

yhirose:
This test fails... (https://github.com/yhirose/cpp-httplib/actions/runs/10781876980/job/29900733140)

paulharris:
Yes, was just a test case, did not include a fix.

Tachi107:
The test failed because it expected the wrong header (`text/plain` as opposed to `application/octet-stream`). So, is cpp-httplib unable to serve empty files or was Paul's conclusion wrong?

yhirose:
> The test failed because it expected the wrong header (`text/plain` as opposed to `application/octet-stream`). So, is cpp-httplib unable to serve empty files or was Paul's conclusion wrong?

You are correct. cpp-httplib serves an empty files with `application/octet-stream`.
https://github.com/yhirose/cpp-httplib/blob/6c93aea59a55eaee128a7abd15ebc88116410254/test/test.cc#L2923

paulharris:
The test was originally hard-crashing due to trying to mmap an empty file.
This was fixed after this test was added, in https://github.com/yhirose/cpp-httplib/commit/9f8db2c230299fee8b9f0aff693262a8671c213a
which was addressing #1933, a related issue.
So now the test was able to progress further to where you saw it fail for a different reason.

PR #1932, opened 2024-09-09
Another attempt at static file handling: set_static_file_content(), which has additional smarts

paulharris:
Does the index.html append, plus handles dir redirection, and checks the url path is valid (ie doesn't .. go up too far the directory tree).

Pushed for you to see, I don't expect this to be accepted without at least a name change!
And perhaps adding an option to specify the content-type.


yhirose:
@paulharris thank you for the pull request, but I am not interested in the implementation since I don't find any situation that I need in my projects...

PR #1937, opened 2024-09-13
Throw exceptions when server's keys can't be loaded.

paulharris:
Hello again!

Here is another PR for your consideration,
there is a good chance you will want to do it a different way, but the goal is to ensure the server doesn't run when it did not successfully load the certificate.

I'm sure I saw a previously reported Issue about this from someone else, but I can't find it at the moment.

Again, good work on the library!

Hrm oh sorry I didn't add a unit test yet ...


yhirose:
Thank you for the pull request, but I am not interested in this.

Issue #1798, opened 2024-03-13
`SSLClientServerTest.*` tests fail with OpenSSL 3.2.1


mgorny:
When attempting to run the test suite on Gentoo Linux amd64, I'm seeing most of the `SSLClientServerTest.*` tests fail:

```
[  FAILED  ] 6 tests, listed below:
[  FAILED  ] SSLClientServerTest.ClientCertPresent
[  FAILED  ] SSLClientServerTest.ClientEncryptedCertPresent
[  FAILED  ] SSLClientServerTest.MemoryClientCertPresent
[  FAILED  ] SSLClientServerTest.MemoryClientEncryptedCertPresent
[  FAILED  ] SSLClientServerTest.TrustDirOptional
[  FAILED  ] SSLClientServerTest.CustomizeServerSSLCtx
```

This is on 548dfff0aef25e36e971af96b49ce7fbb72d840e (but I also reproduce the same result on 0.15.3).

Detailed results:

```
[----------] 8 tests from SSLClientServerTest
[ RUN      ] SSLClientServerTest.ClientCertPresent
test.cc:5160: Failure
Value of: res
  Actual: false
Expected: true
[  FAILED  ] SSLClientServerTest.ClientCertPresent (7 ms)
[ RUN      ] SSLClientServerTest.ClientEncryptedCertPresent
test.cc:5160: Failure
Value of: res
  Actual: false
Expected: true
[  FAILED  ] SSLClientServerTest.ClientEncryptedCertPresent (7 ms)
[ RUN      ] SSLClientServerTest.MemoryClientCertPresent
test.cc:5243: Failure
Value of: res
  Actual: false
Expected: true
[  FAILED  ] SSLClientServerTest.MemoryClientCertPresent (6 ms)
[ RUN      ] SSLClientServerTest.MemoryClientEncryptedCertPresent
test.cc:5243: Failure
Value of: res
  Actual: false
Expected: true
[  FAILED  ] SSLClientServerTest.MemoryClientEncryptedCertPresent (7 ms)
[ RUN      ] SSLClientServerTest.ClientCertMissing
[       OK ] SSLClientServerTest.ClientCertMissing (5 ms)
[ RUN      ] SSLClientServerTest.TrustDirOptional
test.cc:5308: Failure
Value of: res
  Actual: false
Expected: true
[  FAILED  ] SSLClientServerTest.TrustDirOptional (6 ms)
[...]

PR #1940, opened 2024-09-17
test(meson): fix `SSLClientServerTest.*` tests with OpenSSL 3.2.0

Tachi107:
This pull request fixes issue #1798 when building with Meson, while also bumping the minimum required Meson version to 0.62.0. Please do not squash on merge!

Here are the descriptions of the two individual commits:

### build(meson): bump minimum version to 0.62.0

This allows making some minor cleanups

### test(meson): fix `SSLClientServerTest.*` tests with OpenSSL 3.2.0

Since OpenSSL commit <https://github.com/openssl/openssl/commit/342e3652c791bdb06e08abcc169b4456c83ccd00>, the default X.509 certificate format generated with the `openssl req` command has been changed to X.509 v3 from X.509 v1.

For some reason, this change breaks cpp-httplib's `SSLClientServerTest.*` tests.

To fix the test failures, this patch passes the `-x509v1` flag instead of `-x509` when OpenSSL 3.2.0 or newer is detected. To detect the version of a command line utility, Meson 0.62.0 or later is required.

Fixes <https://github.com/yhirose/cpp-httplib/issues/1798>, but only for the Meson build system.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -13,7 +13,7 @@ project(
     'b_lto=true',
     'warning_level=3'
   ],
-  meson_version: '>=0.47.0'
+  meson_version: '>=0.62.0'
 )
 
 # Check just in case downstream decides to edit the source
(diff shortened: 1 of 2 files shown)


yhirose:
@Tachi107 thanks a lot!

PR #1949, opened 2024-09-22
Update main.cc

bhavya2021245:
Added multithreading,configurable parameters and error logging

**Multithreading** :-The server now uses multithreading (svr.new_task_queue = [] { return new httplib::ThreadPool(8); };) to handle up to 8 concurrent requests. This allows the server to scale more efficiently when handling multiple client connections.

**Configurable Parameters**:-The base_dir, host, and port variables remain unchanged, but are now configurable via command-line arguments.

**Error Handling and Logging**:-If the server fails to start (due to issues like port conflicts or missing base directory), a detailed error message is logged, and the server exits with a failure status.

### DIFF
--- a/docker/main.cc
+++ b/docker/main.cc
@@ -11,8 +11,10 @@
 #include <iomanip>
 #include <iostream>
 #include <sstream>
-
 #include <httplib.h>
+#include <csignal>
+#include <string>
+#include <filesystem>
 
 constexpr auto error_html = R"(<html>
 <head><title>{} {}</title></head>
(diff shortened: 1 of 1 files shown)


yhirose:
@bhavya2021245 could you please format main.cc with clangformat? This repo has `.clangformat` and it should be applied. Thanks!

yhirose:
@bhavya2021245 thank you for the update, but it doesn't look like .clangformat configuration has not been applied...
https://github.com/yhirose/cpp-httplib/blob/master/.clang-format

yhirose:
Besides the code formatting problem, I found some other problems. So I am not going to merge this pull request.

> Multithreading :-The server now uses multithreading (svr.new_task_queue = [] { return new httplib::ThreadPool(8); };) to handle up to 8 concurrent requests. This allows the server to scale more efficiently when handling multiple client connections.

The thread count should not be hard-coded, and the value can vary depending on cpu cores and user requirements (cpu bound or I/O bound). cpp-httplib currently uses `std::thread::hardware_concurrency() - 1` threads, or 8 threads if `std::thread::hardware_concurrency() - 1` is less than 8. I may add parameters, so that we can configure the thread count and the max queued request count.

> Configurable Parameters:-The base_dir, host, and port variables remain unchanged, but are now configurable via command-line arguments.

The host should be '0.0.0.0' all the time. As for port, I fixed the port number to 80, because we can map a port to 80 with `-p` parameter like `-p 8080:80`, and I indicated it with the entry `EXPOSE 80` in Dockerfile. I am also ok with `/html` as the base directory.

I took this container image design from The nginx docker image.
https://www.docker.com/blog/how-to-use-the-official-nginx-docker-image/

> Error Handling and Logging:-If the server fails to start (due to issues like port conflicts or missing base directory), a detailed error message is logged, and the server exits with a failure status.

That may be helpful. I'll make a similar change my self which also shows an error message of `Error` in `Result`.

Anyway, thanks for your fine contribution!

PR #1950, opened 2024-09-22
Add CPack support

zjyhjqs:


sum01:
I've only briefly messed with Cpack before, but aren't there quite a few vars that should be set? A lot of things don't seem to have defaults (or not defaults you'd want at least).

I recall that it also required quite a few generator-specific vars (like the dependency variables) be set for some generators (like deb files).

zjyhjqs:
> but aren't there quite a few vars that should be set? A lot of things don't seem to have defaults (or not defaults you'd want at least).

[The main variables required by CPack are defaulted from `project()`](https://cmake.org/cmake/help/latest/module/CPack.html#variables-common-to-all-cpack-generators), so for a modern CMake project it's generally fine.

You could take a try with my presets:
`cmake --workflow --preset "unknown-unknown-clang"`
[CMakeUserPresets.json](https://github.com/user-attachments/files/17101022/CMakeUserPresets.json)

---

> I recall that it also required quite a few generator-specific vars (like the dependency variables) be set for some generators (like deb files).

For the more commonly used packaging tools, .e.g. zip, tgz, that's not necessary. (Indeed, [the mandatary variables for DEB generator have defaults](https://cmake.org/cmake/help/latest/cpack_gen/deb.html))

sum01 (approved):

yhirose:
Thanks for your contribution!

PR #1954, opened 2024-10-02
Add documentation for using Unix domain sockets.

andrewmcdan:
It took me a bit of digging to figure out how to use Unix Sockets, so I added a bit to the Readme about it.


yhirose:
@andrewmcdan thanks for the pull request.

Could you please move the section after "Use poll instead of select" section? 
Also could you make the code example simpler with only necessary information as below?

```c++
// Server
httplib::Server svr("./my-socket.sock");

// Client
httplib::Client cli("./my-socket.sock");
cli.set_address_family(AF_UNIX);
```

Thanks.

andrewmcdan:
Done.

yhirose:
@andrewmcdan thanks for your contribution!

Issue #1952, opened 2024-10-01
Is Sink.is_writable() thread-safe


ochafik:
Hi, I see a [fair bit](https://github.com/search?q=repo%3Ayhirose%2Fcpp-httplib+thread+safe&type=issues&p=2) of work has been done to make httplib threadsafe, wondering if that extends to Sink.is_writable()?

Context: trying to detect closed connections (in a set_content_provider setup) from another thread [in llama.cpp](https://github.com/ggerganov/llama.cpp/pull/9679#pullrequestreview-2339961217). Our writes are still happening only on the http thread, but knowing when to cancel the processing would be amazing.

yhirose:
@ochafik thanks for the interesting question. 

According to the following article, POSIX socket operations like send/recv are atomic.
https://stackoverflow.com/questions/1981372/are-parallel-calls-to-send-recv-on-the-same-socket-valid

`is_writable` method consists of the subsequent `select_write` and `is_socket_alive` calls.
https://github.com/yhirose/cpp-httplib/blob/10d68cff5081d8c7b7cae74000e0c423107dc7b8/httplib.h#L5819-L5822


`select_write` doesn't write, but checks the write status of the socket with `poll` or `select`. I don't know they are atomic, but I think they are because of consistency with other socket system calls.
https://github.com/yhirose/cpp-httplib/blob/10d68cff5081d8c7b7cae74000e0c423107dc7b8/httplib.h#L3104-L3130

`is_socket_alive` ends up calling `recv` eventually, but it's an atomic call as mentioned in the above stackoverflow post.
https://github.com/yhirose/cpp-httplib/blob/10d68cff5081d8c7b7cae74000e0c423107dc7b8/httplib.h#L3189-L3198

So it seems like it's ok to call `is_writable` in different threads, but I am not 100% sure according to my limited knowledge.
Also WinSock2 on Windows may behave differently than Linux or macOS.

Did you find any actual problem in your project? If so I may be able to try to reproduce it on my machine.
Anyway, sorry that I am not able to give you a concrete answer, but hope my answer give you something helpful.

ochafik:
@yhirose thank you so much for the detailed research and prompt reply! 

No actual problem but the question was raised by @ngxson in a [code review](https://github.com/ggerganov/llama.cpp/pull/9679#pullrequestreview-2339961217), hoping to avoid running into problems if that PR gets merged :-D

ngxson:
[...]

PR #1956, opened 2024-10-04
Add {Response,DataSink}::is_alive

ochafik:
This is to be able to detect if the client connection was lost.

Context: https://github.com/yhirose/cpp-httplib/issues/1952 / https://github.com/ggerganov/llama.cpp/pull/9679

As suggested by @ngxson, since is_writable / select_write may be a bit heavyweight, I tried using just select_read (given it's in a context when every other thread is done reading) but it didn't do the trick alone, so settled for is_socket_alive (which uses select_read + read_socket): seems to work.

*Update*: I finally realized having is_alive on DataSink isn't enough, so moved it to Response itself (by lack of a better place?). In the llama.cpp use case we need to test for liveness *before* status / headers are even written back. Added a tiny test to exercise this.

Tested (no new failures): `cmake -B build -DHTTPLIB_TEST=1 -DHTTPLIB_REQUIRE_OPENSSL=0 -DHTTPLIB_COMPILE=1 -DHTTPLIB_TEST=1 -DHTTPLIB_REQUIRE_ZLIB=1 -DHTTPLIB_REQUIRE_BROTLI=1 && cmake --build build -j && ctest --test-dir build -j`

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -530,6 +530,7 @@ public:
 
   std::function<bool(const char *data, size_t data_len)> write;
   std::function<bool()> is_writable;
+  std::function<bool()> is_alive;
   std::function<void()> done;
   std::function<void(const Headers &trailer)> done_with_trailer;
   std::ostream os;
@@ -722,6 +723,7 @@ public:
 
   virtual bool is_readable() const = 0;
   virtual bool is_writable() const = 0;
+  virtual bool is_alive() const = 0;
 
   virtual ssize_t read(char *ptr, size_t size) = 0;
   virtual ssize_t write(const char *ptr, size_t size) = 0;
@@ -2333,6 +2335,7 @@ public:
 
   bool is_readable() const override;
   bool is_writable() const override;
+  bool is_alive() const override;
   ssize_t read(char *ptr, size_t size) override;
   ssize_t write(const char *ptr, size_t size) override;
   void get_remote_ip_and_port(std::string &ip, int &port) const override;
@@ -3205,6 +3208,7 @@ public:
 
   bool is_readable() const override;
   bool is_writable() const override;
+  bool is_alive() const override;
   ssize_t read(char *ptr, size_t size) override;
   ssize_t write(const char *ptr, size_t size) override;
   void get_remote_ip_and_port(std::string &ip, int &port) const override;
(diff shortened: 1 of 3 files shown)


yhirose:
@ochafik thanks for the pull request. I took a look at the change, but I am not comfortable to expose such a method in `Response`.

Also your test code looks working, but it's a very very dangerous code actually.

If I make the following change in  `Server::process_request` in httplib.h to make a response object on heap instead of stack as below, the hidden problem will be revealed.

```cpp
  // Response res; // Original code on stack
  auto resp = detail::make_unique<Response>(); // on heap
  Response& res = *resp;
```

We get a crash when calling `res.is_alive()` the second time in the loop, since `res` has already been gone.

```cpp
        while (count > 0 && res.is_alive()) {
```

I added debug print statements as below, and tested.

```cpp
        std::cout << "enter loop" << std::endl;
        while (count > 0 && res.is_alive()) {
          this_thread::sleep_for(chrono::milliseconds(10));
        }
        std::cout << "leave loop" << std::endl;
```

Here is the result.

```bash
~/Projects/cpp-httplib/test$ make test && ./test --gtest_filter="*ClientCloseDetectionOnResponse*"
clang++ -o test -I.. -g -std=c++11 -I. -Wall -Wextra -Wtype-limits -Wconversion -Wshadow  test.cc include_httplib.cc gtest/gtest-all.cc gtest/gtest_main.cc -DCPPHTTPLIB_OPENSSL_SUPPORT -I/opt/homebrew/opt/openssl@3/include -L/opt/homebrew/opt/openssl@3/lib -lssl -lcrypto -DCPPHTTPLIB_USE_CERTS_FROM_MACOSX_KEYCHAIN -framework CoreFoundation -framework Security -DCPPHTTPLIB_ZLIB_SUPPORT -lz -DCPPHTTPLIB_BROTLI_SUPPORT -I/opt/homebrew/opt/brotli/include -L/opt/homebrew/opt/brotli/lib -lbrotlicommon -lbrotlienc -lbrotlidec -pthread -lcurl
Running main() from gtest/gtest_main.cc
Note: Google Test filter = *ClientCloseDetectionOnResponse*
[==========] Running 1 test from 1 test suite.
[----------] 1 test from LongPollingTest
[ RUN      ] LongPollingTest.ClientCloseDetectionOnResponse
enter loop
libc++abi: terminating due to uncaught exception of type std::__1::bad_function_call: std::exception
Abort trap: 6
```

So at this point, this pull request is not acceptable...

yhirose:
I'll close it for now.

PR #1962, opened 2024-10-14
Remove space between operator"" and _t

pbos:
This should fix a -Wdeprecated-literal-operator instance since this is deprecated as a result of CWG2521 (iiuc C++23).


yhirose:
@pbos thanks!

PR #1965, opened 2024-10-18
Allow empty header values

jimmy-park:
The current implementation rejects the request and responds with `400 Bad Request` if the header value is empty.
However, [RFC 7230](https://datatracker.ietf.org/doc/html/rfc7230#section-3.2) states that empty header values are allowed.

```
header-field   = field-name ":" OWS field-value OWS

field-name     = token
field-value    = *( field-content / obs-fold )
field-content  = field-vchar [ 1*( SP / HTAB ) field-vchar ]
field-vchar    = VCHAR / obs-text

obs-fold       = CRLF 1*( SP / HTAB )
               ; obsolete line folding
               ; see [Section 3.2.4](https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.4)
```

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -4115,7 +4115,7 @@ inline bool parse_header(const char *beg, const char *end, T fn) {
     p++;
   }
 
-  if (p < end) {
+  if (p <= end) {
     auto key_len = key_end - beg;
     if (!key_len) { return false; }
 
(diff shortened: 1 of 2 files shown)


yhirose:
@jimmy-park thanks for the pull request!

PR #1984, opened 2024-11-24
test: Don't check for the exact size of compressed content.

sebastianas:
The testsuite checks for the exact size of the compressed content. The exact size can change if the zlib library is using a different strategy. In thise case using zlib-ng results in a slightly larger content leading to a failure in the test.

Check that the compressed content is less than 10MiB which is a tenth of the orignal content and proves that compression works.


test/test.cc
```diff
@@ -4258,7 +4258,7 @@ TEST_F(ServerTest, PutLargeFileWithGzip2) {
   ASSERT_TRUE(res);
   EXPECT_EQ(StatusCode::OK_200, res->status);
   EXPECT_EQ(LARGE_DATA, res->body);
-  EXPECT_EQ(101942u, res.get_request_header_value_u64("Content-Length"));
+  EXPECT_TRUE(res.get_request_header_value_u64("Content-Length") < (10 * 1024 * 1024));
```
Tachi107, line 4261: Instead of `EXPECT_TRUE`, it'd be better to use [`EXPECT_LT`](https://google.github.io/googletest/reference/assertions.html#EXPECT_LT)

Tachi107 (changes requested):

test/test.cc
yhirose: @sebastianas could you also put some comment before `EXPECT_LT`, so that it clearly explains why this check doesn't use `EXPECT_EQ`? Thanks!

sebastianas: both changes have been made.

yhirose:
@sebastianas thanks for your contribution!

PR #2002, opened 2024-12-04
Available on Win7

maxbad:
May cause abnormal usage Because there is no definition of `_WIN32_WINNT_WIN8` in C:\Program Files (x86)\Microsoft SDKs\Windows\v7.1A\Include\sdkddkver.h


yhirose:
@maxbad thank you for the pull request, but I don't want to make any change to support Win7 which is not officially supported by Microsoft any more. Thanks for your understanding.
https://github.com/yhirose/cpp-httplib?tab=readme-ov-file#windows

PR #2009, opened 2024-12-23
Treat out-of-range last_pos as the end of the content

bobrofon:
RFC-9110 '14.1.2. Byte Ranges':
A client can limit the number of bytes requested without knowing the size of the selected representation. If the last-pos value is absent, **or if the value is greater than or equal to the current length of the representation data, the byte range is interpreted as the remainder of the representation** (i.e., the server replaces the value of last-pos with a value that is one less than the current length of the selected representation).

https://www.rfc-editor.org/rfc/rfc9110.html#section-14.1.2-6

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -5156,7 +5156,18 @@ inline bool range_error(Request &req, Response &res) {
         last_pos = contant_len - 1;
       }
 
-      if (last_pos == -1) { last_pos = contant_len - 1; }
+      // NOTE: RFC-9110 '14.1.2. Byte Ranges':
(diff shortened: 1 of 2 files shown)


yhirose:
@bobrofon thanks for the pull request. Could you please add unit tests for the change?

yhirose:
@bobrofon sorry. You have already added them. :)

yhirose:
@bobrofon thanks for your contribution!

carlosneyzeus (commented):
Obrigada por me curtir gratidão

PR #2010, opened 2024-12-24
fix:set_file_content with range request return 416.

sinnren:
in function "detail::range_error",check range pos and content_length:
`
      // Range must be within content length
      if (!(0 <= first_pos && first_pos <= last_pos &&
            last_pos <= contant_len - 1)) {
        return true;
      }
`

use res.set_file_content in callback, set "content_length_" in res.set_content_provider function, but res.set_content_provider is after "detail::range_error", cause  "detail::range_error" return true(error).

to solve this problem，put "detail::range_error()" after file content served, it's ok.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -7183,14 +7183,6 @@ Server::process_request(Stream &strm, const std::string &remote_addr,
                                       : StatusCode::PartialContent_206;
     }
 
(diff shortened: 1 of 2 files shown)


yhirose:
@sinnren thanks for your contribution!

PR #2022, opened 2025-01-17
Resolve #2017

### LINKED ISSUE #2017: Feature request: ability to check if the connection is still alive
Following up https://github.com/yhirose/cpp-httplib/pull/1956, I would like to discuss more about the ability to add a method to check if the socket is still alive. This can be useful to cancel the underlay processing for a request, as described in this issue: https://github.com/ggerganov/llama.cpp/issues/9273

Currently, one way to do is to use `set_chunked_content_provider`, but the problem is that we cannot change the HTTP status code once we're already inside the callback function.

Therefore, I propose 2 solutions here:

My first suggestion is to add something like `set_async_response`, which works like `set_chunked_content_provider` but send out the whole response, instead of only the body.

Second suggestion is to add `is_alive()` to the `Handler`:
- `Server::routing` already has `Stream &strm` in its scope, so we can pass it to `Server::dispatch_request`
- Then, we can expose it inside `Handler`, by wrapping it inside a function `is_alive()`. I'm not sure how it should look like, but that's the idea

Happy do discuss more to see what's the best way to implement this. Thank you!

yhirose:


ngxson:
This is great, thanks a lot for the rapid response!

I implemented it [this PR](https://github.com/ggerganov/llama.cpp/pull/11285) and it just works.

Do you think it's better to add an example in README? I'd suggest this:

```cpp
// If the handler takes time to finish, you can also poll the connection state

svr.Get("/task", [&](const Request& req, Response& res) {
  const char * result = nullptr;
  process.run(); // for example, starting an external process
  while (result == nullptr) {
    sleep(1);
    if (req.connection.closed) {
      process.kill(); // kill the process
      return;
    }
    result = process.stdout(); // != nullptr if the process finishes
  }
  res.set_content("Done", "text/plain");
});
```

---

Btw, not related to this, but I got this warning when compiling the file:

```
httplib.h:2576:2: warning: extra ';' outside of a function is incompatible with C++98 [-Wc++98-compat-extra-semi]
```

Would you mind to also remove this semicolon, so it can be compiled with `-Werror`? Thanks again.

yhirose:
@ngxson I have updated the code and README according to your suggestions.  Thanks for your fine contribution!

PR #2030, opened 2025-01-26
Fix std::numeric_limits<uint64_t>::max()

baiyies:
Change `std::numeric_limits<uint64_t>::max()` to (`std::numeric_limits<uint64_t>::max)()`. You missed this one.


yhirose:
@baiyies thanks a lot!

PR #2031, opened 2025-01-27
Support building httplib.h on OpenVMS x86 systems

alex-cornford:
Modify for OpenVMS x86 C++. Make tests on OpenVMS currently not supported due to no cmake support. Changes tested on OpenVMS clang C++ and Fedora & GCC

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -218,7 +218,9 @@ using socket_t = SOCKET;
 #include <csignal>
 #include <pthread.h>
 #include <sys/mman.h>
-#include <sys/select.h>
+#ifndef __VMS
+  #include <sys/select.h>
+#endif
 #include <sys/socket.h>
 #include <sys/un.h>
 #include <unistd.h>


yhirose:
@alex-cornford thanks for the report. But this change doesn't make any sense to me... It looks it doesn't include fundamental functions and macros like `select` and `FD_???`...

alex-cornford:
@yhirose I am building this on the OpenVMS operating system, first released in 1977. 
#ifdef __VMS is specific to our legacy operating system. By using a #ifndef my change is ignored on every other OS platform.
Our include files differ from standard unix / linux include files and this change is simple and avoids risks of breaking third party 1980s legacy C code.

I have a simple demo application running on OpenVMS.

yhirose:
@alex-cornford thanks for explaining the background, but I am still not comfortable to accept this pr...

Could you try to build the following simple code to find out which header file should be included for OpenVMS?

```cpp
#include "path/to/httplib.h"

int main(void) {
  httplib::Client cli("http://yhirose.github.io");
  auto res = cli.Get("/hi");
  return 0;
}
```

Then, you can include it as below.

```cpp
#ifdef __VMS
  #include <???> // Specify the header file including `select` and `FD_???` for OpenVMS here to make the example code compiled.
#else
  #include <sys/select.h>
#endif
```

alex-cornford:
@yhirose FD__ functions and and macros are defined in OpenVMS's socket.h. Sorry I cannot connect a client to your test server : security policy excludes internet access for the test OpenVMS system I am using.

All I can offer is that the x86 OpenVMS clang frontend C++ compiler does build successfully with my change, same for my personal Fedora/gcc system.
If you are not comfortable merging this change : that's ok. I'll demo this to third parties as is.

yhirose:
> FD__ functions and and macros are defined in OpenVMS's socket.h.

Should the OpenVMS's socket.h be included as `sys/socket.h`? If so, I am fine with your PR since it's already in httplib.h.

If it should be included as `socket.h`, should the PR look like the following?

```cpp
#ifdef __VMS
  #include <socket.h>
#else
  #include <sys/select.h>
#endif
```

alex-cornford:
@yhirose that looks like a breaking change. You need to include socket.h on other operating systems.
The source includes

#ifdef CPPHTTPLIB_USE_POLL
#include <poll.h>
#endif
#include <csignal>
#include <pthread.h>
#include <sys/mman.h>
#ifndef __VMS
  #include <sys/select.h>
#endif
#include <sys/socket.h>
#include <sys/un.h>
#include <unistd.h>

using socket_t = int;
#ifndef INVALID_SOCKET
#define INVALID_SOCKET (-1)
#endif


etc....
linux platforms etc need both select.h and socket.h
Due to non-standard include files, OpenVMS only needs socket.h

yhirose:
Do you mean that `#include <sys/socket.h>` provides the `select` function and `FD_???` macros on OpenVMS? If not, which include file on OpenVMS contains them? I am just curious. :)

alex-cornford:
@yhirose yes, select is defined within socket.h on OpenVMS.

yhirose:
Great!

PR #2040, opened 2025-02-04
Preparations for WebSockets

falbrechtskirchinger:
This PR lays the groundwork for an upcoming implementation of WebSockets. It makes the following changes:

1. **Stream handler:** Introduce a new callback `StreamHandler`. If set, it replaces all other content-serving mechanisms and – in conjunction with a response handler – provides a minimal interface to implement a WebSocket client. The `Response` struct gained a `set_stream_handler()` function to accomplish the same server-side.
2. **Stream changes:** Add `nonblocking_read_size()` which returns a size that is guaranteed to not block `read()` calls. `read()` may return fewer bytes including none at all. ~~Add timeout setters, to allow other protocol implementations to set more appropriate timeouts in stream handlers.~~
3. **`random_bytes()`:** Refactor `random_string()` into `random_bytes()`. This will be used to generate the masking key for WebSockets.
4. **`select_read()`:** Add a two FD version of `select_read()` with support for infinite timeouts. This change is critical to implement bidirectional communication. The WebSockets implementation will use this to interrupt reads when a message is added to the send queue. Branching on the non-type template parameter ensures the common, single FD case isn't pessimized.
5. **Forward declaration:** Forward declare a few functions used by the WebSockets header to allow it to compile with split `httplib.{h,cc}`.

I'm aware of #1103 and while I liked it initially, IMHO, it falls short of its goal of supporting protocol switching generically. The stream handler mechanism doesn't bill itself as anything but another way to handle the body of an HTTP response. It's meant as a low-level, advanced feature that doesn't have to be exposed in the high-level API. (Protocol switching isn't limited to `GET` requests.)
I've happily taken some inspiration from it, though. (Thanks, @PixlRainbow!)

**Checklist:**
- [x] Explore requirements for WebSocket servers.
- [ ] Replace all occurrences of `std::function<bool(Stream&)>` with `StreamHandler`?
- [ ] Test on platforms other than Linux. Also, test with `CPPHTTPLIB_USE_POLL` where applicable.
- [ ] Test split header/implementation.
- [ ] Documentation.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -517,6 +517,9 @@ using Progress = std::function<bool(uint64_t current, uint64_t total)>;
 struct Response;
 using ResponseHandler = std::function<bool(const Response &response)>;
 
+class Stream;
+using StreamHandler = std::function<bool(Stream &strm)>;
+
 struct MultipartFormData {
   std::string name;
   std::string content;
@@ -634,6 +637,7 @@ struct Request {
 
   // for client
   ResponseHandler response_handler;
+  StreamHandler stream_handler;
   ContentReceiverWithProgress content_receiver;
   Progress progress;
 #ifdef CPPHTTPLIB_OPENSSL_SUPPORT
@@ -1148,6 +1152,7 @@ enum class Error {
   Compression,
   ConnectionTimeout,
   ProxyConnection,
+  StreamHandler,
 
   // For internal use only
   SSLPeerCouldBeClosed_,
@@ -2199,6 +2204,7 @@ inline std::string to_string(const Error error) {
   case Error::Compression: return "Compression failed";
   case Error::ConnectionTimeout: return "Connection timed out";
   case Error::ProxyConnection: return "Proxy connection failed";
+  case Error::StreamHandler: return "Stream handler failed";
   case Error::Unknown: return "Unknown";
   default: break;
   }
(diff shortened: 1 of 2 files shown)


falbrechtskirchinger:
@yhirose I'd appreciate some guidance as I re-evaluate an aspect of this PR:
The addition of `size_t Stream::nonblocking_read_size()` is essentially a hack and the least invasive solution I could think of, to `bool Stream::is_readable()` blocking until timeout.
I remembered reading about someone asking for a global timeout, similar to what Curl offers. This may have seemed difficult to implement but is available as a side-effect of my WebSockets efforts. Would you be amenable to a more invasive change to `Stream`?
Here's my current idea of the revised API:
```c++
// This checks if data is readable NOW and returns immediately
bool Stream::is_readable();

// This behaves like is_readable() today
bool Stream::wait_readable();

enum WaitResult {
  Timeout = 0,
  Readable = (1 << 0),
  Notified = (1 << 1), // EventPipe was signaled from another thread
};

// Waits on socket (assuming SocketStream) and EventPipe, allowing interruption
WaitResult Stream::wait_readable(EventPipe &);
```

Every major platform has a different way of dealing with I/O multiplexing and signaling (`select()` etc. with eventfd on Linux, I/O completion ports on Windows, and kqueue on macOS and the BSDs, etc.).
`EventPipe` is my answer to NOT implementing any of that. It simply wraps a pair of sockets on Windows and a pipe on every other platform. It allows one thread to signal another thread blocking on `select_read()`. The WebSockets code uses it primarily to notify the I/O thread that new messages have been queued and are ready to be sent. I'll implement a simple, very coarse userspace timer on top of it next, to periodically send keep-alive pings, and trigger timeouts when expected messages aren't received within a certain time.

That same mechanism should be able to handle global timeouts. **Are you interested in a PR that explores implementing a global timeout?** The WebSockets PR is growing quite large and I'm concerned about expecting you to review it all, given that you've mentioned you have no personal interest in WebSockets. If I can land part of it independently, that should help. (As an aside, the WebSocket code lives in a separate file. This PR contains the only changes to `httplib.h` I need to make it work and avoid code duplication.)

falbrechtskirchinger:
Re-submitting in smaller pieces.

PR #2041, opened 2025-02-05
Generalize split.py and keep inline namespaces

falbrechtskirchinger:
Generalize `split.py` to split future httplib files. Files are now searched in the script directory and relative search directories. `inline namespace`s will be kept.

This is intended to be able to split the upcoming WebSockets implementation in `httplib-ws.h`.


yhirose:
I'll close it since this is not really necessary.

Issue #609, opened 2020-08-12
Request handler with ContentReader not called without Content-Length header


csk-ableton:
Handlers with a `ContentReader` are called when sending a request that contains the HTTP header `Content-Length: 0` but not if the body is empty and this header is missing. This is surprising and unfortunately not possible to test this with `httplib::Client`, as the `Content-Length` header is appended automatically. The behavior changed in a recent version of the library, which broke our API when used in a browser that is not appending the `Content-Length` header but not unit tests that where using `httplib::Client`.

# Reproduction

Given the following example server:
```cpp
httplib::Server& httpServer;
httpServer.Delete("/delete", []( const httplib::Request& request, httplib::Response& response, const httplib::ContentReader& reader) { 
  response.set_content("ok", "text/plain"); 
});
httpServer.listen("localhost", 80);
```

Make the following curl call
```
curl -v -X DELETE http://localhost/delete
```

Expected:
```
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 80 (#0)
> DELETE /delete HTTP/1.1
> Host: localhost
> User-Agent: curl/7.64.1
> Accept: */*
> 
< HTTP/1.1 200 OK
< Access-Control-Allow-Origin: *
< Content-Length: 2
< Content-Type: text/plain
< 
* Connection #0 to host localhost left intact
ok* Closing connection 0
```

Actual:
```
*   Trying ::1...
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 80 (#0)
> DELETE /delete HTTP/1.1
> Host: localhost
> User-Agent: curl/7.64.1
> Accept: */*
> 
< HTTP/1.1 404 Not Found
< Content-Length: 0
< 
* Connection #0 to host localhost left intact
* Closing connection 0
```

yhirose:
@csk-ableton, thanks for the feedback.

> The behavior changed in a recent version of the library

Which version (or commit number) was working?

csk-ableton:
> Which version (or commit number) was working?

The code worked fine in v0.6.3 but didn't after upgrading to v0.7.0. I tried v0.7.4 but it's the same behavior.

yhirose:
@csk-ableton, thanks! I'll take a look at it when I have time.

csk-ableton:
Great thanks!

[...]

PR #2044, opened 2025-02-08
Refine when content is expected

### LINKED ISSUE #2028: Potential memory corruption in `stream_line_reader`
Hi,

I would like to report a recent problem that I found on the latest version 0.18.5

To test it, I compiled this simple program:

```cpp
#include <httplib.h>

int main(void)
{
  using namespace httplib;

  Server svr;

  svr.set_pre_routing_handler([](const auto& req, auto& res) {
    if (req.method == "OPTIONS") {
      res.set_content("world", "text/html");
      return Server::HandlerResponse::Handled;
    }
    return Server::HandlerResponse::Unhandled;
  });

  svr.Get("/hi", [](const Request& req, Response& res) {
  });

  // Match the request path against a regular expression
  // and extract its captures
  svr.Get(R"(/numbers/(\d+))", [&](const Request& req, Response& res) {
    auto numbers = req.matches[1];
    res.set_content(numbers, "text/plain");
  });

  // Capture the second segment of the request path as "id" path param
  svr.Get("/users/:id", [&](const Request& req, Response& res) {
    auto user_id = req.path_params.at("id");
    res.set_content(user_id, "text/plain");
  });

  // Extract values from HTTP headers and URL query params
  svr.Get("/body-header-param", [](const Request& req, Response& res) {
    if (req.has_header("Content-Length")) {
      auto val = req.get_header_value("Content-Length");
    }
    if (req.has_param("key")) {
      auto val = req.get_param_value("key");
    }
    res.set_content(req.body, "text/plain");
  });

  svr.Get("/stop", [&](const Request& req, Response& res) {
    svr.stop();
  });

  svr.listen("0.0.0.0", 8080);
}
```

Upon running it, I place a breakpoint when calling `parse_request_line`

```cpp
  // Request line and headers
  if (!parse_request_line(line_reader.ptr(), req) ||       // <== breakpoint
      !detail::read_headers(strm, req.headers)) {
    return write_response(strm, close_connection, req, res);
  }
```

The buffer of `stream_line_reader` seems to contains data from somewhere else:

Image

Image

Please note that, this happens 80% of the time. When it work correctly, the buffer must contains HTTP request data:

Image

When the server first started, the first request never has this problem. This suggest that there maybe race condition between threads when accessing a shared variable.

falbrechtskirchinger:
Consider `Content-Length` and `Transfer-Encoding` headers when determining whether to expect content. ~~Don't expect content for `DELETE` requests and~~ don't handle the HTTP/2 connection preface pseudo-method `PRI`.

Fixes #2028.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -5381,10 +5381,14 @@ write_multipart_ranges_data(Stream &strm, const Request &req, Response &res,
 
 inline bool expect_content(const Request &req) {
   if (req.method == "POST" || req.method == "PUT" || req.method == "PATCH" ||
-      req.method == "PRI" || req.method == "DELETE") {
+      req.method == "DELETE") {
     return true;
   }
-  // TODO: check if Content-Length is set
+  if (req.has_header("Content-Length") &&
+      req.get_header_value_u64("Content-Length") > 0) {
+    return true;
+  }
+  if (is_chunked_transfer_encoding(req.headers)) { return true; }
   return false;
 }
 


yhirose:
I added a unit test for #609 for you. You can rebase your branch with the master to get the test.
https://github.com/yhirose/cpp-httplib/commit/eb10c22db10995c4a30f92e01968df1319f8783e

falbrechtskirchinger:
> @falbrechtskirchinger thanks for the pull request, but #609 appears again with this change. Could you take a look at it?

Sure, I'll take a look.


> I added a unit test for #609 for you. You can rebase your branch with the master to get the test. [eb10c22](https://github.com/yhirose/cpp-httplib/commit/eb10c22db10995c4a30f92e01968df1319f8783e)

Thanks!

falbrechtskirchinger:
@yhirose I've re-added `DELETE` to `expect_content()`. Test passes now.

httplib.h
```diff
@@ -5381,10 +5381,14 @@ write_multipart_ranges_data(Stream &strm, const Request &req, Response &res,
 
 inline bool expect_content(const Request &req) {
   if (req.method == "POST" || req.method == "PUT" || req.method == "PATCH" ||
-      req.method == "PRI" || req.method == "DELETE") {
+      req.method == "DELETE") {
```
ngxson, line 5384: Nice, thanks for the fix. Btw as explained in https://github.com/yhirose/cpp-httplib/issues/2028#issuecomment-2645912843 , we should also add `OPTIONS` here (or maybe replace `HEAD` with `OPTIONS` here, I'm not sure if the browser even use HEAD or not)

falbrechtskirchinger: No problem. As long as `Content-Length` is set, things should work. The explicit handling of `DELETE` is specifically for handlers with content receivers. For `OPTIONS` requests, we have no such handlers.

I cannot reproduce any issues with `OPTIONS`. If that still fails, let me know what you're doing and I'll check on my end.

yhirose:
@falbrechtskirchinger thanks for the fine contribution!

PR #2046, opened 2025-02-09
Fix check for URI length to prevent incorrect HTTP 414 errors

brettp:
This fixes a bug where the full stream's length, including POST data, is checked instead of just the URI. This results in false `414 URI Too Long` errors. The fix is to check only the URI part of the first line of the stream.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -7234,18 +7234,18 @@ Server::process_request(Stream &strm, const std::string &remote_addr,
 #endif
 #endif
 
-  // Check if the request URI doesn't exceed the limit
-  if (line_reader.size() > CPPHTTPLIB_REQUEST_URI_MAX_LENGTH) {
-    Headers dummy;
-    detail::read_headers(strm, dummy);
-    res.status = StatusCode::UriTooLong_414;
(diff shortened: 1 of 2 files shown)


yhirose:
@brettp could you please add at least one unit text in `test/test.cc`?

httplib.h
```diff
@@ -7162,16 +7162,23 @@ Server::process_request(Stream &strm, const std::string &remote_addr,
 #endif
 #endif
 
+  auto req_line = line_reader.ptr();
+
+  // Extract URI from the request line
+  std::string method, uri, version;
+  std::istringstream iss(req_line);
+  iss >> method >> uri >> version;
```
falbrechtskirchinger, line 7170: This is effectively what `parse_request_line()` does. Shouldn't the check be moved there?

brettp:
@yhirose - Test added for request lines larger than max URI length but with a URI itself that is shorter.

@falbrechtskirchinger - I remove the extra parsing by checking the length of `req.target` after `parse_request_line()` is called.

yhirose:
@brettp thanks for your fine contribution!

Issue #1601, opened 2023-06-27
Client Get operation stalls in getaddrinfo when network is down


TheMostDiligent:
Hello!

We are seeing the following issue when the network is completely unavailable: the Get operations stalls in the [getaddrinfo](https://github.com/yhirose/cpp-httplib/blob/bd9612b81e6f39ab24a4be52fcda93658c0ca2dc/httplib.h#L2705) for about 30 seconds. We set the connection timeout and read timeout, but they don't have any effect.

Is it possible that this logic also respects the connection timeout?

yhirose:
@TheMostDiligent thanks for the feedback. As far as I know, `getaddrinfo` is a blocking function and doesn't provide timeout option. So there is no way for me to deal with it... Do you know how other HTTP libraries handle this situation?

TheMostDiligent:
I am not very profound in HTTP handling. However, I tried to research the subject a bit and apparently there is an alternative function: [getaddrinfo_a](https://man7.org/linux/man-pages/man3/getaddrinfo_a.3.html) that supports asynchronous wait mode (`GAI_NOWAIT`). 
In this mode, the function exits immediately and an [event](https://man7.org/linux/man-pages/man7/sigevent.7.html) is triggered when the operation completes. I think this will allow to wait for the event in the loop and abort the loop if timeout elapses. Or even better, if the user-provided callback returns false similar to the callback of the Get function.

yhirose:
It doesn't look like only Linux supports it...

TheMostDiligent:
> It looks like only Linux supports it

It does seem so. On Mac, there is [DNSServiceGetAddrInfo](https://developer.apple.com/documentation/dnssd/1804700-dnsservicegetaddrinfo) that allows async wait. [This article](https://eggerapps.at/blog/2014/hostname-lookups.html) describes it in detail. Looks like it is quite a hassle to implement this...

On Windows, I actually did not see that problem.

yhirose:
It seems like we may use `GetAddrInfoEx` which supports timeout parameter.
https://learn.microsoft.com/en-us/windows/win32/api/ws2tcpip/nf-ws2tcpip-getaddrinfoexw

In any case, supporting this feature on all three platforms and ensuring they work correctly will not likely be an easy task for me. So please don't expect me to implement it in the near future. A pull request is always welcome! :)

yhirose:
This stackoverflow post is helpful to write thread timeout code.
[...]

PR #2048, opened 2025-02-10
Fix #2034

### LINKED ISSUE #2034: Global timeout feature (same as "--max-time" curl option)
Hi,

I've finally had some time to investigate the issue reported [here](https://github.com/yhirose/cpp-httplib/issues/1733), and created a small example:

A client that wants to have a reply in at most 2100 ms:

```c++
#include <httplib.h>

int main()
{

  httplib::Client http_client("http://127.0.0.1:8082");
  auto timeout = std::chrono::milliseconds{2100};
  http_client.set_connection_timeout(timeout);
  http_client.set_read_timeout(timeout);
  http_client.set_write_timeout(timeout);
  auto start = std::chrono::steady_clock::now();
  http_client.Get("/");
  auto end = std::chrono::steady_clock::now();
  auto elapsed = std::chrono::duration_cast<std::chrono::milliseconds>(end - start);
  std::cout << "Elapsed time: " << elapsed.count() << "ms\n";
}
```

A simple server to test (in python):

```python
from http.server import BaseHTTPRequestHandler, HTTPServer
import time

class SimpleHTTPRequestHandler(BaseHTTPRequestHandler):
    def do_GET(self):
        if self.path == "/":
            self.send_response(200)
            self.send_header("Content-type", "text/html")
            self.end_headers()
            for i in range(5):
                time.sleep(1)
                self.wfile.write(f"Response part {i+1}\n".encode('utf-8'))            
        else:
            self.send_response(404)
            self.end_headers()

def run(server_class=HTTPServer, handler_class=SimpleHTTPRequestHandler):
    server_address = ('', 8082)
    httpd = server_class(server_address, handler_class)
    print("Starting httpd server on port 8082...")
    httpd.serve_forever()

if __name__ == "__main__":
    run()
```

It will print something like

Elapsed time: 5001ms

while I was expecting something close to

Elapsed time: 2100ms

Perhaps there is another way to set a timeout for the request?

Thanks

yhirose:


httplib.h
```diff
@@ -2332,10 +2366,12 @@ void split(const char *b, const char *e, char d,
...
-bool process_client_socket(socket_t sock, time_t read_timeout_sec,
-                           time_t read_timeout_usec, time_t write_timeout_sec,
-                           time_t write_timeout_usec,
-                           std::function<bool(Stream &)> callback);
+bool process_client_socket(
+    socket_t sock, time_t read_timeout_sec, time_t read_timeout_usec,
+    time_t write_timeout_sec, time_t write_timeout_usec,
+    time_t global_timeout_msec,
+    std::chrono::time_point<std::chrono::steady_clock> start_time,
+    std::function<bool(Stream &)> callback);
```
falbrechtskirchinger, line 2374: Wouldn't it make sense at some point to move those/some parameters into a `Session` or `SocketConfig` struct? It's quite unwieldy already.

falbrechtskirchinger:
Do you want to see my solution to this at all? It would have the added benefit of making requests cancellable in the future if that were ever desired. Also, it keeps additional timeout logic out of the stream classes, which partially clashes with my stream handler requirements (though, that can be worked around). And I intend to address the DNS issue as well. That's not (yet) included in your global timeout.

I'll need another day or two to get it ready.

httplib.h
yhirose: Yes, the code can be refactored later.

yhirose:
> Do you want to see my solution to this at all? ~It would have the added benefit of making requests cancellable in the future if that were ever desired.~ Also, it keeps additional timeout logic out of the stream classes, which partially clashes with my stream handler requirements (though, that can be worked around). And I intend to address the DNS issue as well. That's not (yet) included in your global timeout.
> 
> I'll need another day or two to get it ready.
> 
> Edit: Never mind, canceling doesn't require event pipes and is already implemented.

Currently, I am not very much interested in it unless it is absolutely needed. But if there are things that improve the httplib code base, could you split your change into several smaller pull requests? I usually don't accept a big pull request, because I don't have time to understand such.

As for the DNS issue, could you take a look at https://github.com/yhirose/cpp-httplib/issues/1601? If you have a multi platform solution, could you send it as a pull request? Thanks!

falbrechtskirchinger:
> > Do you want to see my solution to this at all? ~It would have the added benefit of making requests cancellable in the future if that were ever desired.~ Also, it keeps additional timeout logic out of the stream classes, which partially clashes with my stream handler requirements (though, that can be worked around). And I intend to address the DNS issue as well. That's not (yet) included in your global timeout.
> > I'll need another day or two to get it ready.
> > Edit: Never mind, canceling doesn't require event pipes and is already implemented.
> 
> Currently, I am not very much interested in it unless it is absolutely needed. But if there are things that improve the httplib code base, could you split your change into several smaller pull requests?

I've thought more about it and my code no longer adds anything of value to the core of httplib.

> I usually don't accept a big pull request, because I don't have time to understand such.

Well, I hope you might accept a larger PR (WebSockets) as part of `example/` or possibly `contrib/`. 

> 
> As for the DNS issue, could you take a look at #1601? If you have a multi platform solution, could you send it as a pull request? Thanks!

Unfortunately, I was only interested in the issue as a vehicle to get WebSockets-related code included/reviewed. Sorry! I use httplib mostly for IPC using UNIX domain sockets and right now WebSockets for interacting locally with Chrome DevTools.
I would have just used threads. It's not great, but it works everywhere.

PR #2049, opened 2025-02-11
Don't run CI twice (on push AND pull request)

falbrechtskirchinger:
I noticed your GitHub Actions running twice (seen [here](https://github.com/yhirose/cpp-httplib/pull/2039/checks)). This should fix it.


yhirose:
@falbrechtskirchinger thanks a lot!

Issue #2043, opened 2025-02-08
SOVERSION does not indicate binary compatibility accurately


mgorny:
Currently, `SOVERSION` on the shared library is set to match two version components:

https://github.com/yhirose/cpp-httplib/blob/996acc52537a62435769bf1f873ad0504d6cf620/CMakeLists.txt#L196

This indicates that all 0.18.x releases should be backwards compatible. However, this is not the case. The most recent example is #2022 adding `is_connection_closed` field in middle of `Request` class. This caused all field offsets to change, causing software linked to the previous version of the library to start accessing wrong fields.

Please either update `SOVERSION` to correctly indicate binary compatible, or — even better — set stronger binary compatibility guarantees. This library is used by some huge stuff (for example, LLVM) and people need to spend hours rebuilding it whenever ABI of cpp-httplib changes.

mgorny:
Well, my first suggestion would be to look at [abi-compliance-checker](https://github.com/lvc/abi-compliance-checker) tool. If you build two versions of the shared library with debug symbols, you can use it like this:

```
abi-dumper -lver 0.18.5 path/to/libcpp-httplib.so.0.18.5 -o old.dump
abi-dumper -lver 0.18.6 path/to/libcpp-httplib.so.0.18.6 -o new.dump
abi-compliance-checker -l cpp-httplib -old old.dump -new new.dump
```

…and it will generate a nice HTML report that tells if you anything ABI-breaking happened, and explain why it breaks ABI. Long story short, if it points out any backwards compatibility breakage, you need to raise library SOVERSION.

Of course, anytime that happens, it means that all consumers linking to the shared library need to rebuilt. This is especially problematic for LLVM since it takes a few hours to build, and an average modern Gentoo system needs 2-3 versions of LLVM simultaneously, so basically a cpp-httplib ABI change means whole day spent rebuilding LLVM.

Unfortunately, changes to private fields of public classes also count as ABI changes. There are some ways to avoid that, though. If the class is never allocated by the caller, you could add new fields to the very end — in that case ABI is preserved. If  the class is allocated by the caller, you can still avoid ABI changes by using the PImpl idiom.

yhirose:
@mgorny could you explain more about the "shared library" such as 'libcpp-httplib.so.0.18.5'? (I am not maintaining such binary static library files, since cpp-httplib is "A C++11 single-file header-only cross platform HTTP/HTTPS library".)

mgorny:
[...]

PR #2054, opened 2025-02-14
ci: add abidiff workflow

Tachi107:
This CI workflow checks ABI compatibility between the pushed commit and the previous, helping preventing accidental ABI breaks.

Helps with https://github.com/yhirose/cpp-httplib/issues/2043

Cc: @mgorny

### DIFF
--- a/.github/workflows/abidiff.yaml
+++ b/.github/workflows/abidiff.yaml
@@ -1,0 +1,64 @@
+# SPDX-FileCopyrightText: 2025 Andrea Pappacoda <andrea@pappacoda.it>
+# SPDX-License-Identifier: MIT
+
+name: abidiff
+
+on: [push, pull_request]
+
+defaults:
+  run:
+    shell: sh
+
+jobs:
+  abi:
+    runs-on: ubuntu-latest
+    container:
+      image: debian:testing
+
+    steps:
+    - name: Install dependencies
+      run: apt -y --update install --no-install-recommends
+        abigail-tools
+        ca-certificates
+        g++
+        git
+        libbrotli-dev
+        libssl-dev
+        meson
+        pkg-config
+        python3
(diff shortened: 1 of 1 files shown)


Tachi107:
Note: I'm not sure this works correctly on pull requests, since they may contain more than one commit.

.github/workflows/abidiff.yaml
```diff
@@ -0,0 +1,62 @@
...
+        path: current
+
+    - uses: actions/checkout@v4
+      with:
+        path: previous
+        fetch-depth: 2
+
+    - name: Checkout previous
+      working-directory: previous
+      run: git checkout HEAD^
```
yhirose, line 43: Could you try `git checkout $(git describe --tags --abbrev=0 master)`?
`$(git describe --tags --abbrev=0 master)` gives us the commit hash for the latest version tag.

Tachi107: You mean diffing the current commit with the previous release? Brilliant! Didn't think about it :)

yhirose:
> Note: I'm not sure this works correctly on pull requests, since they may contain more than one commit.

Thanks a lot for taking care of it! I made a comment about how to get the commit hash for the latest version tag. Could you try with it to see how it goes?

Tachi107:
> Thanks a lot for taking care of it! I made a comment about how to get the commit hash for the latest version tag. Could you try with it to see how it goes?

Done! The job now fails because it has detected breaking changes since the last tagged release, and abidiff returns a status code of `12`, which means `ABIDIFF_ABI_CHANGE` and `ABIDIFF_ABI_INCOMPATIBLE_CHANGE` according to its [manpage](https://sourceware.org/libabigail/manual/abidiff.html).

yhirose:
Thanks for the great work!

PR #2057, opened 2025-02-16
Fix compilation failures with include <windows.h>

falbrechtskirchinger:
Recent changes used `min` without parentheses, which is known to conflict with `#include <windows.h>`.


yhirose:
Thanks!

PR #2058, opened 2025-02-16
Detect additional CMake build failures

falbrechtskirchinger:
Add `include_httplib.cc` to the main test executable (already done in `Makefile`), and add `include_windows_h.cc` to the main test executable on Windows to test if including `windows.h` conflicts with `httplib.h`.

Detects breakage fixed in #2057.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -26,7 +26,7 @@ endif()
 
 find_package(CURL REQUIRED)
 
-add_executable(httplib-test test.cc)
+add_executable(httplib-test test.cc include_httplib.cc $<$<BOOL:${WIN32}>:include_windows_h.cc>)
 target_compile_options(httplib-test PRIVATE "$<$<CXX_COMPILER_ID:MSVC>:/utf-8;/bigobj>")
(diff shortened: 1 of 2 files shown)


yhirose:
@sum01 @abouvier @jimmy-park @Tachi107 do you have any comment since I am not very miuch familiar to CMake?

Tachi107:
¯\_(ツ)_/¯

The change looks good, but I don't really care about Windows

falbrechtskirchinger:
> ¯\_(ツ)_/¯
> 
> The change looks good, but I don't really care about Windows

Neither do I, but I've been trained to always parenthesis `min` and `max` by repeated issues in `nlohmann/json`.

sum01:
It seems fine, but I've never really looked at the project's test cmake stuff anyways.

But isn't it easier to just `#define NOMINMAX` instead of dealing with those issues anyways?

falbrechtskirchinger:
> It seems fine, but I've never really looked at the project's test cmake stuff anyways.
> 
> But isn't it easier to just `#define NOMINMAX` instead of dealing with those issues anyways?

They have been dealt with previously, so I'm just continuing the practice. And it seems to continuously bite people, as evidenced by the issue trackers.

yhirose:
Thanks all!

Issue #1481, opened 2023-02-02
detail::is_socket_alive() is not work for https connection


chisheng:
is_socket_alive() is always return ture even though peer closed TLS and TCP connecton,  I dont konwn how to detect  whether TLS connection is alived.

yhirose:
@chisheng thanks for the feedback. This method is not intended to be used by users, rather for internal use in cpp-httplib to detect a current socket is writable. Also there exist some unit cases where you can see it returns 'false'. One of them is `ServerTest.ClientStop`. Hope it helps.

chisheng:
Sorry，I didn't explain my problem clearly.Pseudocode
int main()
{
httplib::SSLClient cli();
cli.post();//first request is success
sleep(60);//The server will close the connection after 60 seconds
cli.post()//second request will failed，is_socket_alive() return true for TLS socket ,it should reconnect TLS instead of use previous one. 
}

yhirose:
@chisheng thanks for the more details. I'll try to reproduce it.

> sleep(60);//The server will close the connection after 60 seconds

Does it actually mean that your server will close **within** 60 seconds?

chisheng:
yes,the keep-live timeout of http server  is 60s

yhirose:
@chisheng I now confirmed it's a bug. Thanks for the report.

yhirose:
@chisheng I fixed it. Could you try it with the latest httplib.h in your project?

yhirose:
@chisheng it seems like my fix disable KeepAlive connection completely... Sorry about that. I'll work on it again.

yhirose:
@chisheng I now fixed it. Could you try the latest httplib.h in the master branch? If it works on your machine, I'll bump up the version number. Thanks!

chisheng:
> @chisheng I now fixed it. Could you try the latest httplib.h in the master branch? If it works on your machine, I'll bump up the version number. Thanks!

    sorry,I think it is still have a problem,I usring 0.9.7 because It wasn't implemented at the time that using post method with provicer ，I add extended post port with provider According to my situation in small Small memory device。

Function with provider may be not work ，beacause i  find write_request return false，SSL_peek is not executed.and SSL_peek（）return 0 mean TLS broken? did not use SSL_get_error() to  get error ,I guess SSL_get_error  return  SSL_ERROR_WANT_READ when The server response delay ，SSL_peek return 0 in this situation.

yhirose:
[...]

PR #2053, opened 2025-02-14
Refactor setting socket options

falbrechtskirchinger:
Add `detail::set_socket_opt()` and `detail::set_socket_opt_time()` to avoid repetition of platform-specific code.

PR #2059, opened 2025-02-16
Simplify SSL shutdown

falbrechtskirchinger:
There are multiple problems with the `MaxTimeoutTest` timeout exceedance. As a first step, this simplifies SSL shutdown and also makes it correct on Windows (where `SSL_shutdown()` was only called once, AFAICT).

To handle max timeout, we need to bring in the timeout values and request start time, which is messy. Worse, even if `ssl_delete` is measured to take 0ms, max timeout is still exceeded elsewhere. This was the case with both versions I've tested.

As a sneak peek, here's what I tried today and hinted at [yesterday](https://github.com/yhirose/cpp-httplib/pull/2053#issuecomment-2661074043):
<details>
<summary>Code</summary>

```c++
// In nonblocking mode, SSL_shutdown() returns 0 until the close_notify
// alerts have been sent and recevied, with SSL_get_error() providing
// details
set_nonblocking(sock, true);

auto res = 0;
time_t timeout_sec, timeout_usec;
while ((res = SSL_shutdown(ssl)) != 1) {
  auto err = SSL_get_error(ssl, res);
  switch (err) {
  case SSL_ERROR_WANT_READ: {
    auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(
                        std::chrono::steady_clock::now() - start_time)
                        .count();
    calc_actual_timeout(max_timeout_msec, duration, read_timeout_sec,
                        read_timeout_usec, timeout_sec, timeout_usec);
    if (timeout_sec == 0 && timeout_usec == 0) { break; }
    if (select_read(sock, timeout_sec, timeout_usec) > 0) { continue; }
    break;
  }
  case SSL_ERROR_WANT_WRITE: {
    auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(
                        std::chrono::steady_clock::now() - start_time)
                        .count();
    calc_actual_timeout(max_timeout_msec, duration, write_timeout_sec,
                        write_timeout_usec, timeout_sec, timeout_usec);
    if (timeout_sec == 0 && timeout_usec == 0) { break; }
    if (select_write(sock, timeout_sec, timeout_usec) > 0) { continue; }
    break;
  }
  default: break;
  }
  break;
}

set_nonblocking(sock, false);
```

</details>

I can share both versions and the related changes in a PR for you to decide what to do with.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -9062,18 +9062,13 @@ inline void ssl_delete(std::mutex &ctx_mutex, SSL *ssl, socket_t sock,
   // Note that it is not always possible to avoid SIGPIPE, this is merely a
   // best-efforts.
   if (shutdown_gracefully) {
-#ifdef _WIN32
     (void)(sock);
-    SSL_shutdown(ssl);
-#else
-    detail::set_socket_opt_time(sock, SOL_SOCKET, SO_RCVTIMEO, 1, 0);
-
-    auto ret = SSL_shutdown(ssl);
-    while (ret == 0) {
-      std::this_thread::sleep_for(std::chrono::milliseconds{100});
-      ret = SSL_shutdown(ssl);
+    // SSL_shutdown() returns 0 on first call (indicating close_notify alert
+    // sent) and 1 on subsequent call (indicating close_notify alert received)
+    if (SSL_shutdown(ssl) == 0) {
+      // Expected to return 1, but even if it doesn't, we free ssl
+      SSL_shutdown(ssl);
     }
-#endif
   }
 
   std::lock_guard<std::mutex> guard(ctx_mutex);


yhirose:
@falbrechtskirchinger thanks for the pull request. I am a bit scared to touching `ssl_delete`, since the code fixed some very sensitive bugs. Could you please carefully check the following issues and code changes first, then make sure that your change doesn't break any of them? If there are not enough unit test cases, it's ok to add additional tests in 'test/test.cc'.

* https://github.com/yhirose/cpp-httplib/pull/701
* https://github.com/yhirose/cpp-httplib/pull/1527
* https://github.com/yhirose/cpp-httplib/pull/1921

falbrechtskirchinger:
> @falbrechtskirchinger thanks for the pull request. I am a bit scared to touching `ssl_delete`, since the code fixed some very sensitive bugs. Could you please carefully check the following issues and code changes first, then make sure that your change doesn't break any of them? If there are not enough unit test cases, it's ok to add additional tests in 'test/test.cc'.
> 
>     * [Fix multiple threading bugs including #699 and #697 #701](https://github.com/yhirose/cpp-httplib/pull/701)
> 
>     * [Fix #1481 (with content provider) #1527](https://github.com/yhirose/cpp-httplib/pull/1527)
> 
>     * [Fix KeepAliveTest.SSLClientReconnectionPost problem #1921](https://github.com/yhirose/cpp-httplib/pull/1921)

After some reading, I certainly understand your hesitation. Still, I believe this to be the correct solution per the man page `SSL_shutdown(3)`:
```
Shutdown Lifecycle
   Regardless of whether a shutdown was initiated locally or by the peer, if
   the underlying BIO is blocking, a call to SSL_shutdown() will return firstly
   once a close_notify alert message is written to the peer (returning 0), and
   upon a second and subsequent call, once a corresponding message is received
   from the peer (returning 1 and completing the shutdown process). Calls to
   SSL_shutdown() with a blocking underlying BIO will also return if an error
   occurs.
```
In case of an error, we `SSL_free()` anyway, but otherwise, we attempt to shut down the TLS connection properly/gracefully by sending and receiving the close_notify alert message.

`SSL_shutdown()` should be invoked exactly twice with a blocking socket. The old code called once on Windows and looped unnecessarily everywhere else (or tried to, since `1` should be returned on the second call, unless an error occurs).

Concerning max timeout, I've tried using a blocking socket and a non-blocking socket. Both implementations worked, though I prefer the non-blocking solution, as it doesn't rely on the internals of `SSL_shutdown()` for timing out.

yhirose:
@falbrechtskirchinger thanks for the answer and the reference. It seems this pull request change is better than the current code, and I like this simpler code. I'll merge it and see if any unexpected issue will arise. Thanks for the excellent change and research!

PR #2060, opened 2025-02-16
Print timeout exceedance in MaxTimeoutTest

falbrechtskirchinger:
Just added some helpful output to better judge the size of the problem.


yhirose:
Thanks!

PR #2062, opened 2025-02-16
Add style check to workflow

falbrechtskirchinger:
Add a non-fatal check to see if sources have been formatted. Unfortunately, there are minor version differences. Version 18 (the default in `ubuntu-latest`) formats differently than versions 19 and 20.
I'm not including the examples for now to not make too much noise.


yhirose:
@falbrechtskirchinger thanks for the nice tool. 👍

Instead of using the clang-format-19, could you please try to use the version in the latest-ubuntu by adjusting .clang-format file in the repo?

Could you also exclude /test/httplib.h and /test/httplib.cc from the style check since they are automatically generated?

Also could you make an adjustment to `clean:` section to remove files generated by the style check like *.formatted?

Thanks!

falbrechtskirchinger:
> @falbrechtskirchinger thanks for the nice tool. 👍
> 
> Instead of using the clang-format-19, could you please try to use the version in the latest-ubuntu by adjusting .clang-format file in the repo?

The problem is two lines, where an initialization of a struct is mistaken for a definition of one and a space is inserted, where it doesn't belong. Changing the style of the initialization will probably fix that and simplify the workflow.

> Could you also exclude /test/httplib.h and /test/httplib.cc from the style check since they are automatically generated?

Sure.

> Also could you make an adjustment to `clean:` section to remove files generated by the style check like *.formatted?

`.formatted` files are removed via this line:
```sh
rm -f $$file.formatted; \
```

Do you want it in `clean` anyway?

yhirose:
> Do you want it in clean anyway?

Oh, I missed that... You can have to change `clean:` 😃

falbrechtskirchinger:
You'll notice these lines in the diff of `httplib.h`:

```diff
-  struct in6_addr addr6{};
-  struct in_addr addr{};
+  struct in6_addr addr6 {};
+  struct in_addr addr {};
```

Changing these initializations to ` = {};` will make them format consistently across versions 18, 19, and 20.

yhirose:
> You'll notice these lines in the diff of `httplib.h` (in the style check output):
> 
> ```diff
> -  struct in6_addr addr6{};
> -  struct in_addr addr{};
> +  struct in6_addr addr6 {};
> +  struct in_addr addr {};
> ```
> 
> Changing these initializations to ` = {};` will make them format consistently across versions 18, 19, and 20.

Sounds good to me. Please go ahead to make the change in httplib.h.

falbrechtskirchinger:
> Sounds good to me. Please go ahead to make the change `= {};` in httplib.h, as well as other format errors. Thanks!

Done. #2063

I've also added the example files, as it only affected one file.

yhirose:
Thanks for the adjustments!

Issue #2061, opened 2025-02-16
SSLClient get connection server certificate without redoing lib work


urosg80:
As I understand to get the server certificate for a SSLClient request I can use set_server_certificate_verifier and use my own function but I must then redo all the server certificate validation that is normally done inside the lib.

Is there another way?

yhirose:
@urosg80 As you mentioned, once we use set_server_certificate_verifier, then it skips the normal verification in cpp-httplib and we are totally responsible for the verification... Here is the code.

https://github.com/yhirose/cpp-httplib/blob/748f47b3772d39b9c8705fc636b2530d13c7f13c/httplib.h#L9626-L9653

If we want to do both our manual check and the built-in check, this code needs to be adjusted.

You can find the background of this method at https://github.com/yhirose/cpp-httplib/issues/1431.

urosg80:
Yes I found the reference and the code responsible inside the library to do the verification but to avoid code duplication I implemented a simple change to the "server_certificate_verifier" signature and built-in verification code part.

#https://github.com/urosg80/cpp-httplib/commit/c27d1fdbb579f0e6690bb20bbae2d38ececcb362

if you think this could be usable for others I can publish a pull request

yhirose:
@urosg80 yes, it's useful to others! Could you please send a pull request for the change?

urosg80:
pull request published #2064

yhirose:
#2064 has been merged.

PR #2064, opened 2025-02-17
server_certificate_verifier extended to reuse built-in verifier

urosg80:
#2061 https://github.com/yhirose/cpp-httplib/issues/2061

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -435,6 +435,12 @@ private:
 
 } // namespace detail
 
+enum SSLVerifierResponse {
+  Verified, // connection certificate is verified and accepted
+  CheckAgain, // use the built-in certificate checker again
+  Declined // connection certificate was process but is declined
+};
+
 enum StatusCode {
   // Information responses
   Continue_100 = 100,
@@ -1483,7 +1489,7 @@ public:
 #ifdef CPPHTTPLIB_OPENSSL_SUPPORT
   void enable_server_certificate_verification(bool enabled);
   void enable_server_hostname_verification(bool enabled);
-  void set_server_certificate_verifier(std::function<bool(SSL *ssl)> verifier);
+  void set_server_certificate_verifier(std::function<SSLVerifierResponse(SSL *ssl)> verifier);
 #endif
 
   void set_logger(Logger logger);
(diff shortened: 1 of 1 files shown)


httplib.h
```diff
@@ -435,6 +435,12 @@ struct scope_exit {
 
 } // namespace detail
 
+enum SSLVerifierResponse {
+  Verified, // connection certificate is verified and accepted
+  CheckAgain, // use the built-in certificate checker again
+  Declined // connection certificate was process but is declined
+};
```
falbrechtskirchinger, line 442: Might I suggest different names:

```cpp
enum SSLVerifierResponse {
  NoDecisionMade,        // no decision has been made, use built-in certificate verifier
  CertificateAccepted,   // connection certificate is verified and accepted
  CertificateRejected    // connection certificate was processed but is rejected
};
```

Also, order `NoDecisionMade` to the top. It makes more sense to be the default enum value and is certainly less dangerous than `CertificateAccepted`.

Just two cents from a random contributor...

httplib.h
```diff
@@ -9623,12 +9629,20 @@ inline bool SSLClient::initialize_ssl(Socket &socket, Error &error) {
...
-            if (!server_certificate_verifier_(ssl2)) {
-              error = Error::SSLServerVerification;
-              return false;
-            }
-          } else {
+          // set default status to CheckAgain
+          SSLVerifierResponse verificationStatus = SSLVerifierResponse::CheckAgain;
+
+          if (server_certificate_verifier_) 
+            verificationStatus = server_certificate_verifier_(ssl2);
```
falbrechtskirchinger, line 9636: This code base uses curlies even around single statements. Also, please run `clang-format`. This PR would fail #2062.

falbrechtskirchinger (commented):
Two notes. Not sure how @yhirose is going to feel about changing public API.

httplib.h
urosg80: Proposed change I think is not bad, will update PR.

But since I first posted #2061 and asked if I should publish PR and got an "OK" from @yhirose I think it should be OK to be proposing public API change, although not a major one IMHO; final decision is still his.

httplib.h
```diff
@@ -9623,12 +9634,20 @@ inline bool SSLClient::initialize_ssl(Socket &socket, Error &error) {
         }
 
         if (server_certificate_verification_) {
+          SSLVerifierResponse verification_status_ =
```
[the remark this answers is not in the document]
urosg80: done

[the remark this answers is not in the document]
urosg80: sorry, missed that; fixed

yhirose:
@urosg80 the code looks good. Thanks for your contribution!

urosg80:
> @urosg80 the code looks good. Thanks for your contribution!

Great and thank you for the great library!

PR #2065, opened 2025-02-17
Make poll() the default

falbrechtskirchinger:
`poll()` is available on:
  - AIX
  - FreeBSD
  - Genode
  - Haiku
  - HP-UX
  - Illumos/Solaris
  - ***Linux***
  - MINIX 3
  - ***MacOSX***
  - OpenBSD
  - Plan 9
  - QNX
  - RTEMS
  - VxWorks
  - ***Windows***
  - Zephyr
  - z/OS USS

`poll()` is not available on:
  - NonStop OS
  - OpenVMS

`select()` can still be enabled by defining `CPPHTTPLIB_USE_SELECT`.

Most users likely won't see any performance improvements (or the contrary), but the WebSockets code may see a minuscule improvement.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -31,14 +31,21 @@ jobs:
       (github.event_name == 'pull_request'  &&
        github.event.pull_request.head.repo.full_name != github.event.pull_request.base.repo.full_name) ||
       (github.event_name == 'workflow_dispatch' && github.event.inputs.test_linux == 'true')
+    strategy:
+      matrix:
+        select_impl: ['select', 'poll']
     steps:
       - name: checkout
         uses: actions/checkout@v4
       - name: install libraries
         run: sudo apt-get update && sudo apt-get install -y libbrotli-dev libcurl4-openssl-dev
       - name: build and run tests
+        env:
+          SELECT_IMPL: ${{ matrix.select_impl }}
         run: cd test && make
       - name: run fuzz test target
+        env:
+          SELECT_IMPL: ${{ matrix.select_impl }}
         run: cd test && make fuzz_test
 
   macos:
@@ -48,12 +55,19 @@ jobs:
       (github.event_name == 'pull_request'  &&
        github.event.pull_request.head.repo.full_name != github.event.pull_request.base.repo.full_name) ||
       (github.event_name == 'workflow_dispatch' && github.event.inputs.test_macos == 'true')
+    strategy:
+      matrix:
+        select_impl: ['select', 'poll']
     steps:
       - name: checkout
         uses: actions/checkout@v4
       - name: build and run tests
+        env:
+          SELECT_IMPL: ${{ matrix.select_impl }}
         run: cd test && make
       - name: run fuzz test target
+        env:
+          SELECT_IMPL: ${{ matrix.select_impl }}
         run: cd test && make fuzz_test
 
   windows:
@@ -63,6 +77,9 @@ jobs:
       (github.event_name == 'pull_request'  &&
        github.event.pull_request.head.repo.full_name != github.event.pull_request.base.repo.full_name) ||
       (github.event_name == 'workflow_dispatch' && github.event.inputs.test_windows == 'true')
+    strategy:
+      matrix:
+        select_impl: ['select', 'poll']
     steps:
     - name: Prepare Git for Checkout on Windows
       run: |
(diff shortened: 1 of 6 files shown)


yhirose:
@falbrechtskirchinger I am fine to change the default behavior from `select` to `poll`, but `select` support is still very important to me.

My biggest concern with this pull request is that `select` is no longer tested at all on GitHub Actions CI. Could you make necessary changes to the GitHub Actions Workflow files to run the unit test suite with `select` on all the platform? I know it will cause the testing time double, but it's absolutely necessary cost. Thanks!

falbrechtskirchinger:
> My biggest concern with this pull request is that `select` is no longer tested at all on GitHub Actions CI. Could you make necessary changes to the GitHub Actions Workflow files to run the unit test suite with `select` on all the platform? I know it will cause the testing time double, but it's absolutely necessary cost. Thanks!

Agreed. I was worried that `poll()` hadn't been tested before, so testing both is a good idea and 20 jobs can run in parallel, so test time shouldn't be any longer.

falbrechtskirchinger:
@sum01 @abouvier @jimmy-park @Tachi107 I'm sure @yhirose would appreciate you looking over the CMake changes in this PR. I've added an option `HTTPLIB_USE_SELECT` which defines `CPPHTTPLIB_USE_SELECT`. Thanks!

falbrechtskirchinger:
@yhirose All builds output the command line now (added `/v:m /clp:ShowCommandLine /nologo` to Windows builds). You can verify that `CPPHTTPLIB_USE_SELECT` is defined where needed.

yhirose:
Looks great!

Tachi107:
Poll is already enabled in the Debian cpp-httplib package, so it is already widely deployed :)

Note to self (or to whoever wants to do so): the option should also be added to the meson build scripts

falbrechtskirchinger:
> Poll is already enabled in the Debian cpp-httplib package, so it is already widely deployed :)

Good to hear. Shockingly the Gentoo ebuild doesn't enable it.

> Note to self (or to whoever wants to do so): the option should also be added to the meson build scripts

I didn't care enough to add an option for the archaic `select()`. Is meson even available on any OS that doesn't have `poll()`? Does meson run without `poll()`? :laughing:

Tachi107:
Il 18 febbraio 2025 13:07:53 CET, Florian Albrechtskirchinger ***@***.***> ha scritto:
>falbrechtskirchinger left a comment (yhirose/cpp-httplib#2065)

>I didn't care enough to add an option for the archaic `select()`. Is meson even available on any OS that doesn't have `poll()`? Does meson run without `poll()`? :laughing: 

Yeah, makes sense! Maybe it should just enable it by default, or when poll is detected on the system. Will do.

alex-cornford:
> @falbrechtskirchinger this suggestion sounds good to me.
> 
> @alex-cornford, as far as I know, you are the only user that uses OpenVMS. So once I accept this pull request, you will have to define `CPPHTTPLIB_USE_SELECT` from the next version. :)

@yhirose  OpenVMS builds and runs with and without CPPHTTPLIB_USE_SELECT defined.

falbrechtskirchinger:
> > @falbrechtskirchinger this suggestion sounds good to me.
> > @alex-cornford, as far as I know, you are the only user that uses OpenVMS. So once I accept this pull request, you will have to define `CPPHTTPLIB_USE_SELECT` from the next version. :)
> 
> @yhirose OpenVMS builds and runs with and without CPPHTTPLIB_USE_SELECT defined.

@alex-cornford Great! So, am I just misinformed then and OpenVMS does have `poll()` or did I make a mistake? Adding some print statements to check...

alex-cornford:
> > > @falbrechtskirchinger this suggestion sounds good to me.
> > > @alex-cornford, as far as I know, you are the only user that uses OpenVMS. So once I accept this pull request, you will have to define `CPPHTTPLIB_USE_SELECT` from the next version. :)
> > 
> > 
> > @yhirose OpenVMS builds and runs with and without CPPHTTPLIB_USE_SELECT defined.
> 
> @alex-cornford Great! So, am I just misinformed then and OpenVMS does have `poll()` or did I make a mistake? Adding some print statements to check...
> 
> Edit: No, it's definitely using `poll()` by default and `select()` with `CPPHTTPLIB_USE_SELECT` defined. 👍

The OpenVMS C++ compiler supports poll() and select(). I don't know how it works internally.

yhirose:
@falbrechtskirchinger according to your list, 'NonStop OS' is the only environment and I don't think there is any user using cpp-httplib on the OS. So I am now thinking of going further, which means removing select() completely and using poll() only. It can simplify the code and reduce my maintenance cost. What do you think?

falbrechtskirchinger:
> @falbrechtskirchinger according to your list, 'NonStop OS' is the only environment and I don't think there is any user using cpp-httplib on the OS. So I am now thinking of going further, which means removing select() completely and using poll() only. It can simplify the code and reduce my maintenance cost. What do you think?

I have double-checked my list since I was wrong about OpenVMS. 
- VxWorks only has `select()` (it has several functions including the word poll, which threw off my search).
- I've re-confirmed some of the other ones, I was less sure about: RTEMS (only since 2021), QNX, Zephyr
- Genode and NonStop OS have neither `select()` nor `poll()`.

So, it's only VxWorks, AFAICT. But, cpp-httplib doesn't build on VxWorks right now anyway:
```c
#  ifdef VXWORKS
#   include <sockLib.h>
#   include <selectLib.h>
```
[https://www.ee.torontomu.ca/~courses/ee8205/Data-Sheets/Tornado-VxWorks/vxworks/ref/selectLib.html#select](https://www.ee.torontomu.ca/~courses/ee8205/Data-Sheets/Tornado-VxWorks/vxworks/ref/selectLib.html#select)

Removing `select()` doesn't seem unreasonable. Just give me the go-ahead, and I'll submit a PR.

yhirose:
> Removing select() doesn't seem unreasonable. Just give me the go-ahead, and I'll submit a PR.

Good. Could you please proceed to make the change? Thanks a lot!

PR #2067, opened 2025-02-19
build(meson): automatically use poll or select as needed

Tachi107:
Follow-up to 6e73a63153e936e753211a5608acf336fb848a37


yhirose:
@Tachi107 thanks!

PR #2069, opened 2025-02-19
Refactor streams: rename `is_*` to `wait_*` for clarity

falbrechtskirchinger:
- Replace `is_readable()` with `wait_readable()` and `is_writable()` with `wait_writable()` in the Stream interface.
- Implement a new `is_readable()` function with semantics that more closely reflect its name. It returns immediately whether data is available for reading, without waiting.
- Update call sites of `is_writable()`, removing redundant checks.

<hr />

**Note:** The new `is_readable()` capability is *essential* for the WebSockets code, as there's currently no way to tell if `read()` would block. I have an alternative proposal to add `has_buffered_data()`, which is identical to `is_readable()` in this PR. This alternative is less invasive and requires no other code changes, while this PR also aims to enhance semantic clarity.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -751,7 +751,8 @@ public:
   virtual ~Stream() = default;
 
   virtual bool is_readable() const = 0;
-  virtual bool is_writable() const = 0;
+  virtual bool wait_readable() const = 0;
+  virtual bool wait_writable() const = 0;
 
   virtual ssize_t read(char *ptr, size_t size) = 0;
   virtual ssize_t write(const char *ptr, size_t size) = 0;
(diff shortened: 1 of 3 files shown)


yhirose:
@falbrechtskirchinger thanks for the improvement! `is_readable()` is no longer used, right? If so, could you please remove it?

falbrechtskirchinger:
> @falbrechtskirchinger thanks for the improvement! `is_readable()` is no longer used, right? If so, could you please remove it?

Please read the note in the OP. `is_readable()` is critical for the PRs I'll submit later today. Once all three features are in, I'll add unit tests.

yhirose:
Sorry that I forgot to pay attention to the note... The code looks good. Thanks!

falbrechtskirchinger:
> Sorry that I forgot to pay attention to the note... The code looks good. Thanks!

No worries! :+1:

PR #2073, opened 2025-02-20
Add client stream handler

falbrechtskirchinger:
Introduce a new request callback, stream_handler. If set, it replaces all other content-serving mechanisms. The handler is not called in response to HEAD or CONNECT requests, or when following a redirect. Content-related default header fields are not added to the request.

In conjunction with a response handler, it provides a minimal interface to implement a WebSocket client.


yhirose:
I'll close it for now. Please let me know when you work on this.

PR #2077, opened 2025-02-20
Correct default thread pool size in README.md

falbrechtskirchinger:
https://github.com/yhirose/cpp-httplib/blob/a4b2c61a65f4ff3d18b3dd24d2abae40f30be7de/httplib.h#L129-L134


yhirose:
Thanks for the correction!

PR #2078, opened 2025-02-20
Remove `select()` and use `poll()`

falbrechtskirchinger:
As requested.


falbrechtskirchinger:
I could have renamed `select_*()` to `poll_*()` ... :man_shrugging:

yhirose:
You are right, but I am ok to stay with the name for now. 😆

PR #2079, opened 2025-02-20
Parallelize testing with/without SSL on Windows & set concurrency group

falbrechtskirchinger:
- Start two Windows runners: one with SSL, and one without.
- Set concurrency group on workflows such that pushes to the same branch/PR cancel currently running workflows.

This does apply to multiple PR merges to master in short succession, which could be changed if not desired.


yhirose:
Good improvement!

Issue #538, opened 2020-06-22
Enable redirect tests when httplib.org issue resolves


yhirose:
https://github.com/postmanlabs/httpbin/issues/617

PR #2089, opened 2025-03-05
Fix HttpWatch tests

falbrechtskirchinger:
HttpWatch changed the formatting of the returned JSON. Normalize it by removing all whitespace.

PR #2090, opened 2025-03-05
Switch redirect tests to `httpbingo.org`

falbrechtskirchinger:
Redirect tests fail using `httpbin.org` or `nghttp2.org/httpbin`. The location header value contains a string representation of a Python byte string (e.g., b'http://www.google.com/'), which results in a 404 error.

Note: `httpbingo.org` allows only a few destination URLs for the redirect, hence the change to `example.com`.

Related: #538


yhirose:
@falbrechtskirchinger thanks for taking care of this, too as well as #2089! It looks like some checks still fail. Did you figure out why?

falbrechtskirchinger:
@yhirose Checks are passing now. :smiley:

yhirose:
@falbrechtskirchinger thanks a lot for the great job!

PR #2092, opened 2025-03-06
Move `detail::read_file()` to `test/test.cc`

falbrechtskirchinger:
The unit test code is the only user of the function.

`read_file()` now throws an exception if the file isn't found. This improves usability for CMake users by providing clearer error messages when running the test from the wrong directory.


yhirose:
@falbrechtskirchinger thanks for catching it!

Issue #2084, opened 2025-03-01
Can't read content > 4GB with Content Receiver on 32bit platform


GEEKiDoS:
I'm using the client

GEEKiDoS:
seems like the size_t issue, modified some of variable types to uint64_t works fine
- `read_len` in `read_content_with_length` 
- `payload_max_length` on read_content
- `std::numeric_limits<size_t>` to std::numeric_limits<uint64_t> in detail::read_content call

falbrechtskirchinger:
Some other read functions already use `uint64_t`, FYI. Seems like an oversight.

yhirose:
@GEEKiDoS @falbrechtskirchinger thanks for the report and investigation! 

@GEEKiDoS Since I don't have a 32bit machine, I am not able to reproduce and confirm this issue at this point. Could you make a Unit test case in `test/test.cc` which can reproduce this issue, and send me a pull request including the test case? Meanwhile, I'll try to find an easy way to build a 32bit executable and run it on 64 bit Ubuntu on my 64 bit CPU machine. Thanks a lot!

falbrechtskirchinger:
> Meanwhile, I'll try to find an easy way to build a 32bit executable and run it on 64 bit Ubuntu on my 64 bit CPU machine. Thanks a lot!

That's as easy as adding `-m32`. I have some experience with this 32-bit and `size_t` issue from [nlohmann/json](https://github.com/nlohmann/json). I implemented the 32-bit unit tests in nlohmann/json#3523.

I did start work on a PR but quickly gave up after realizing what other problems this issue reveals:
- Only one compressor can handle >4GB on 32-bit without modification.
- File handling is severely limited without memory-mapped files (1GB on some implementations). Memory-mapped files are capped by the address space, which can be worked around with offsets.
- Some places check `std::string::max_size()` but not all. None set the proper status code.

It's more realistic to document the limitations of 32-bit systems first and then address some of the low-hanging fruit. Triggering all issues in unit tests may not be possible due to RAM limits.

If I find the time later, I'll work on documentation and some `Makefile` changes to allow local testing.

yhirose:
[...]

PR #2095, opened 2025-03-06
Run 32-bit compiled unit tests on Ubuntu

falbrechtskirchinger:
This PR lays the groundwork for future changes to address #2084.

- Add `EXTRA_CXXFLAGS` to the `Makefile` to inject `-m32` for compiling in 32-bit mode.
- Test 32-bit and 64-bit in CI on Ubuntu requiring installation of additional dependencies. (Dependencies are installed with the `:i386` architecture qualifier for 32-bit.)


yhirose:
Looks good to me!

PR #2102, opened 2025-03-12
Remove SSLInit

falbrechtskirchinger:
Quote: "As of version 1.1.0 OpenSSL will automatically allocate all resources that it needs so no explicit initialisation is required."


yhirose:
@falbrechtskirchinger I didn't know the fact. Thanks for the improvement!

Issue #2097, opened 2025-03-11
Crash when calling `std::exit` while server running or client requests in flight


Y--:
Hi!

Thanks for the great library!

I think there is a race condition when the process is terminated explicitly (for example with `std::exit`) while a server is processing queries and/or a client is executing some.

I crafted a small reproduction below (note that in the actual code we're not "sleeping" in the `atexit` but have the same issue nonetheless):

```c++
#include "httplib.h"

struct ServerWrapper {
  httplib::Server server;
  std::unique_ptr<std::thread> s_thread;
};

static std::unique_ptr<ServerWrapper> s_wrapper;

void mainServer() {
  s_wrapper->server.Get("/hi", [](const httplib::Request &req, httplib::Response &res) {
    res.set_content("Quack", "text/plain");
  });

  s_wrapper->server.listen("localhost", 8080);
}

void exitThread() {
  std::cout << "Will exit soon..." << std::endl;
  std::this_thread::sleep_for(std::chrono::milliseconds(500));
  std::cout << "Calling std::exit now." << std::endl;
  std::exit(0);
}

void runClient() {
  httplib::Client cli("localhost", 8080);
  while (true) {
    auto res = cli.Get("/hi");
    std::this_thread::sleep_for(std::chrono::milliseconds(1));
  }
}

void atex() {
  std::cout << "Entering std::atexit callback, will wait a bit and stop server" << std::endl;
  // Wait a bit before stopping server to simulate delayed exit
  std::this_thread::sleep_for(std::chrono::milliseconds(500));
  std::cout << "[std::atexit] Waited, will stop server now." << std::endl;
  s_wrapper->server.stop();
  std::cout << "[std::atexit] Server stopped." << std::endl;
  s_wrapper->s_thread->join();
  std::cout << "[std::atexit] Server thread completed." << std::endl;
}

int main() {
  std::atexit(atex);
  s_wrapper = std::make_unique<ServerWrapper>();
  s_wrapper->s_thread = std::make_unique<std::thread>(&mainServer);

  std::thread cliThread(&runClient);

  std::thread(&exitThread).join();
  return 0;
}
```

Can be compiled with:
```bash
$ g++ -g -O0 -std=c++20 test.cc
```

Then execution looks like:
```bash
$  ./a.out
Will exit soon...
Calling std::exit now.
Entering std::atexit callback, will wait a bit and stop server
[1]    37193 segmentation fault  ./a.out
```

[...]

PR #2103, opened 2025-03-12
Avoid static std::string

falbrechtskirchinger:
Replace static std::string objects with constexpr character arrays and use compile-time string length calculations.

Tangentially related to #2097.


yhirose:
Good improvement!

PR #2100, opened 2025-03-12
Update vendored gtest to 1.12.1

falbrechtskirchinger:
Update googletest to the last version supporting C++11.

This is required to compile with `-Wexit-time-destructors`.

Steps to verify:
- Download gtest v1.12.1 from here: https://github.com/google/googletest/releases/tag/release-1.12.1
- Unpack and `cd` into `googletest` directory.
- Run: `find src include -type f -exec md5sum {} \; >sums`
- Clone this PR: `git clone --depth 1 --branch update-gtest  https://github.com/falbrechtskirchinger/cpp-httplib.git`
- `cd` into the `test/gtest` directory.
- Run: `md5sum -c path/to/sums; echo $?`
- The exit code should be `0`.

PR #2105, opened 2025-03-12
Control exit-time destructors with new static variable macro

### LINKED ISSUE #2097: Crash when calling `std::exit` while server running or client requests in flight

falbrechtskirchinger:
Introduce the `CPPHTTPLIB_DEFINE_STATIC` macro to define static variables with optional dynamic allocation, enabled by defining `CPPHTTPLIB_NO_EXIT_TIME_DESTRUCTORS`. Dynamic allocation prevents running their destructors, which avoids race conditions with atexit handlers that may depend on them. Also includes a test to verify this functionality and updates the documentation to explain the usage and rationale for disabling exit-time destructors.

Note: gtest v1.12.1 does not include the fix for exit-time destructors (after finding the last C++11 version, I forgot to re-check that it includes the relevant commit), necessitating compiling gtest separately and without `-Wexit-time-destructors`.

Resolves #2097.

**To do:**
- [X] Add regression test for #2097.
- [X] Compile with `-Wexit-time-destructors`. (Requries: #2103 #2100)
- [X] Add documentation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ test/httplib.cc
 test/httplib.h
 test/test
 test/server_fuzzer
+test/test_no_exit_time_dtors
 test/test_proxy
 test/test_split
 test/test.xcodeproj/xcuser*
--- a/README.md
+++ b/README.md
@@ -963,6 +963,16 @@ Include `httplib.h` before `Windows.h` or include `Windows.h` by defining `WIN32
 > [!NOTE]
 > Windows 8 or lower, Visual Studio 2015 or lower, and Cygwin and MSYS2 including MinGW are neither supported nor tested.
 
+### Exit-time destructors
+
+By default, the library relies on exit-time destructors for the cleanup of its static objects when the program terminates. To disable these exit-time destructors, define the preprocessor macro `CPPHTTPLIB_NO_EXIT_TIME_DESTRUCTORS` before including `httplib.h`.
+
+> [!NOTE]
+> When exit-time destructors are disabled, all static variables are allocated on the heap and are not deleted, to prevent their destructors from being called at exit time. This results in purposeful memory leaks, but since the program is exiting, it typically does not affect the application's behavior.
+
+> [!NOTE]
+> If you use `std::atexit()` to register a function that accesses client or server objects from this library, it is recommended to disable exit-time destructors. This ensures that the objects remain valid when your registered function runs, avoiding potential issues with destructors being called before your function executes.
+
 License
 -------
 
--- a/httplib.h
+++ b/httplib.h
@@ -145,6 +145,16 @@
 #define CPPHTTPLIB_LISTEN_BACKLOG 5
 #endif
 
+#ifndef CPPHTTPLIB_DEFINE_STATIC
+#ifdef CPPHTTPLIB_NO_EXIT_TIME_DESTRUCTORS
+#define CPPHTTPLIB_DEFINE_STATIC(var_type, var, init)                          \
+  static var_type &var = *new typename std::remove_cv<var_type>::type init
+#else
+#define CPPHTTPLIB_DEFINE_STATIC(var_type, var, init)                          \
+  static var_type var = typename std::remove_cv<var_type>::type init
+#endif
+#endif
+
 /*
  * Headers
  */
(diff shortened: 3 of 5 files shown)


yhirose:
@falbrechtskirchinger thanks for the pull request. Could you please take a look at my https://github.com/yhirose/cpp-httplib/issues/2097#issuecomment-2722037034, and give me your thought?

falbrechtskirchinger:
Closing per our discussion in #2097.

PR #2107, opened 2025-03-12
Wrap poll()/WSAPoll() in a function and build compiled library on Windows

### LINKED ISSUE #2106: Mingw32 error (poll was not declared) + strange "poll"  def/undef in splitted header.
Here is #define poll. https://github.com/yhirose/cpp-httplib/blob/37399af9960eb0fef6f5270d60c14903648629d9/httplib.h#L197
Here is #undef poll. https://github.com/yhirose/cpp-httplib/blob/37399af9960eb0fef6f5270d60c14903648629d9/httplib.h#L10381

When splitted to httplib.cc and httplib.h files, "#define poll" and "#undef poll" remains in httplib.h file. 
"poll" itself not used in splitted httplib.h, but in httplib.cc  it comes after "#undef poll", if we have _WIN32 defined.

Seems that "#define poll" and "#undef poll" should be in "httplib.cc " when httplib.h is splitted into .h and .cc parts.

(I have no idea how or why it compiles in Windows.)

falbrechtskirchinger:
Instead of using a macro for `poll()` on Windows, which breaks when the implementation is compiled separately, add a `detail::poll_wrapper()` function that dispatches to either `::poll()` or `::WSAPoll()`.

Build compiled library on Windows to catch regressions.

Fixes #2106.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -98,9 +98,17 @@ jobs:
       matrix:
         config:
         - with_ssl: false
+          compiled: false
+          run_tests: true
           name: without SSL
         - with_ssl: true
+          compiled: false
+          run_tests: true
           name: with SSL
+        - with_ssl: false
+          compiled: true
+          run_tests: false
+          name: compiled
     name: windows ${{ matrix.config.name }}
     steps:
     - name: Prepare Git for Checkout on Windows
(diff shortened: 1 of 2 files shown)


yhirose:
@falbrechtskirchinger looks good to me!

PR #2108, opened 2025-03-12
Move poll define/undef into c++ file. to fix https://github.com/yhirose/cpp-httplib/issues/2106

Chuvi-w:
It is normal to be #defined  in namespace.


yhirose:
@Chuvi-w thanks! But it's already cared by #2107.

PR #2110, opened 2025-03-13
Make `random_string()` thread-safe

falbrechtskirchinger:
By making the random engine thread_local, each thread now has its own independent random sequence, ensuring safe concurrent access. Additionally, using an immediately invoked lambda expression to initialize the engine eliminates the need for separate static seed variables.


yhirose:
Thanks for the improvement!

PR #2112, opened 2025-03-17
cmake: only validate component when the required library is found

abouvier:
I blame @sum01 from https://github.com/yhirose/cpp-httplib/pull/2088 for this


yhirose:
@abouvier thanks for the quick response!

PR #2114, opened 2025-03-18
Fix start_time shadow variable

jfsimoneau:


yhirose:
@jfsimoneau thanks for the improvement!

PR #2115, opened 2025-03-20
Add AF_UNIX support on windows

p1-0tr:
Add support for using unix domain sockets on Windows.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -187,6 +187,7 @@ using ssize_t = long;
 #include <io.h>
 #include <winsock2.h>
 #include <ws2tcpip.h>
+#include <afunix.h>
 
 #ifndef WSA_FLAG_NO_HANDLE_INHERIT
 #define WSA_FLAG_NO_HANDLE_INHERIT 0x80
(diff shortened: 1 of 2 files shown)


yhirose:
@p1-0tr could you please add unit tests? Then, I'll start reviewing. Thanks!

p1-0tr:
@yhirose I enabled the existing test. Only the basic case, because windows only has support for basic features on `AF_UNIX` (no fd passing, or other anciliary data, autobind, `socketpair`, etc. - https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/#unsupported\unavailable: )

yhirose:
@p1-0tr I just reviewed the code, but the condition looks a bit complicated. Can it be simpler as shown below?

```cpp
      if (socket_options) { socket_options(sock); }

#ifdef _WIN32
      // Setting SO_REUSEADDR seems not to work well with AF_UNIX on windows, so
      // avoid setting default_socket_options.
      detail::set_socket_opt(sock, SOL_SOCKET, SO_REUSEADDR, 0);
#endif
```

p1-0tr:
@yhirose Yes, that works.

I went with a comparison on the settings callback as I thought it would be more in line with the principle of least surprise (from the user's point of view). But I am happy with the simpler approach :)

yhirose:
@p1-0tr thanks for the contribution!

yhirose:
@p1-0tr could you take a look at this error on GitHub Actions CI?

https://github.com/yhirose/cpp-httplib/actions/runs/14047884306/job/39343816422

p1-0tr:
@yhirose I created a fix for it - https://github.com/yhirose/cpp-httplib/pull/2118 . Unfortunately ordering of `winsock2.h` and `afunix.h` includes matters.

PR #2116, opened 2025-03-24
Default using Windows Schannel for SSL/TLS on Windows

### LINKED ISSUE #1978: Proposal: Default to Windows Automatic Root Certificates Update for Improved User Experience
Hi @yhirose,

Following up on the discussion in https://github.com/yhirose/cpp-httplib/issues/1431#issuecomment-2476060561, I wanted to propose enabling Windows Automatic Root Certificates Update as the default behavior for cpp-httplib on Windows. This feature would enhance user experience by leveraging the latest system root certificates automatically, reducing the need for users to manually manage certificates.

Currently, users can achieve this functionality by setting a custom verification callback, but making it the default on Windows would streamline the setup and make cpp-httplib more convenient out of the box. Additionally, this default change could allow us to skip preloading system root certificates during client construction, minimizing initialization overhead and potentially improving performance.

Would love to hear your thoughts on this proposal, and whether you see it aligning with the project’s direction.

Thank you!

solarispika:
Follow
https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-certgetcertificatechain for related flags.

Closes #1978

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -185,6 +185,9 @@ using ssize_t = long;
 #endif // NOMINMAX
 
 #include <io.h>
+#if defined(CPPHTTPLIB_OPENSSL_SUPPORT)
+#define CERT_CHAIN_PARA_HAS_EXTRA_FIELDS
+#endif
 #include <winsock2.h>
 #include <ws2tcpip.h>
 
@@ -5580,34 +5583,7 @@ inline bool is_ssl_peer_could_be_closed(SSL *ssl, socket_t sock) {
          SSL_get_error(ssl, 0) == SSL_ERROR_ZERO_RETURN;
 }
 
-#ifdef _WIN32
-// NOTE: This code came up with the following stackoverflow post:
-// https://stackoverflow.com/questions/9507184/can-openssl-on-windows-use-the-system-certificate-store
-inline bool load_system_certs_on_windows(X509_STORE *store) {
-  auto hStore = CertOpenSystemStoreW((HCRYPTPROV_LEGACY)NULL, L"ROOT");
-  if (!hStore) { return false; }
-
-  auto result = false;
-  PCCERT_CONTEXT pContext = NULL;
-  while ((pContext = CertEnumCertificatesInStore(hStore, pContext)) !=
-         nullptr) {
-    auto encoded_cert =
-        static_cast<const unsigned char *>(pContext->pbCertEncoded);
-
-    auto x509 = d2i_X509(NULL, &encoded_cert, pContext->cbCertEncoded);
-    if (x509) {
-      X509_STORE_add_cert(store, x509);
-      X509_free(x509);
-      result = true;
-    }
-  }
-
-  CertFreeCertificateContext(pContext);
-  CertCloseStore(hStore, 0);
-
-  return result;
-}
-#elif defined(CPPHTTPLIB_USE_CERTS_FROM_MACOSX_KEYCHAIN) && defined(__APPLE__)
+#if defined(CPPHTTPLIB_USE_CERTS_FROM_MACOSX_KEYCHAIN) && defined(__APPLE__)
 #if TARGET_OS_OSX
 template <typename T>
 using CFObjectPtr =
@@ -5697,7 +5673,7 @@ inline bool load_system_certs_on_macos(X509_STORE *store) {
   return result;
 }
 #endif // TARGET_OS_OSX
-#endif // _WIN32
+#endif // CPPHTTPLIB_USE_CERTS_FROM_MACOSX_KEYCHAIN && __APPLE__
 #endif // CPPHTTPLIB_OPENSSL_SUPPORT
 
 #ifdef _WIN32
@@ -9643,14 +9619,11 @@ inline bool SSLClient::load_certs() {
       }
     } else {
       auto loaded = false;
-#ifdef _WIN32
-      loaded =
-          detail::load_system_certs_on_windows(SSL_CTX_get_cert_store(ctx_));
-#elif defined(CPPHTTPLIB_USE_CERTS_FROM_MACOSX_KEYCHAIN) && defined(__APPLE__)
+#if defined(CPPHTTPLIB_USE_CERTS_FROM_MACOSX_KEYCHAIN) && defined(__APPLE__)
 #if TARGET_OS_OSX
       loaded = detail::load_system_certs_on_macos(SSL_CTX_get_cert_store(ctx_));
 #endif // TARGET_OS_OSX
-#endif // _WIN32
+#endif // CPPHTTPLIB_USE_CERTS_FROM_MACOSX_KEYCHAIN && __APPLE__
       if (!loaded) { SSL_CTX_set_default_verify_paths(ctx_); }
     }
   });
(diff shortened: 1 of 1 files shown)


solarispika:
I noticed that there is `verify_result_` for storing OpenSSL result and is retrieved by `get_openssl_verify_result()`, but I have no idea how to set it. Please advise.

yhirose:
@solarispika sorry for the delay. According to this comment https://github.com/yhirose/cpp-httplib/issues/1978#issuecomment-2703522574, you mentioned you ended up bypassing CRL in your production server.

Do you think that the current pull request which doesn't have the bypassing code will affect a number of Windows users? If not many, I don't mind merging this code. But it has a risk to affect many users, I would like you to implement a feature flag like `CPPHTTPLIB_USE_WINDOWS_AUTOMATIC_ROOT_CERTIFICATES_UPDATE`. If it's not set, the original code will be used.

solarispika:
Hi @yhirose 

I am not sure how many of them will be, possibly the number being proportional to users located in China.
I can add a toggle for this feature.
It looks like you prefer users to enable this feature, but not users to disable it, right?

yhirose:
@solarispika , (1) If a number of users will be affected by this, I prefer making it an opt-in feature with `CPPHTTPLIB_USE_WINDOWS_AUTOMATIC_ROOT_CERTIFICATES_UPDATE`.

(2) But if we expect only few users will be affected, we can enable this feature by the default and uses can disable it with `CPPHTTPLIB_DISABLE_WINDOWS_AUTOMATIC_ROOT_CERTIFICATES_UPDATE`.

I prefer #2.

solarispika:
@yhirose 
I agree with you on enabling the feature by default.
I also update CMake which enables the feature `HTTPLIB_USE_WINDOWS_AUTOMATIC_ROOT_CERTIFICATES_UPDATE` by default, and if disabled, define `CPPHTTPLIB_DISABLE_WINDOWS_AUTOMATIC_ROOT_CERTIFICATES_UPDATE`.

yhirose:
Could you please take a look at unit test errors on 'test / windows with SSL (pull_request)'?

solarispika:
Sure, it looks like https://github.com/yhirose/cpp-httplib/pull/2169 saves openssl errors which I didn't notice.
I'll try to fix it.

solarispika:
@yhirose 

I found that it is hard to map errors between Win32 API and OpenSSL.

What do you recommend? Is it proper to mask those checks when Schannel is used?

yhirose:
@solarispika I actually don't know what do to. Could you please investigate why those errors occur before making any change?

solarispika:
@yhirose
These errors are due to https://github.com/yhirose/cpp-httplib/pull/2169 adding `ssl_error` and `ssl_openssl_error` member functions to `class Result` for retrieving OpenSSL errors, and some tests being extended for these errors.

As those member functions are defined for OpenSSL, it is inappropriate for me to use it directly for errors coming from Windows API.

httplib.h
```diff
@@ -10529,13 +10537,17 @@ inline bool SSLClient::initialize_ssl(Socket &socket, Error &error) {
           }
 
           if (verification_status == SSLVerifierResponse::NoDecisionMade) {
+#if !defined(_WIN32) ||                                                        \
+    defined(CPPHTTPLIB_DISABLE_WINDOWS_AUTOMATIC_ROOT_CERTIFICATES_UPDATE)
```
yhirose, line 11456: @solarispika why does this check need to be skipped when `CPPHTTPLIB_DISABLE_WINDOWS_AUTOMATIC_ROOT_CERTIFICATES_UPDATE` is defined?

solarispika: I think this check and check from Windows are both checking certificate, so it is a either-or behavior.

So, if Windows automatic root certificate updates is disabled, then using old behavior; otherwise, use new one.

Am I missing something?

yhirose:
@solarispika it looks that the Windows code returns only `SSLServerVerification` and `SSLServerHostnameVerification` which are kind of general errors. However, it seems like some of wincrypt APIs can tell us more about errors, and they could be helpful for users. How about to make `wincrypt_verify_result_` and `get_wincrypt_verify_result()` to store such a wincrypt specific error? Is it too much?

solarispika:
> @solarispika it looks that the Windows code returns only `SSLServerVerification` and `SSLServerHostnameVerification` which are kind of general errors. However, it seems like some of wincrypt APIs can tell us more about errors, and they could be helpful for users. How about to make `wincrypt_verify_result_` and `get_wincrypt_verify_result()` to store such a wincrypt specific error? Is it too much?
> 

@yhirose I have no experience using those APIs, so I can't give any suggestions on it now.
It needs further investigations, and I think they might be useful.

BLesnau:
@solarispika @yhirose I have a project that I think would benefit from these changes. I may have some time to help out if there is anything else that is needed to get this merged. Is there anything specific that you have in mind? I did some testing with this change and updating "normal" certificates on my machine. I saw it use updated certificates automatically as expected. I did not specifically test with root certificates yet though, so I did not yet see that working, but it seems like it should.

I see that a couple of the checks failed, but the logs are now gone, so I can't tell what went wrong. I didn't see a way to manually re-run those actions. I was thinking of making a commit change to have the actions run again, but I didn't want to mess with this PR without warning or make my own fork and PR just to see the logs.

ponie419-max:
Thanks for any help someone will not stop attacking me and I think it's
time for there computer privileges to go away for life.

Me

BLesnau:
I was trying to test some of this myself, and I think I may not know enough about how to test it properly. I tried with self-signed certs. I tried reproducing the problem first, but the certificate was always available as long and I closed and reopened my app that uses `cpp-httplib` after I imported the cert. I tried importing it specifically as a root cert for `LocalMachine`, but it is always being propogated to both `LocalMachine` and `CurrentUser` no matter how I try to import the cert. Maybe I am going about the testing all wrong. I'm pretty new to this with the certs.

Any tips on how to test it would be great. Even if testing isn't straightforward, I'm still good with helping to get this PR moving closer to being merged.

ponie419-max:
No clue thanks for the help

MY DIC

solarispika:
@BLesnau The error come from ssl and openssl errors are not set correspondingly if I remember correctly.
@yhirose Per your suggestion, I think recording SChannel errors is great.
SChannel errors can be retrieved by `GetLastError()` and can be stored separately, `wincrypt_error_` and `wincrypt_chain_error_`.
SSL errors are cleared to 0 before starting SChannel's verification.
Users can check for Error::SSLServerVerification and use those error for debug.

BLesnau:
@solarispika @yhirose As part of looking at this PR, I was trying to understand more about how cpp-httplib does certain things (very new to me). Feel free to tell me to bring this discussion elsewhere if it's not related enough to this PR.

Without these Schannel changes, does cpp-httplib use the certs on Windows already or is it only the certs that are part of OpenSSL when it was compiled? Based on my testing, it does look like the Windows certs are retrieved when the client initializes. If cpp-httplib does use the Windows certs at init time, is there any specific code I should look at to understand it better?

Edit: I think I found the relevant code with the `load_system_certs_on_windows()` method.

PR #2118, opened 2025-03-25
Fix compilation error on windows

p1-0tr:
afunix.h uses types declared in winsock2.h, and so has to be included after it. Including afunix.h first will result in a somewhat unhelpful compilation error:

error C3646: 'sun_family': unknown override specifier

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -184,10 +184,12 @@ using ssize_t = long;
 #define NOMINMAX
 #endif // NOMINMAX
 
-#include <afunix.h>
 #include <io.h>
 #include <winsock2.h>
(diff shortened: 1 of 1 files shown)


yhirose:
@p1-0tr thanks for finding the problem! Could you please change the code as follows, so that clangformat can still sort include file entries in the block, and keep `afunix.h` out of the block. Thanks!

```cpp
#include <io.h>
#include <winsock2.h>
#include <ws2tcpip.h>

// `afunix.h` uses types declared in winsock2.h, so has to be included after it.
#include <afunix.h>
```

yhirose:
@p1-0tr thanks!

PR #2119, opened 2025-03-26
Speed up header splitting with smarter file handling

Silicon27:
- The script now reads and writes files in bigger chunks instead of line-by-line (like pouring a bucket instead of using an eyedropper)

- Uses a smarter way to find and remove inline keywords

- Handles file paths more reliably across different operating systems

- Still does exactly the same job, just quicker.

### DIFF
--- a/out/httplib.cc
+++ b/out/httplib.cc
@@ -1,0 +1,7807 @@
+#include "httplib.h"
+namespace httplib {
+
+/*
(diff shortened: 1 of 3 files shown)


yhirose:
@Silicon27 thanks for the pull request. However, I am probably not merging this, since the even current implementations is fast enough (60ms on my MacBook). (Also the output files, `out/httplib.h` and `out/httplib.cc` shouldn't be in the repository.)

PR #2127, opened 2025-04-09
feat: close client sock when server stop

tyronecai:
A problem found when hacking llama-server coredump (https://github.com/ggml-org/llama.cpp/pull/12831#issuecomment-2788759918)
just think these sockets should be released

thanks !

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -1115,10 +1115,18 @@ private:
                          MultipartContentHeader multipart_header,
                          ContentReceiver multipart_receiver) const;
 
+  void record_client_sock(socket_t sock);
+  
+  void remove_client_sock(socket_t sock);
+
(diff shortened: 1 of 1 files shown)


yhirose:
@tyronecai thanks for the pull request, but we shouldn't close the active client sockets there, because the in-flight requests in the thread pool should be complete and then they should be closed.

yhirose:
@tyronecai can you solve the problem that is happening on your side?

tyronecai:
you are right, thanks for reply.

> @tyronecai can you solve the problem that is happening on your side? You can use `Server::is_running()` method to check whether the server is still running or not.

yhirose:
Hope your project goes well!

PR #2132, opened 2025-04-19
Do not close socket in case of protocol upgrade

snakelizzard:
Http should not processing should stop at this point, socket will be reused for another protocol.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -7711,8 +7711,9 @@ inline bool ClientImpl::send_(Request &req, Response &res, Error &error) {
       socket_requests_are_from_thread_ = std::thread::id();
     }
 
-    if (socket_should_be_closed_when_request_is_done_ || close_connection ||
-        !ret) {
+    if ((socket_should_be_closed_when_request_is_done_ || close_connection ||
+         !ret) &&
+        res.status != StatusCode::SwitchingProtocol_101) {
       shutdown_ssl(socket_, true);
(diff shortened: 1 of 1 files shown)


yhirose:
@snakelizzard thanks for the pull request. Could you please add at least one unit test in `test/test.cc`? Then, I'll review the code. Thanks!

yhirose:
@falbrechtskirchinger, @PixlRainbow I am thinking of closing the pull request. Do you have any thoughts? Thanks!

PixlRainbow:
@snakelizzard I would expect that to fit within the "style" of httplib, the socket should be kept alive by blocking the thread from a content IO callback passed to httplib, instead of "handing off" the socket after the httplib thread exits; the protocol implementation is called from within that callback.
However, I am unsure of how @falbrechtskirchinger's planned switching protocol support is implemented.

snakelizzard:
Hello.

Sorry, I have not replied about unit tests from the beginning. This is related to @PixlRainbow note on style. I feel my solution is a bit hacky and does not follow the idea of the project. This is what I achieved by 20 min debugging when I realized httplib does not support protocol switch. I feel the value of my patch to publish a quick and dirty solution to whatever is agreed to take it on their own risk.

To make it right, I need better understand the code. And create unit tests. Unfortunately, I lack time to do all this. Please just close PR if it does not fit into @falbrechtskirchinger plans.

PR #2137, opened 2025-05-09
Default status codes available as an enum

ChiefyChief23:
I find it useful to have status codes available just to throw down with the industry standard meaning behind each for the developer to easily use them, so I thought I'd add them.

I got the following status codes from [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status)

I'd like to know what everyone thinks of the naming convention? I initially chose this because I'd like to lead with the number and then the meaning and the only nice way to do that I have found is with a leading `_`.

Feedback as always is welcome


ChiefyChief23:
Well I missed that one whoops

PR #2145, opened 2025-05-21
Detect if afunix.h exists

p1-0tr:
Older versions of the windows 10 SDK may not have `afunix.h`. So add logic to detect if the header exists, and enable `AF_UNIX` only on supported windows versions.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -192,8 +192,13 @@ using ssize_t = long;
 #include <winsock2.h>
 #include <ws2tcpip.h>
 
+#if defined(__has_include)
+#if __has_include(<afunix.h>)
 // afunix.h uses types declared in winsock2.h, so has to be included after it.
 #include <afunix.h>
+#define CPPHTTPLIB_HAVE_AFUNIX_H 1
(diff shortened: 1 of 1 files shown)


p1-0tr:
@yhirose I'm not sure why the `abidiff` step is failing. My change does not touch any function definitions, so it should not fail.

yhirose:
@p1-0tr you don't need to worry about the abidiff error.

httplib.h
```diff
@@ -192,8 +192,11 @@ using ssize_t = long;
 #include <winsock2.h>
 #include <ws2tcpip.h>
 
+#if __has_include(<afunix.h>)
```
yhirose, line 195: @p1-0tr this is not compatible with C++11.

p1-0tr: I know it only got standardised as of c++17, but has been supported by all major compilers before that. I've added an additional check that should protect against errors on compilers that do not support that macro.

yhirose:
@p1-0tr thanks for the adjustments!

PR #2151, opened 2025-06-02
Handle the return status of a non-blocking SSL handshake

tilongzs:
`set_nonblocking(sock, true);`
Set the socket to non-blocking mode. SSL_connect_or_accept(ssl) may return SSL_ERROR_WANT_READ or SSL_ERROR_WANT_WRITE, indicating that it needs to wait for more data or can write data. If the code does not check these error codes and retry, but instead directly considers a return value != 1 as a failure (as in the condition if (!setup(ssl) || SSL_connect_or_accept(ssl) != 1)), it will prematurely terminate the handshake, resulting in an SSL Connection Failed.

When using Discord's webhooks, I kept getting an SSL Connection Failed error. After debugging, I found that setting the socket to non-blocking mode was the cause. Of course, removing `set_nonblocking(sock, true)` and related code is the simplest and most effective solution, but this goes against the original intention of using a non-blocking mode socket.

The following is the test code for using Discord's webhooks. Please replace `webhook_url ` with the actual URL.

```
#include <iostream>
#include <string>
#include <nlohmann/json.hpp>

#include "httplib.h"

using namespace nlohmann;

void notify(const std::string& message)
{
	std::string webhook_url = "https://discordapp.com/api/webhooks/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx";

	httplib::Client client("https://discordapp.com");
	client.set_connection_timeout(0, 300000);
	client.set_read_timeout(5, 0);
	client.set_write_timeout(5, 0);

	std::string webhook_path = webhook_url.substr(webhook_url.find("/api/webhooks"));

	json payload = {
		{"content", message},
		{"username", "My Webhook"}, 
		{"avatar_url", ""}, 
		{"embeds", json::array({
			json{
				{"title", "Webhook Test"},
				{"description", "This is a test message sent via cpp-httplib."},
				{"color", 0x00ff00}
			}
		})}
	};

	httplib::Headers headers = {
		{"Content-Type", "application/json"}
	};

	auto res = client.Post(webhook_path,payload.dump(), "application/json");
	if (res) {
		if (res->status == 204) {
			std::cout << "Webhook message sent successfully!" << std::endl;
		}
		else if (res->status == 429) {
			json error = json::parse(res->body);
			std::cerr << "Rate limit hit, retry after: " << error["retry_after"].get<int>() << "ms" << std::endl;
		}
		else {
			std::cerr << "Failed to send webhook: Status " << res->status << ", Body: " << res->body << std::endl;
		}
	}
	else {
		std::cerr << "Request failed: " << httplib::to_string(res.error()) << std::endl;
		std::cin.get();
	}
}

int main() {
	notify("Hello, this is a test message from cpp-httplib!");

	return 0;
}
```

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -1,4 +1,4 @@
-//
+﻿//
 //  httplib.h
 //
 //  Copyright (c) 2025 Yuji Hirose. All rights reserved.
@@ -9158,7 +9158,7 @@ inline SSL *ssl_new(socket_t sock, SSL_CTX *ctx, std::mutex &ctx_mutex,
     BIO_set_nbio(bio, 1);
     SSL_set_bio(ssl, bio, bio);
 
-    if (!setup(ssl) || SSL_connect_or_accept(ssl) != 1) {
+    if (!setup(ssl)) {
       SSL_shutdown(ssl);
       {
         std::lock_guard<std::mutex> guard(ctx_mutex);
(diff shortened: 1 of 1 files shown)


yhirose:
@tilongzs thanks for the pull request. Could you please provide some unit test cases in `test/test.cc`? Then, I'll take a look at it. Thanks!

tilongzs:
I don't know how to write unit test code, so forget it, I'll just use it myself.

yhirose:
@tilongzs I tried to reproduce this problem on my machine, but the following code just worked on my machine. I don't know why you didn't get success with SSL_connect_or_accept on your machine... Your network setting could be wrong?

```cpp
#include "httplib.h"
#include <iostream>

int main() {
    auto domain = "https://discord.com";
    auto path = "/api/webhooks/xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx";

    auto body = R"({
      "content": "Hello, this is a test message from cpp-httplib sent via webhook!"
    })";

    httplib::Client cli(domain);
    auto res = cli.Post(path, body, "application/json");

    if (res) {
        std::cout << "Status: " << res->status << std::endl;
    } else {
        std::cerr << "Failed to send request: " << httplib::to_string(res.error()) << std::endl;
    }

    return 0;
}
```

```bash
> ./send
Status: 204
```

tilongzs:
@yhirose I found the root of our disagreement. You're right, your code indeed doesn’t return `SSL_connect_or_accept`. After my debugging, I discovered the issue lies in my code’s `set_connection_timeout(0, 300000)`. When this setting is increased, it return `SSL_connect_or_accept`.

```
	httplib::Client cli(domain);
	cli.set_connection_timeout(0, 300000);
//	cli.set_connection_timeout(0, 100);
```

Additionally, I found that when using` cli.set_connection_timeout(0, 100)`, it returns `Connection timed out`, which aligns with the literal meaning of `set_connection_timeout`. However, why does `cli.set_connection_timeout(0, 300000)` return `SSL connection failed`?

After applying the changes from my initially submitted "httplib.h," using` cli.set_connection_timeout(0, 300000)` correctly returns `Status: 204`instead of `SSL connection failed.` When using` cli.set_connection_timeout(0, 100)`, it still returns `Connection timed out` because it indeed timed out.

yhirose:
@tilongzs if you got the `SSL connection failed` error, it means that you didn't give enough time to `cli.set_connection_timeout`. In this case, the TCP connection actually succeeded, but `SSL_connet` failed due to the timeout. Here is the place you got the timeout. Hope it helps!

https://github.com/yhirose/cpp-httplib/blob/365cbe37fac46726c1a9a89b1f7e5a06431d0d1d/httplib.h#L9207

tilongzs:
@yhirose So, if the TCP connection has already been successfully established, there should not be an `SSL connection failed` error due to insufficient time.
The code I pulled precisely addresses this issue. Although the likelihood of this scenario occurring is low, I still hope you can reconsider. Unless you add a `set_ssl_connection_timeout(time_t sec, time_t usec) `interface.

PR #2152, opened 2025-06-06
Feature/multipart headers

kwach:
* feature: additional field: header - providing access to headers per form-data part
* UT test provided
```
Note: Google Test filter = MultipartFormDataTest.AccessPartHeaders
[==========] Running 1 test from 1 test suite.
[----------] 1 test from MultipartFormDataTest
[ RUN      ] MultipartFormDataTest.AccessPartHeaders
[       OK ] MultipartFormDataTest.AccessPartHeaders (1004 ms)
[----------] 1 test from MultipartFormDataTest (1004 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (1004 ms total)
[  PASSED  ] 1 test.
```

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -5048,10 +5048,10 @@ public:
 
           // parse and emplace space trimmed headers into a map
           if (!parse_header(
-                 header.data(), header.data() + header.size(),
-                 [&](const std::string &key, const std::string &val) {
-                   file_.headers.emplace(key, val);
-                 })) {
+                  header.data(), header.data() + header.size(),
+                  [&](const std::string &key, const std::string &val) {
+                    file_.headers.emplace(key, val);
+                  })) {
             is_valid_ = false;
             return false;
           }
--- a/httplib.h
+++ b/httplib.h
@@ -5046,12 +5046,14 @@ public:
             return false;
           }
 
-          // split header string by ':' and emplace space trimmed into headers map
-          auto colon_pos = header.find(':');
-          if (colon_pos != std::string::npos) {
-            auto key = trim_copy(header.substr(0, colon_pos));
-            auto val = trim_copy(header.substr(colon_pos + 1));
-            file_.headers.emplace(key, val);
+          // parse and emplace space trimmed headers into a map
+          if (!parse_header(
+                 header.data(), header.data() + header.size(),
+                 [&](const std::string &key, const std::string &val) {
+                   file_.headers.emplace(key, val);
+                 })) {
+            is_valid_ = false;
+            return false;
           }
 
           constexpr const char header_content_type[] = "Content-Type:";
--- a/httplib.h
+++ b/httplib.h
@@ -544,6 +544,7 @@ struct MultipartFormData {
   std::string content;
   std::string filename;
   std::string content_type;
+  Headers headers;
 };
 using MultipartFormDataItems = std::vector<MultipartFormData>;
 using MultipartFormDataMap = std::multimap<std::string, MultipartFormData>;
@@ -5045,6 +5046,14 @@ public:
             return false;
           }
 
+          // split header string by ':' and emplace space trimmed into headers map
+          auto colon_pos = header.find(':');
+          if (colon_pos != std::string::npos) {
+            auto key = trim_copy(header.substr(0, colon_pos));
+            auto val = trim_copy(header.substr(colon_pos + 1));
+            file_.headers.emplace(key, val);
+          }
+
           constexpr const char header_content_type[] = "Content-Type:";
 
           if (start_with_case_ignore(header, header_content_type)) {
@@ -5144,6 +5153,7 @@ private:
     file_.name.clear();
     file_.filename.clear();
     file_.content_type.clear();
+    file_.headers.clear();
   }
 
   bool start_with_case_ignore(const std::string &a, const char *b) const {
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -29,6 +29,7 @@ find_package(CURL REQUIRED)
 add_executable(httplib-test test.cc include_httplib.cc $<$<BOOL:${WIN32}>:include_windows_h.cc>)
 target_compile_options(httplib-test PRIVATE "$<$<CXX_COMPILER_ID:MSVC>:/utf-8;/bigobj>")
 target_link_libraries(httplib-test PRIVATE httplib GTest::gtest_main CURL::libcurl)
+include(GoogleTest)
 gtest_discover_tests(httplib-test)
 
 file(
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -29,7 +29,6 @@ find_package(CURL REQUIRED)
 add_executable(httplib-test test.cc include_httplib.cc $<$<BOOL:${WIN32}>:include_windows_h.cc>)
 target_compile_options(httplib-test PRIVATE "$<$<CXX_COMPILER_ID:MSVC>:/utf-8;/bigobj>")
 target_link_libraries(httplib-test PRIVATE httplib GTest::gtest_main CURL::libcurl)
-include(GoogleTest)
 gtest_discover_tests(httplib-test)
 
 file(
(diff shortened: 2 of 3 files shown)


test/CMakeLists.txt
```diff
@@ -29,6 +29,7 @@ find_package(CURL REQUIRED)
 add_executable(httplib-test test.cc include_httplib.cc $<$<BOOL:${WIN32}>:include_windows_h.cc>)
 target_compile_options(httplib-test PRIVATE "$<$<CXX_COMPILER_ID:MSVC>:/utf-8;/bigobj>")
 target_link_libraries(httplib-test PRIVATE httplib GTest::gtest_main CURL::libcurl)
+include(GoogleTest)
```
yhirose: @sum01 @jimmy-park is this change necessary?

httplib.h
```diff
@@ -5045,6 +5046,14 @@ class MultipartFormDataParser {
             return false;
           }
 
+          // split header string by ':' and emplace space trimmed into headers map
```
yhirose: You don't have to parse the header text yourself. You can replace the code from line no 5043 through 5047 with the following code, so that you can avoid the extra parsing.

```cpp
          if (!parse_header(
                  header.data(), header.data() + header.size(),
                  [&](const std::string &key, const std::string &val) {
                    file_.headers.emplace(key, val);
                  })) {
            is_valid_ = false;
            return false;
          }
```

test/CMakeLists.txt
jimmy-park: This line is redundant because the GoogleTest module is already included in `find_package(GTest)`.

https://github.com/Kitware/CMake/blob/3bca7bca7b716ad0a618071d8052fbf2d9e4ad21/Modules/FindGTest.cmake#L242

yhirose: @kwach could you please remove the line?

kwach: No problem, however for me it generated an error: ```CMake Error at test/CMakeLists.txt:33 (gtest_discover_tests):
  Unknown CMake command "gtest_discover_tests".```

httplib.h
kwach: Ok!

yhirose:
@kwach could you please run clang-format? (You don't have to worry about the 'abidiff/abi' error.
https://github.com/yhirose/cpp-httplib/actions/runs/15518234423/job/43688499197?pr=2152

kwach:
> @kwach could you please run clang-format? (You don't have to worry about the 'abidiff/abi' error. https://github.com/yhirose/cpp-httplib/actions/runs/15518234423/job/43688499197?pr=2152

heh, ok. I've enabled workflows on my fork and now it passes. I'm just learnig about workflows, sorry :D

test/test.cc
```diff
@@ -8010,6 +8010,68 @@ TEST(MultipartFormDataTest, ContentLength) {
...
+    ASSERT_EQ("text2", it->second.content);
+    auto &headers = it->second.headers;
+    ASSERT_EQ(3, headers.size());
+    auto customHeader = headers.find("x-whatever");
+    ASSERT_TRUE(customHeader != headers.end());
+    ASSERT_NE("customvalue", customHeader->second);
+    ASSERT_EQ("CustomValue", customHeader->second);
+    ASSERT_TRUE(headers.find("X-Test") == headers.end()); // text1 header
+
+    handled = true;
```
yhirose, line 8040: Could you explain why `handled` is needed?

test/test.cc
```diff
@@ -8010,6 +8010,68 @@ TEST(MultipartFormDataTest, ContentLength) {
...
+
+  Server svr;
+  svr.Post("/test", [&](const Request &req, Response &) {
+    ASSERT_EQ(2u, req.files.size());
+
+    auto it = req.files.begin();
+    ASSERT_EQ("text1", it->second.name);
+    ASSERT_EQ("text1", it->second.content);
+    ASSERT_EQ(1, it->second.headers.count("Content-Length"));
+    auto content_lenght = it->second.headers.find("CONTENT-length");
```
yhirose: Is `lenght` misspelled?

test/test.cc
```diff
@@ -8010,6 +8010,68 @@ TEST(MultipartFormDataTest, ContentLength) {
...
+    ASSERT_EQ("5", content_lenght->second);
+    ASSERT_EQ(3, it->second.headers.size());
+
+    ++it;
+    ASSERT_EQ("text2", it->second.name);
+    ASSERT_EQ("text2", it->second.content);
+    auto &headers = it->second.headers;
+    ASSERT_EQ(3, headers.size());
+    auto customHeader = headers.find("x-whatever");
```
yhirose: Could you rename `customHeader` to be `custom_header`, since we use the 'snake_case' instead of 'camelCase' for local variables.

test/test.cc
kwach: It's testing case insensitive compare

kwach: oh, there! typo... yup

test/test.cc
kwach: I reused code from the MultipartFormDataTest

kwach: then there is     
```
ASSERT_TRUE(handled);
``` 
in 
```  
auto se = detail::scope_exit([&] {...}
```

kwach:
```
[==========] 365 tests from 88 test suites ran. (272323 ms total)
[  PASSED  ] 365 tests.

  YOU HAVE 2 DISABLED TESTS
```

yhirose:

yhirose:
@kwach after this pr has been merged, we now get lots of warnings when compiling `test/test.cc`.

```
test.cc:3939:39: warning: missing field 'headers' initializer [-Wmissing-field-initializers]
 3939 |       {"text1", "text default", "", ""},
      |                                       ^
 ```
 
 The current code looks as below.
 
 ```cpp
 TEST_F(ServerTest, MultipartFormData) {
  MultipartFormDataItems items = {
      {"text1", "text default", "", ""},
      {"text2", "aωb", "", ""},
      {"file1", "h\ne\n\nl\nl\no\n", "hello.txt", "text/plain"},
      {"file2", "{\n  \"world\", true\n}\n", "world.json", "application/json"},
      {"file3", "", "", "application/octet-stream"},
      {"file4", "", "", "   application/json  tmp-string    "}};

  auto res = cli_.Post("/multipart", items);

  ASSERT_TRUE(res);
  EXPECT_EQ(StatusCode::OK_200, res->status);
}
```

In order to fix these warnings, we have to add `{}`...
 
 ```cpp
 TEST_F(ServerTest, MultipartFormData) {
  MultipartFormDataItems items = {
      {"text1", "text default", "", "", {}},
      {"text2", "aωb", "", "", {}},
      {"file1", "h\ne\n\nl\nl\no\n", "hello.txt", "text/plain", {}},
      {"file2", "{\n  \"world\", true\n}\n", "world.json", "application/json", {}},
      {"file3", "", "", "application/octet-stream", {}},
      {"file4", "", "", "   application/json  tmp-string    ", {}}};

  auto res = cli_.Post("/multipart", items);

  ASSERT_TRUE(res);
  EXPECT_EQ(StatusCode::OK_200, res->status);
}
```

The `headers` fields is useful on the server side, but it looks weird on the client side. It's because the client isn't interested in the `headers` field at all, but a call needs to set `{}` to the MultipartFormData object when we use the initializer syntax.

We may need to add a comment like this:

```cpp
struct MultipartFormData {
  std::string name;
  std::string content;
  std::string filename;
  std::string content_type;
  
  // for server use only
  Headers headers;
};
```

I'll think more about how to deal with this issue.

kwach:
Well, do you want me to add the `{}` to silence the warnings? 
As for server only use comment: perhaps not. I think that's ok. Just like in my case. Someone could pass more headers as a client. Then, that would be simple to pass nonempty object there in initialization. Perhaps thers is some coding to do? Wouldn't be fast, but I could do a PR

yhirose:
I have already fixed it at https://github.com/yhirose/cpp-httplib/commit/aabd0634aeac096a422e5c21a0b62b5e5f7590ef.
With this fix, users won’t need to make any changes to their codebase at all. :)

kwach:
Ok, sorry it got a bit messy :-/ And thanks 🙏On Jun 30, 2025 15:11, yhirose ***@***.***> wrote:yhirose left a comment (yhirose/cpp-httplib#2152)
I have already fixed it at aabd063.

PR #2159, opened 2025-06-23
Fix bad request for multipart form data with boundary split

herbrechtsmeier:
Fix a bad request (400) response for multipart form data requests with a large header which leads to a split inside the boundary because of the read buffer size inside the SocketStream class. The parse function inside the MultipartFormDataParser wrongly erase the receive buffer if it doesn't find the boundary inside the buffer during first call.


yhirose:
@herbrechtsmeier could you please resolve build errors, first? :)

herbrechtsmeier:
Sorry, I didn't realize I was using a private function because CMake/gcc built and ran the tests.

yhirose:
@herbrechtsmeier thanks for your fine contribution!

PR #2166, opened 2025-06-28
Cleaner API

yhirose:


Jithesh-j:
**Review of the Diff:** 
Backward Compatibility: Existing code using Post, Get, Put, etc., will still compile without modification.
Improved Granularity: The integration into MultipartFormData and ContentProvider ensures that even complex, streamed uploads can now be monitored.

**Review of the PR Description:**
The "Cleaner" Aspect: Previously, a developer might have had to wrap a ContentProvider just to count bytes. Now, they simply pass a lambda.
Standardization: Using uint64_t for both current and total ensures the API is future-proof for files larger than 4GB, which is critical for modern high-definition media or large datasets.

**Impact to Test Cases:**
State Verification: The tests explicitly check EXPECT_TRUE(progress_called) and verify that progress_values is not empty, ensuring the callbacks are actually triggered by the internal loop.

PR #2168, opened 2025-07-05
Add non-blocking getaddrinfo option to Cmake

sum01:
Adds Cmake option `HTTPLIB_USE_NON_BLOCKING_GETADDRINFO` default on. Also adds the `HTTPLIB_IS_USING_NON_BLOCKING_GETADDRINFO` post-build var

Ref #1601, #2167, and https://github.com/yhirose/cpp-httplib/issues/1601#issuecomment-3021357070


yhirose:
@sum01 Thanks a lot!

PR #2174, opened 2025-07-07
build(meson): add non_blocking_getaddrinfo option

Tachi107:
This new option automatically enables the new non-blocking name resolution when the appropriate libraries are found, automatically adding them to the list of required dependencies. It will gracefully fall back to the old behaviour when no library is found.

This complements commit ea850cbfa74e2dff228c49bf94542ce5331d73b5.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -16,11 +16,12 @@ project(
   meson_version: '>=0.62.0'
 )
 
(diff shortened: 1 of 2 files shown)


yhirose:
Thanks a lot!

PR #2176, opened 2025-07-08
build(meson): simplify build options

Tachi107:
The "cpp-httplib_" prefix of build options is now dropped, as Meson build options are already namespaced for each project. The old names remain as deprecated aliases for the new ones.


yhirose:
@Tachi107 Thanks!

PR #2181, opened 2025-07-13
build(meson): warn/fail on 32-bit machines

bgilbert:
On 32-bit Windows, `meson setup` fails with an unclear error:

    meson.build:25:16: ERROR: Could not get define 'CPPHTTPLIB_VERSION'

The actual problem is that `httplib.h` `#error`s out.

Have the Meson logic explicitly check for a 32-bit host and warn or error, matching the check in `httplib.h`.  Phrase the Windows error in a way that triggers WrapDB CI's unsupported architecture check.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -18,6 +18,14 @@ project(
 
 cxx = meson.get_compiler('cpp')
 
+if cxx.sizeof('void *') != 8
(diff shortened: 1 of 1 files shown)


yhirose:
@Tachi107 could you please review this pr? Thanks!

Issue #2082, opened 2025-02-25
Add helpers to URL-encode user input with GET requests


ungive:
I noticed that this library offers no way to URL encode certain characters in the URL path, like `&`. My request path is `/v2/searchresults/Piri Tommy Villiers - on & on/relationships/tracks` which contains an ampersand, yet a client's `Get` method does not automatically encode it (which it can't anyway, since the parameter to `Get` includes query parameters as well), nor is there a public function/method to URL encode part of the path, when there is data that may contain characters like these (e.g. when the path contains user input).

I need to include another library or manually write a method to URL encode my user input here: `Piri Tommy Villiers - on & on`.

The path above is used with the [TIDAL API](https://developer.tidal.com/apiref?spec=search-v2&ref=get-searchresults-relationship-tracks-v2&at=THIRD_PARTY) and I get 0 API results when I don't encode the ampersand:

```cpp
for (auto ampersand : { "%26", "&" }) {
    std::cerr << ampersand << std::endl;
    httplib::Client cli("https://openapi.tidal.com");
    cli.set_bearer_token_auth("...");
    std::string path = "/v2/searchresults/Piri Tommy Villiers - on " + std::string(ampersand) + " on/relationships/tracks";
    auto result = cli.Get(httplib::append_query_params(path,
        {
            { "include", "tracks,tracks.artists,tracks.albums" },
            { "countryCode", "DE" },
        }));
    ASSERT_EQ(httplib::Error::Success, result.error());
    auto body = nlohmann::json::parse(result->body);
    EXPECT_GT(body["data"].size(), 0);
}
```

This fails with `&` for obvious reasons.

IMO this library should offer a public `encode_url_component` method of some kind or a method to sanitize user input for use in the request path (or a query parameter), such that it works well with `httplib::append_query_params` and `httplib::Client::Get`. I can't naively replace `&` with `%26` and `?` with `%3F` because `httplib::detail::encode_url` does not actually encode all characters it should be encoding, if I put a tabstop in the URL (`\t`) then my request fails as well (it's not replaced with `%09`):

https://github.com/yhirose/cpp-httplib/blob/71ba7e7b1b328fe0de6cfbd3e94e5e0ddd4b4073/httplib.h#L2850-L2880

ungive:
[...]

PR #2184, opened 2025-07-19
Prevent re-encoding of path when following a redirect

JoeMylinn:
The current implementation re-encodes the path and query when following a redirect, unless `path_encode_` is set to `false`.
However, the query is never decoded. This causes issues if the query contains a space `' '` (encoded as `'+'`). Since it's never decoded, but later re-encoded, the space `' '` (encoded as `'+'`) gets changed into a `'+'` (encoded as `'%2B'`). Thus, the value of the parameter gets changed.

This fixes it by disabling path encoding on the redirect client.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -8636,7 +8636,7 @@ inline bool ClientImpl::redirect(Request &req, Response &res, Error &error) {
   if (location.empty()) { return false; }
 
   thread_local const std::regex re(
-      R"((?:(https?):)?(?://(?:\[([a-fA-F\d:]+)\]|([^:/?#]+))(?::(\d+))?)?([^?#]*)(\?[^#]*)?(?:#.*)?)");
+      R"((?:(https?):)?(?://(?:\[([a-fA-F\d:]+)\]|([^:/?#]+))(?::(\d+))?)?([^?#]*)(?:\?([^#]*))?(?:#.*)?)");
 
   std::smatch m;
   if (!std::regex_match(location, m, re)) { return false; }
@@ -8661,18 +8661,16 @@ inline bool ClientImpl::redirect(Request &req, Response &res, Error &error) {
   if (next_host.empty()) { next_host = host_; }
   if (next_path.empty()) { next_path = "/"; }
 
-  auto path = next_path + next_query;
+  if (!next_query.empty()) { detail::parse_query_text(next_query, req.params); }
 
   // Same host redirect - use current client
   if (next_scheme == scheme && next_host == host_ && next_port == port_) {
-    this->set_path_encode(
-        false); // Disable path encoding, redirects should already be encoded
-    return detail::redirect(*this, req, res, path, location, error);
+    return detail::redirect(*this, req, res, next_path, location, error);
   }
 
   // Cross-host/scheme redirect - create new client with robust setup
   return create_redirect_client(next_scheme, next_host, next_port, req, res,
-                                path, location, error);
+                                next_path, location, error);
 }
 
 // New method for robust redirect client creation
@@ -8758,8 +8756,7 @@ inline void ClientImpl::setup_redirect_client(ClientType &client) {
   client.set_keep_alive(keep_alive_);
   client.set_follow_location(
       true); // Enable redirects to handle multi-step redirects
-  client.set_path_encode(
-      false); // Disable encoding, redirects should already be encoded
+  client.set_path_encode(path_encode_);
   client.set_compress(compress_);
   client.set_decompress(decompress_);
 
--- a/httplib.h
+++ b/httplib.h
@@ -8661,10 +8661,11 @@ inline bool ClientImpl::redirect(Request &req, Response &res, Error &error) {
   if (next_host.empty()) { next_host = host_; }
   if (next_path.empty()) { next_path = "/"; }
 
-  auto path = detail::decode_path(next_path, true) + next_query;
+  auto path = next_path + next_query;
 
   // Same host redirect - use current client
   if (next_scheme == scheme && next_host == host_ && next_port == port_) {
+    this->set_path_encode(false); //Disable path encoding, redirects should already be encoded
     return detail::redirect(*this, req, res, path, location, error);
   }
 
@@ -8756,7 +8757,7 @@ inline void ClientImpl::setup_redirect_client(ClientType &client) {
   client.set_keep_alive(keep_alive_);
   client.set_follow_location(
       true); // Enable redirects to handle multi-step redirects
-  client.set_path_encode(path_encode_);
+  client.set_path_encode(false); //Disable encoding, redirects should already be encoded
   client.set_compress(compress_);
   client.set_decompress(decompress_);
 
--- a/httplib.h
+++ b/httplib.h
@@ -8665,7 +8665,8 @@ inline bool ClientImpl::redirect(Request &req, Response &res, Error &error) {
 
   // Same host redirect - use current client
   if (next_scheme == scheme && next_host == host_ && next_port == port_) {
-    this->set_path_encode(false); //Disable path encoding, redirects should already be encoded
+    this->set_path_encode(
+        false); //Disable path encoding, redirects should already be encoded
     return detail::redirect(*this, req, res, path, location, error);
   }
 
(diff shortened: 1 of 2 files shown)


yhirose:
@JoeMylinn thanks for the pull request. Could you please add enough unit tests for this in `test/test.cc`? Then, I'll take a look at it. Thanks!

JoeMylinn:
@yhirose I've added a test for this, let me know if it's enough.

yhirose:
@ungive, do you have any thought of this pull request? Thanks!
https://github.com/yhirose/cpp-httplib/issues/2082

yhirose:
@JoeMylinn could you please fix the style formatting error?

JoeMylinn:
@yhirose done

yhirose:
'test/style-check' still gives errors...

JoeMylinn:
my bad, shouldn't give any errors now

httplib.h
```diff
@@ -8661,10 +8661,12 @@ inline bool ClientImpl::redirect(Request &req, Response &res, Error &error) {
   if (next_host.empty()) { next_host = host_; }
   if (next_path.empty()) { next_path = "/"; }
 
-  auto path = detail::decode_path(next_path, true) + next_query;
+  auto path = next_path + next_query;
 
   // Same host redirect - use current client
   if (next_scheme == scheme && next_host == host_ && next_port == port_) {
+    this->set_path_encode(
+        false); // Disable path encoding, redirects should already be encoded
```
ungive: Doesn't this set `path_encode_` for all subsequent requests, overriding what the user configured?

```cpp
Client cli(HOST, PORT);
cli.set_follow_location(true);
cli.set_path_encode(true); // encode paths
auto res = cli.Get("/a b"); // encoded as "/a%20b", redirects
// path_encode_ is now false
auto res = cli.Get("/c d"); // not encoded anymore, sent as "/c d"
```

Just a thought, I haven't tested this code.

JoeMylinn: Yeah I believe you are right. That shouldn't be the case of course. Maybe we can just always create a new redirect client, even if it's the same host and never reuse the original one?

DeltaRaidcore: Or you go back to your original suggestion (we talked about on discord) and decode the host+query so that it can be re-encoded in the client as normal. @JoeMylinn

JoeMylinn: Yeah maybe the best option is to always decode both the path and query, and re-encode it before passing the new path to the original or redirect client.

JoeMylinn: I still don't like the decode and re-encode thing. Especially since we'd always have to encode if we decode it first, or it could break URLs with encoded characters if `path_encode_` is set to `false`.

We could just parse the query parameters and add them to the request, something like this:

```cpp
 auto path = next_path;

 if (!next_query.empty())
 {
   detail::parse_query_text(next_query, req.params);
 }
```

@yhirose any thoughts on what solution you'd prefer?

yhirose: @JoeMylinn I still don't fully understand the whole situation, yet. Could you please try with v0.22.0 to see if the same issue happens?

JoeMylinn: @yhirose Yes, v0.22.0 has the same issue.

I'll try to explain the issue a bit better. We need several things for it to happen:
1. `follow_location_` set to `true`
2. `path_encode_` set to `true`
3. Get redirected, with the query part of the Location containing a `+`.

For example, let's say the redirect location is `https://example.com/test?say=Hello+World`. The value of the parameter `say` *should* be `Hello World`, since `'+'` encodes space `' '` in a query. However, the current implementation just appends the query to the path, and then (if `path_encode_` is set to `true`) encodes the whole path, including the query. This encodes it as `Hello%2BWorld`, instead of `Hello+World` or `Hello%20World`, changing the value of the of the parameter to `Hello+World`.

I've noticed this issue because it causes downloads from GitHub to fail, same as in #2185.

I originally fixed this by just setting `path_encode_` to `false` when following a redirect.
However this is flawed, since it might override the user configured settings for the client, when the location is on the same host.

I think there are mainly two ways to fix this:
1. Always use  a new redirect client, even if the location is on the same host, with `path_encode_` set to `false`.
2. Parse the query properly, and pass the parameters to the new request.

DeltaRaidcore: The question is why is only the host decoded and not the entire query?

JoeMylinn: I am not sure, but decoding just the path can already cause issues even if there isn't a query. If `path_encode_` is `false`, and the path contains an encoded character, it will get decoded but not encoded again and the request will fail. So we'd have to always encode the path, regardless of the settings on the original client.

I just feel like it's simpler to never decode the path in the first place, and just parse the query. Unless there is a reason for it that I am missing.

yhirose:
I am still trying to understand the real cause of this problem and learn how other HTTP libraries handle this issue. Thanks for your patience.

Here is my investigation with Claude Code:
```
⏺ How Other HTTP Libraries Handle This Problem

  Based on research findings, here's how major HTTP libraries address this issue:

  1. libcurl Solution Approach

  Design Characteristics:
  - CURLU_URLENCODE flag for explicit control
  - Space→+ conversion executed at set time for query parts
  - At retrieval time, CURLU_URLDECODE performs +→space conversion

  Redirect Handling:
  Set time: "Hello World" → "Hello+World" (CURLU_URLENCODE)
  Redirect time: Preserves encoding state
  Get time: "Hello+World" → "Hello World" (CURLU_URLDECODE)

  2. Python requests Solution Approach

  Design Characteristics:
  - Automatic encoding via params argument
  - Space→+ conversion with urllib.parse.quote_plus()
  - Preserves encoding state during redirects

  Implementation Example:
  # Automatic handling (recommended)
  params = {'query': 'Hello World'}
  requests.get(url, params=params)  # Automatically encodes properly

  # Manual control
  from urllib.parse import quote_plus
  encoded = quote_plus("Hello World")  # "Hello+World"

  3. Go net/http Solution Approach

  Design Characteristics:
  - Context-specific encoding
    - url.QueryEscape(): For queries (preserves + character)
    - url.PathEscape(): For paths (encodes + character)
  - Structured processing via url.Values

  Implementation Example:
  // For queries
  query := url.QueryEscape("Hello World")  // "Hello+World"

  // For paths
  path := url.PathEscape("math+symbol")    // "math%2Bsymbol"

  // Structured approach
  params := url.Values{}
  params.Add("query", "Hello World")
  queryStr := params.Encode()  // Properly encoded

  4. cpp-httplib Issues and Improvement Direction

  Current Problems:
  - Single encoding function handles everything
  - Context-agnostic (same processing for paths and queries)
  - Double encoding during redirects

  Lessons from Other Libraries:

  A. Context Separation Approach (Go style)

  // Proposed implementation
  std::string encode_query_component(const std::string& s);
  std::string encode_path_component(const std::string& s);

  B. Configuration Flag Control (libcurl style)

  client.set_query_encoding_mode(PRESERVE_PLUS | RFC3986_STRICT);

  C. Structured Parameter Processing (Python requests style)

  // Decompose URL into components for proper handling
  struct URLComponents {
      std::string scheme, host, path;
      std::map<std::string, std::string> query_params;
  };

  5. Recommended Solution

  Phased Approach:
  1. Immediate fix: Disable path_encode_ during redirects
  2. Medium-term improvement: Separate encoding for queries and paths
  3. Long-term improvement: Flexible configuration options like other libraries

  This analysis reveals that cpp-httplib has room for improvement compared to other major
  libraries in terms of context awareness and configuration flexibility for URL encoding.
 ```

**Comment by DeltaRaidcore** [comment]
I think it's safe to assume that the `Location` header already correctly encodes the full query.

I do not know how httplib handles it internally, specifically, whether redirect clients are disposed after use.

1. If they are disposed after use, encoding could be disabled.
2. If they are not disposed - meaning modifying the client directly impacts future usage of the same host - then the safe route would be to treat just the `Location` header separate to normal URLs and fully decode host + query, based on the assumption that the `Location` header is correctly encoded to begin with.

**Comment by JoeMylinn** [comment]
It depends on if the redirect is on the same host or not. When on the same host, the original client is used, and modifying it would affect future usage of it. That's the issue with my original fix.

Always decoding the path and query has it's own issues, since it won't be encoded again if `path_encode_` is `false`.

Since as you've said, the `Location` header should always be encoded already anyway, I think the best option would be to skip decoding the path entirely and just parse the query like so:
```cpp
if (!next_query.empty()) { detail::parse_query_text(next_query, req.params); }
```
This will ensure the query is always properly parsed, we don't have to worry about doing any encoding or decoding in the redirect function, and this could be further adjusted in the future to allow config options to change how a `+` should be treated in a query (although I really don't see why you'd want to, as that wouldn't follow the standard).

JoeMylinn:
I went ahead and committed my suggestion, I think this is the best solution for now.

yhirose:
@JoeMylinn I implemented the 2nd recommendation that the Claude Code suggested in #2190. This PR fixes not only this issue, but also #2185.

```
2. Medium-term improvement: Separate encoding for queries and paths
```

PR #2197, opened 2025-08-07
CMake: Check pointer size at configure time

thbeu:
I only noticed at (cross) compile time that 32-bit platforms are no more supported (breaking change after upgrading from v0.22.0 to v0.24.0). This check can easily moved to configure time.


yhirose:
@thbeu Thanks!

PR #2198, opened 2025-08-08
Fix mingw-w64 building

apocelipes:
Since v0.23.1 cpp-httplib can not be complied with mingw-w64. The error message looks like this:

```console
> D:\a\_temp\msys64\mingw64\bin\x86_64-w64-mingw32-g++ -c -m64 -std=c++11 -IC:\Users\runneradmin\AppData\Local\.xmake\packages\c\cpp-httplib\v0.25.0\cd2fddbf26ef4285ab9708e0209f52e0\include -o D:\a\_temp\msys64\tmp\.xmake\250808\_97BBC115DC5B437086AEA0BA215614B0.o D:\a\_temp\msys64\tmp\.xmake\250808\_7D0C25F8195246278D1C039D74425C23.cpp
> checking for c++ includes(httplib.h)
> checking for c++ snippet(test)
checkinfo: ...amdir\core\sandbox\modules\import\core\tool\compiler.lua:84: @programdir\modules\core\tools\gcc.lua:1035: In file included from D:\a\_temp\msys64\tmp\.xmake\250808\_7D0C25F8195246278D1C039D74425C23.cpp:2:
C:\Users\runneradmin\AppData\Local\.xmake\packages\c\cpp-httplib\v0.25.0\cd2fddbf26ef4285ab9708e0209f52e0\include/httplib.h: In member function 'bool httplib::detail::mmap::open(const char*)':
C:\Users\runneradmin\AppData\Local\.xmake\packages\c\cpp-httplib\v0.25.0\cd2fddbf26ef4285ab9708e0209f52e0\include/httplib.h:3081:14: error: '::CreateFile2' has not been declared; did you mean 'CreateFileW'?
 3081 |   hFile_ = ::CreateFile2(wpath.c_str(), GENERIC_READ, FILE_SHARE_READ,
      |              ^~~~~~~~~~~
      |              CreateFileW
```

image

You can find the whole CI log here: [link](https://github.com/xmake-io/xmake-repo/actions/runs/16819883022/job/47644569540?pr=7837)

The function `::CreateFile2` needs `_WIN32_WINNT >= 0x0602`, however mingw-w64 in our CI cannot detect the correct value of it and set the value to `0x0600` and `0x0502`, that makes the code cannot use `CreateFile2`.

Because cpp-httplib requires Windows 10 or newer, let's just set the `_WIN32_WINNT` to `_WIN32_WINNT_WIN10` if we are using mingw-w64. Now mingw-w64 should compile the code.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -124,6 +124,22 @@ if(CMAKE_SIZEOF_VOID_P LESS 8)
 	message(WARNING "Pointer size ${CMAKE_SIZEOF_VOID_P} is not supported. Please use a 64-bit compiler.")
 endif()
 
+# check CreateFile2 for mingw
+INCLUDE(CheckCXXSourceCompiles)
+if(MINGW)
+	check_cxx_source_compiles("#include <windows.h>
+	#include <fileapi.h>
+	int main() {
+		HANDLE h = CreateFile2(L\"nul\", GENERIC_READ, FILE_SHARE_READ, OPEN_EXISTING, NULL);
+		if (h != INVALID_HANDLE_VALUE) CloseHandle(h);
+		return 0;
+	}" HAVE_CREATEFILE2)
+
+	if(HAVE_CREATEFILE2)
+		add_compile_definitions(HAVE_CREATEFILE2)
+	endif()
+endif()
+
 # Set some variables that are used in-tree and while building based on our options
 set(HTTPLIB_IS_COMPILED ${HTTPLIB_COMPILE})
 set(HTTPLIB_IS_USING_CERTS_FROM_MACOSX_KEYCHAIN ${HTTPLIB_USE_CERTS_FROM_MACOSX_KEYCHAIN})
--- a/httplib.h
+++ b/httplib.h
@@ -3078,8 +3078,13 @@ inline bool mmap::open(const char *path) {
   auto wpath = u8string_to_wstring(path);
   if (wpath.empty()) { return false; }
 
+#if defined(HAVE_CREATEFILE2)
   hFile_ = ::CreateFile2(wpath.c_str(), GENERIC_READ, FILE_SHARE_READ,
                          OPEN_EXISTING, NULL);
+#else
+  hFile_ = ::CreateFileW(wpath.c_str(), GENERIC_READ, FILE_SHARE_READ, NULL,
+                         OPEN_EXISTING, FILE_ATTRIBUTE_NORMAL, NULL);
+#endif
 
   if (hFile_ == INVALID_HANDLE_VALUE) { return false; }
 


thbeu (commented):
I consider this a kludge. Any chance to define/redefine _WIN32_WINNT to the appropriate value outside? Otherwise what about a macro HAVE_CREATEFILE2?

apocelipes:
> I consider this a kludge. Any chance to define/redefine _WIN32_WINNT to the appropriate value outside? Otherwise what about a macro HAVE_CREATEFILE2?

What can we do if `HAVE_CREATEFILE2` was false? I think fall back to `CreateFileW` is acceptable.

Code could be like:

```c++
#if defined(HAVE_CREATEFILE2)
  hFile_ = ::CreateFile2(wpath.c_str(), GENERIC_READ, FILE_SHARE_READ,
                         OPEN_EXISTING, NULL);
#else
  hFile_ = ::CreateFileW(wpath.c_str(), GENERIC_READ, FILE_SHARE_READ, NULL,
                         OPEN_EXISTING, FILE_ATTRIBUTE_NORMAL, NULL);
#endif
```

Edited: CreateFileW has different arguments, fixed.

apocelipes:
The code has been updated. I add the macro `HAVE_CREATEFILE2`.

yhirose:
@apocelipes thanks for the pull request. However, I don't prefer seeing `__MINGW32__` and introducing another macro for this purpose. I made a different pull request to fix this issue at #2203. Could you please try it to see if it works on your machine, since I don't have the mingw-w64 environment? Thanks!

yhirose:
Or, I am thinking to use only CreateFileW in httplib.h. :)

apocelipes:
@yhirose Thanks for your review. 

I think use `CreateFile2` is fine because it is a modern and safer API.

The code can be simplified and no need to introduce a new macro. You could take a look at the new code:

```
#if _WIN32_WINNT >= _WIN32_WINNT_WIN8
  hFile_ = ::CreateFile2(wpath.c_str(), GENERIC_READ, FILE_SHARE_READ,
                         OPEN_EXISTING, NULL);
#else
  hFile_ = ::CreateFileW(wpath.c_str(), GENERIC_READ, FILE_SHARE_READ, NULL,
                         OPEN_EXISTING, FILE_ATTRIBUTE_NORMAL, NULL);
#endif
```

This is more simple than use `std::once_flag` and no runtime performance cost.

thbeu:
We already have the `_WIN32_WINNT` version check in

https://github.com/yhirose/cpp-httplib/blob/3f44c80fd345e859d6f61b2dda566aa3329ce35e/httplib.h#L29-L34

CreateFileW will not be used then.

apocelipes:
So, the `_WIN32_WINNT` is greater than `_WIN32_WINNT_WIN8` but `CreateFile2` does not exist. It is toolchain's bug.

I can add a cmake check to set the macro `HAVE_CREATEFILE2` but maybe it's not worth for it.

apocelipes:
A cmake compile check has been added. PTAL.

apocelipes:
I upgraded the toolchain and it fixed. I think this PR is no longer needed. Thank you for your advice.

PR #2203, opened 2025-08-09
Fix #2198

yhirose:


thbeu (commented):
I prefer the compile-time switch over this run-time switch since we know a-priori that CreateFile2 is available on MSVC and Win8.

apocelipes:
@yhirose Thanks for watching into this issue. 

This patch cannot work, because my mingw-w64 does not have `CreateFile2`. The code use it on runtime so compile generates compiler errors.

I think we need to introduce a macro to determine which function can be used at compiling time.

PR #2204, opened 2025-08-09
Add condition to check CMAKE_SYSTEM_VERSION defined

toge:
`CMAKE_SYSTEM_VERSION` is defined on specific system (ex. Windows and macOS).
Please add a condition to check CMAKE_SYSTEM_VERSION defined.


yhirose:
@toge thanks for the pull request!
@sum01 @jimmy-park could you please review it for me as usual?

sum01:
Isn't it only not defined, as per spec, if cross-compiling and you forget to set it?
In a normal build it should get set automatically.

yhirose:
@toge could you reply to the comment from @sum01?

toge:
@yhirose @sum01 
I apologize for my misunderstanding.

This occurred while creating the vcpkg port, but since it is a limitation on the vcpkg side, I will address it with a vcpkg patch.

I am sorry for taking up your valuable time.

PR #2208, opened 2025-08-16
build(meson): fix new build option names

Tachi107:
This is a follow-up to commit 4ff7a1c858faeab14d0f9e8d533a9787cab1de08, which introduced new simplified build options and deprecated the old ones. I forgot to also change the various get_option() calls, effectively rendering the new option names useless, as they would not get honoured.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -39,12 +39,12 @@ endif
 deps = [dependency('threads')]
 args = []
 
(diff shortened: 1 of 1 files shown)


yhirose:
@Tachi107 thanks!

PR #2222, opened 2025-08-26
osx: fix inconsistent use of the macro TARGET_OS_OSX

sergio-nsk:
Fixed the build error on iOS:

```
httplib.h:3583:3: error: unknown type name 'CFStringRef'
  870 |   CFStringRef hostname_ref = CFStringCreateWithCString(
```

Note, `TARGET_OS_OSX` is defined but is 0 when `TARGET_OS_IOS` is 1,
and vise versa. Hence, `TARGET_OS_MAC` should have been used, that is
set to 1 for the both targets.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -3578,7 +3578,7 @@ inline int getaddrinfo_with_timeout(const char *node, const char *service,
   }
 
   return ret;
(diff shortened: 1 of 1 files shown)


yhirose:
@sergio-nsk thanks!

Issue #2229, opened 2025-08-28
Bounds checking fix on regex


homer6:
This fixes a compiler warning on build. This may be fixed already, but here is the fix and version that I was using.

The fix adds proper bounds checking before accessing regex match groups, which should eliminate the array bounds warning.

<img width="837" height="408" alt="Image" src="https://github.com/user-attachments/assets/d56bcfb4-b336-4028-84f4-ad5d688c7d07" />

//  https://github.com/yhirose/cpp-httplib
//  9452c0a4b69c5e4e31169ed32e961d330695122c  (May 13, 2022)
//  Release v0.10.7
//  httplib.h

yhirose:
@homer6 thanks for the report. Could you please send a pull request?

homer6:
All set. Thanks

#2230

yhirose:
I closed it. My comment in #2230 explains the reason.

PR #2230, opened 2025-08-28
Added bounds checks to parse_www_authenticate

homer6:
Added bounds checks. Removes compilers warning. Requires further testing.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -6385,13 +6385,16 @@ inline bool parse_www_authenticate(const Response &res,
         auto beg = std::sregex_iterator(s.begin(), s.end(), re);
         for (auto i = beg; i != std::sregex_iterator(); ++i) {
           const auto &m = *i;
+          // Ensure we have at least the expected capture groups
+          if (m.size() < 4) continue;
+          
           auto key = s.substr(static_cast<size_t>(m.position(1)),
                               static_cast<size_t>(m.length(1)));
-          auto val = m.length(2) > 0
-                         ? s.substr(static_cast<size_t>(m.position(2)),
+          auto val = (m.size() > 2 && m.length(2) > 0)
+                        ? s.substr(static_cast<size_t>(m.position(2)),
                                     static_cast<size_t>(m.length(2)))
-                         : s.substr(static_cast<size_t>(m.position(3)),
-                                    static_cast<size_t>(m.length(3)));
+                        : (m.size() > 3 ? s.substr(static_cast<size_t>(m.position(3)),
+                                                    static_cast<size_t>(m.length(3))) : "");
           auth[key] = val;
         }
         return true;


yhirose:
@homer6 thanks for the pull request. Could you please add a unit test case in `test/test.cc` including normal and abnormal cases? Also the style-check fails.
Thanks!

yhirose:
@homer6 I looked into this problem carefully and checked with GitHub Copilot. My conclusion is that this pull request is not necessary, since the purpose of this change is to suppress warnings produced by certain compilers. But if you find any specific cases to cause the out of bounds error, could you put it in `test/test.cc`? Otherwise, I am thinking to close this pull request and #2229.

**Question:**  
Here is the regular expression. Is the bounds check in the pull request really necessary?

`std::regex(R"~((?:(?:,\s*)?(.+?)=(?:"(.*?)"|([^,]*))))~")`

---

**Answer:**  
This regular expression  
`std::regex(R"~((?:(?:,\s*)?(.+?)=(?:"(.*?)"|([^,]*))))~")`  
is commonly used to parse authentication headers in the form `key="value"` or `key=value`. The groups are assigned as follows:

- `(.+?)` ... key
- `("(.*?)"|([^,]*))` ... value (with or without double quotes)

### Number of Match Groups
With this regex and `std::smatch match`:
- `match[0]` ... entire match
- `match[1]` ... key
- `match[2]` ... value with double quotes  
- `match[3]` ... value without double quotes

### Do you really need a bounds check?

#### 1. Using `std::regex_search` or `std::regex_match`
- **If a match is found (`std::regex_match` returns `true`), `match.size()` will always be the number of capturing groups plus one (for the whole match).**
- So accessing `match[1]` or `match[2]` is safe and will not cause out-of-bounds access.
- If `match[n].matched` is `false`, that means the group was empty, but accessing it is still safe.

#### 2. Compiler Warnings
- Some static analysis tools or compilers may warn that `match[n]` could be out-of-bounds.
- However, if the match succeeded, that's not possible by design.

#### 3. Typical cpp-httplib Code Example
For example:
```cpp
if (std::regex_search(str, match, re)) {
    auto key = match[1].str();
    auto value = match[2].matched ? match[2].str() : match[3].str();
}
```
With this usage, **a bounds check is not fundamentally required**.

---

### Conclusion

- **The bounds check added in the pull request serves mainly to suppress warnings, not to increase safety.**
- "Is it really necessary?" → As long as the code and regex are correct, it is **not necessary**.
- If you get warnings in your environment, adding `if (match.size() > n)` will suppress them, but it does **not improve actual safety**.

---

#### Reference: Official Documentation
- [cppreference: std::match_results](https://en.cppreference.com/w/cpp/regex/match_results)
  > "If the regular expression match was successful, then the number of sub-matches is equal to the number of capturing parentheses in the regular expression plus one for the full match."

---

If you suspect there is a bug where `match[n]` could be out-of-bounds, please provide the relevant code for further investigation.

homer6:
Will do. Thx!

PR #2231, opened 2025-09-02
Avoid multiple reallocations when receive a big buffer in post method call.

Dwordcito:
**Memory optimization in POST method handling**
- Improved buffer management to avoid multiple reallocations when processing large POST requests
- This change reduces memory fragmentation and improves performance for applications that handle substantial POST data

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -209,11 +209,11 @@ using ssize_t = __int64;
 #endif // _MSC_VER
 
 #ifndef S_ISREG
-#define S_ISREG(m) (((m) & S_IFREG) == S_IFREG)
+#define S_ISREG(m) (((m)&S_IFREG) == S_IFREG)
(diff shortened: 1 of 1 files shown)


Dwordcito:
@yhirose proxy test was broken in another PR.

yhirose:
@Dwordcito, thanks for the pull request. However, I am not going to merge it because this modification affects a request handler with ContentReceiver. In this case, res.body will not be used at all, but this change allocates unnecessary memory.

```cpp
  svr.Post("/content_receiver", [&](const Request &, Response &res,
                                    const ContentReader &content_reader) {
    content_reader([&](const char *data, size_t data_length) {
      // Handle `data` without pushing it to `res.body` at all.
      return true;
    });
  });
```

PR #2234, opened 2025-09-08
Fix: Fix Windows Cross-Compilation

apocelipes:
When cross-compiling for Windows systems, CMake will report an error because the CMAKE_SYSTEM_VERSION variable has not been set:

image

Checking whether this variable has been set resolves the error. A warning message is specifically output for this situation.


thbeu:
Duplicate of #2204.

apocelipes:
> Duplicate of #2204.

This is not vcpkg. And seems like this error happens with many tools. So I think it should be fixed.

apocelipes:
> Duplicate of #2204.

See the document:
image

So check whether the variable is set and print a warning message is the right way for handling cross-compilation.

thbeu:
Yes, it should be fixed in the lib.

yhirose:
@sum01 @jimmy-park could you please review this pull request? Thanks!

CMakeLists.txt
```diff
@@ -117,8 +117,14 @@ if(BUILD_SHARED_LIBS AND WIN32 AND HTTPLIB_COMPILE)
...
 
-if(CMAKE_SYSTEM_NAME MATCHES "Windows" AND ${CMAKE_SYSTEM_VERSION} VERSION_LESS "10.0.0")
-	message(SEND_ERROR "Windows ${CMAKE_SYSTEM_VERSION} or lower is not supported. Please use Windows 10 or later.")
+if(CMAKE_SYSTEM_NAME MATCHES "Windows")
+	if(CMAKE_SYSTEM_VERSION)
+		if(${CMAKE_SYSTEM_VERSION} VERSION_LESS "10.0.0")
+			message(SEND_ERROR "Windows ${CMAKE_SYSTEM_VERSION} or lower is not supported. Please use Windows 10 or later.")
+		endif()
+	else()
+		message(WARNING "The target is Windows but CMAKE_SYSTEM_VERSION is not set, the default system version is set to Windows 10 or later.")
```
sum01: Is it actually setting a default system version anywhere? I don't remember it being in the Cmake at least.

apocelipes: CMake >= 3.24 on Windows ignores this variable and always finds the latest version of the SDK, so no need to set it.

For version < 3.24, we can use `set(CMAKE_SYSTEM_VERSION "10.0.19041.0")`. Users can also pass values from command lines to it.

I added it. @sum01 PTAL.

sum01: Any particular reason for that specific version? Just curious.

apocelipes: > Any particular reason for that specific version? Just curious.

The version 10.0.19041.0 is for Windows10 20H1 (ver 2004) and later. This should include most Windows 10 instances which still in running.

sum01 (approved):

yhirose:
Thanks all!

PR #2237, opened 2025-09-12
Fix client use i pv6 address with default port, host was not properly wrapped by [], java servlet parsing header host will report an error

lifw555:


yhirose:
@lifw555 thanks for the pull request. Could you please take a look at broken unit tests?

test/test.cc
```diff
@@ -2239,7 +2239,7 @@ TEST(RedirectFromPageWithContentIP6, Redirect) {
     res.set_content("Hello World!", "text/plain");
   });
 
-  auto th = std::thread([&]() { svr.listen("::1", 1234); });
+  auto th = std::thread([&]() { svr.listen("[::1]", 1234); });
```
yhirose: Shouldn't be changed.

test/test.cc
```diff
@@ -2374,7 +2374,7 @@ TEST(BindServerTest, DISABLED_BindDualStack) {
     EXPECT_EQ("Hello World!", res->body);
   }
   {
-    Client cli("::1", PORT);
+    Client cli("[::1]", PORT);
```
yhirose: Shouldn't be changed.

test/test.cc
```diff
@@ -3650,7 +3650,7 @@ void performance_test(const char *host) {
 
 TEST(BenchmarkTest, localhost) { performance_test("localhost"); }
 
-TEST(BenchmarkTest, v6) { performance_test("::1"); }
+TEST(BenchmarkTest, v6) { performance_test("[::1]"); }
```
yhirose: Shouldn't be changed.

yhirose:
@lifw555 I added comments. Those tests shouldn't be changed, since cpp-httplib accepts IPv6 addresses that are not fully qualified. (https://en.wikipedia.org/wiki/IPv6_address)

So could you please revert the changes and update httplib.h so that all the existing tests pass as they are? (In other words, the current change in httplib.h isn't sufficient and introduces a breaking change that may cause problems for existing users' code.)

lifw555:
sorry，I will restore the code I modified back to its original state。

lifw555:
This fixed should be correct

lifw555:
The default ports (80, 443) in Linux do not allow listening, so the test case cannot pass. I don't know how to modify it.

yhirose:
@lifw555 I resolved this problem in a different way. Please let me know if this fix doesn't work on your environment with Java Servlet.

lifw555:
> @lifw555 I resolved this problem in a different way. Please let me know if this fix doesn't work on your environment with Java Servlet.

it is worked.

PR #2241, opened 2025-09-20
build(meson): use C++17 for gtest >= 1.17.0

Tachi107:


yhirose:
@Tachi107 thanks for the fix!

Tachi107:
👍

PR #2249, opened 2025-10-15
[cmake] FindBrotli: do not add `Brotli::` targets if they already exist

crueter:
Not checking for this is terrible practice.


yhirose:
@sum01 @jimmy-park could you please review this change?

crueter:
It's worth noting that this is kind of a band-aid, there's some other weird stuff going on with the FindBrotli call that I looked at but this ended up being the most elegant. If nobody else looks I will take a look later, but the gist of it is that ALIAS targets are evil

yhirose:
@crueter @sum01 thanks for the fix!

PR #2253, opened 2025-10-25
[build] Fix MinGW missing GetAddrInfoExCancel definition

crueter:
MinGW does not define GetAddrInfoExCancel in its wstcpi whatever header, so to get around this we can just load it with GetProcAddress et al.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -3560,9 +3560,10 @@ inline int getaddrinfo_with_timeout(const char *node, const char *service,
         ::WaitForSingleObject(event, static_cast<DWORD>(timeout_sec * 1000));
     if (wait_result == WAIT_TIMEOUT) {
 #ifdef __MINGW32__
-      typedef INT (WSAAPI *PFN_GETADDRINFOEXCANCEL)(HANDLE *CancelHandle);
-      auto wsdll = LoadLibraryW((wchar_t*) "ws2_32.lib");
-      PFN_GETADDRINFOEXCANCEL GetAddrInfoExCancel = (PFN_GETADDRINFOEXCANCEL) GetProcAddress(wsdll, "GetAddrInfoExCancel");
+      typedef INT(WSAAPI * PFN_GETADDRINFOEXCANCEL)(HANDLE * CancelHandle);
+      auto wsdll = LoadLibraryW((wchar_t *)"ws2_32.lib");
+      PFN_GETADDRINFOEXCANCEL GetAddrInfoExCancel =
+          (PFN_GETADDRINFOEXCANCEL)GetProcAddress(wsdll, "GetAddrInfoExCancel");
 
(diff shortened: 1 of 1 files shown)


yhirose:
Thanks for the pull request, but unfortunately cpp-httplib doesn't support MinGW.
https://github.com/yhirose/cpp-httplib?tab=readme-ov-file#windows
https://github.com/search?q=repo%3Ayhirose%2Fcpp-httplib+MinGW&type=issues

crueter:
> Thanks for the pull request, but unfortunately cpp-httplib doesn't support MinGW. https://github.com/yhirose/cpp-httplib?tab=readme-ov-file#windows https://github.com/search?q=repo%3Ayhirose%2Fcpp-httplib+MinGW&type=issues

Supporting MinGW is a good target to have. I confirmed through my testing that this makes it fully supported, and all functionality works as expected.

Not to mention that MSVC is absolutely abysmal and a complete joke of a compiler, so allowing users who wish to stay away from MSVC the chance to use httplib would be highly appreciated.

oo8dev:
> > Thanks for the pull request, but unfortunately cpp-httplib doesn't support MinGW. https://github.com/yhirose/cpp-httplib?tab=readme-ov-file#windows https://github.com/search?q=repo%3Ayhirose%2Fcpp-httplib+MinGW&type=issues
> 
> Supporting MinGW is a good target to have. I confirmed through my testing that this makes it fully supported, and all functionality works as expected.
> 
> Not to mention that MSVC is absolutely abysmal and a complete joke of a compiler, so allowing users who wish to stay away from MSVC the chance to use httplib would be highly appreciated.

Yup, this should be re-opened, especially that this PR won't hurt anybody.

oo8dev:
Temporary fix for those who are affected (add this code before including httplib.h):
```cpp
#ifdef _WIN32
#	define WIN32_LEAN_AND_MEAN
#	include <windows.h>
#	include <ws2tcpip.h>

#	ifdef __MINGW32__
typedef INT(WSAAPI *PFN_GETADDRINFOEXCANCEL)(HANDLE *CancelHandle);
static PFN_GETADDRINFOEXCANCEL func_get_addr_info_ex_cancel = nullptr;

static INT WSAAPI GetAddrInfoExCancel(LPHANDLE lpHandle) {
	if(func_get_addr_info_ex_cancel == nullptr) {
		auto *wsdll = LoadLibraryW((wchar_t *)"ws2_32.lib");
		func_get_addr_info_ex_cancel = (PFN_GETADDRINFOEXCANCEL)GetProcAddress(wsdll, "GetAddrInfoExCancel");
	}

	return func_get_addr_info_ex_cancel(lpHandle);
}
#	endif
#endif

#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wmissing-field-initializers"
#include "httplib.h"
#pragma GCC diagnostic pop
```

PR #2262, opened 2025-11-03
More robust error handling

chansikpark:
Crossing some t's and dotting some i's following #2260, I decided to try [this suggestion from Copilot](https://github.com/yhirose/cpp-httplib/pull/2260#issuecomment-3477014404):

>  It’s also useful to add a test where the client continues sending a large body and the server immediately returns 414 and closes the socket.

I found that the client SIGPIPEs. And even after handling these, the response object doesn't get populated when the socket is closed mid-transfer (ie for sufficiently large transfers).

So I made the following changes:

- Check if socket is alive before waiting to write to it to prevent SIGPIPE
- Consume SIGPIPE on client
- Client reads response line and headers after bad request before returning

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -8618,11 +8618,7 @@ inline ClientImpl::ClientImpl(const std::string &host, int port,
                               const std::string &client_key_path)
     : host_(detail::escape_abstract_namespace_unix_domain(host)), port_(port),
       host_and_port_(detail::make_host_and_port_string(host_, port, is_ssl())),
-      client_cert_path_(client_cert_path), client_key_path_(client_key_path) {
-#ifndef _WIN32
-  signal(SIGPIPE, SIG_IGN);
-#endif
-}
+      client_cert_path_(client_cert_path), client_key_path_(client_key_path) {}
 
 inline ClientImpl::~ClientImpl() {
   // Wait until all the requests in flight are handled.


chansikpark:
I'll see if I can do a little more due dilligence before marking for review.

Also 

> Move one basic test assertion out of WIN32-only block

Should be

> Move one basic test assertion out of non-WIN32 block

httplib.h
```diff
@@ -8618,7 +8618,11 @@ inline ClientImpl::ClientImpl(const std::string &host, int port,
                               const std::string &client_key_path)
     : host_(detail::escape_abstract_namespace_unix_domain(host)), port_(port),
       host_and_port_(detail::make_host_and_port_string(host_, port, is_ssl())),
-      client_cert_path_(client_cert_path), client_key_path_(client_key_path) {}
+      client_cert_path_(client_cert_path), client_key_path_(client_key_path) {
+#ifndef _WIN32
+  signal(SIGPIPE, SIG_IGN);
```
thbeu: A C/C++ library shall not call signal.

chansikpark: I'd been wondering why eg the README has this note:

>  If your program needs to avoid being terminated on SIGPIPE, the only fully general way might be to set up a signal handler for SIGPIPE to handle or ignore it yourself.

While the server code consumes SIGPIPE.. :

https://github.com/yhirose/cpp-httplib/blob/4b2b851dbb91f9d6c2299976b5d03f7c5c1a312d/httplib.h#L7322-L7328

In any case, I'd concede that the other two changes probably make the client-side `signal` unnecessary for what I was trying to do. Been meaning to play around a little more to better understand what I'm doing. Will revert particular change in the meantime.

yhirose:
@chansikpark, what is the current status of this issue?

chansikpark:
No additional progress. Might be a few more weeks from my end. Turns out, a
surgical wound needs a bit more healing. 🙃

Spixmaster (commented):
`bool wrote = write_request(strm, req, close_connection, error);`

The past participle of "write" is "written". The simple past does not fit here from my perspective.

yhirose:
@chansikpark no rush. Hope you get better soon!

PR #2266, opened 2025-11-09
CMake: Add HTTPLIB_SHARED option, don't define BUILD_SHARED_LIBS

### LINKED ISSUE #2263: Defining `BUILD_SHARED_LIBS` option may break superproject build when using `add_subdirectory`
I was surprised to discover that adding `add_subdirectory(cpp-httplib)` to our project `CMakeLists.txt` has broken the build because some libraries that were supposed to be built as shared have suddenly started being built as static. The reason for this is that this library defines `BUILD_SHARED_LIBS` as `OFF` and, as CMake [documentation says](https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html):

> Note that if bringing external dependencies directly into the build, such as with [FetchContent](https://cmake.org/cmake/help/latest/module/FetchContent.html#module:FetchContent) or a direct call to [add_subdirectory()](https://cmake.org/cmake/help/latest/command/add_subdirectory.html#command:add_subdirectory), and one of those dependencies has such a call to [option(BUILD_SHARED_LIBS ...)](https://cmake.org/cmake/help/latest/command/option.html#command:option), the top level project must also call [option(BUILD_SHARED_LIBS ...)](https://cmake.org/cmake/help/latest/command/option.html#command:option) before bringing in its dependencies.

and our project doesn't define this option. We probably could define it, but I think it would be better if cpp-httplib didn't define it instead, especially when `HTTPLIB_COMPILE` is `OFF`, i.e. when it's not even used by this library itself. I believe the best approach would be to use this option as default value for some new `HTTPLIB_SHARED` option and only use this option to decide which kind of library to build. This would have the following desirable consequences:

1. `BUILD_SHARED_LIBS` will still be respected if it's defined (by the superproject or directly on the command line when configuring).
2. It will become possible to specify whether static or shared cpp-httplib library should be built by setting `HTTPLIB_SHARED` without affecting anything else.
3. Adding this project to the existing build won't break anything else any more.

Please let me know if you'd like me to submit a PR implementing this proposal.

vadz:
To avoid surprises in the projects consuming the library, don't define BUILD_SHARED_LIBS option ourselves but just use its value, if provided, to define HTTPLIB_SHARED option which can be also set directly to specify whether we should build static or shared library.

Closes #2263.

-----

Note: I didn't update the documentation because I didn't find any mention of how to build the library there, please let me know if I've missed anything. I also didn't update the use of `BUILD_SHARED_LIBS` in `test/make-shared-library.sh` just to show that it still works the same as before.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 #[[
 	Build options:
-	* BUILD_SHARED_LIBS (default off) builds as a shared library (if HTTPLIB_COMPILE is ON)
+	* Standard BUILD_SHARED_LIBS is supported and sets HTTPLIB_SHARED default value.
 	* HTTPLIB_USE_OPENSSL_IF_AVAILABLE (default on)
 	* HTTPLIB_USE_ZLIB_IF_AVAILABLE (default on)
 	* HTTPLIB_USE_BROTLI_IF_AVAILABLE (default on)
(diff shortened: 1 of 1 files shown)


yhirose:
@sum01 @jimmy-park could you please review this change for me as usual? Thanks!

vadz:
No idea why the CI tests fail but this seems completely unrelated to the changes here.

sum01 (approved):

yhirose:
Thanks all!

PR #2268, opened 2025-11-10
Add Client methods with both content provider and receiver

clarkok:
Add most `Post` overloads for `Client`, with both content provider and content receiver.

* Result Post(const std::string &path, size_t content_length, ContentProvider content_provider, const std::string &content_type, ContentReceiver content_receiver, UploadProgress progress = nullptr)
* Result Post(const std::string &path, ContentProviderWithoutLength content_provider, const std::string &content_type, ContentReceiver content_receiver, UploadProgress progress = nullptr)
* Result Post(const std::string &path, const Headers &headers, size_t content_length, ContentProvider content_provider, const std::string &content_type, ContentReceiver content_receiver, DownloadProgress progress = nullptr)
* Result Post(const std::string &path, const Headers &headers, ContentProviderWithoutLength content_provider, const std::string &content_type, ContentReceiver content_receiver, DownloadProgress progress = nullptr)

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -1421,14 +1421,18 @@ public:
   Result Post(const std::string &path, const char *body, size_t content_length, const std::string &content_type, UploadProgress progress = nullptr);
   Result Post(const std::string &path, const std::string &body, const std::string &content_type, UploadProgress progress = nullptr);
   Result Post(const std::string &path, size_t content_length, ContentProvider content_provider, const std::string &content_type, UploadProgress progress = nullptr);
+  Result Post(const std::string &path, size_t content_length, ContentProvider content_provider, const std::string &content_type, ContentReceiver content_receiver, UploadProgress progress = nullptr);
   Result Post(const std::string &path, ContentProviderWithoutLength content_provider, const std::string &content_type, UploadProgress progress = nullptr);
+  Result Post(const std::string &path, ContentProviderWithoutLength content_provider, const std::string &content_type, ContentReceiver content_receiver, UploadProgress progress = nullptr);
   Result Post(const std::string &path, const Params &params);
   Result Post(const std::string &path, const UploadFormDataItems &items, UploadProgress progress = nullptr);
   Result Post(const std::string &path, const Headers &headers);
   Result Post(const std::string &path, const Headers &headers, const char *body, size_t content_length, const std::string &content_type, UploadProgress progress = nullptr);
   Result Post(const std::string &path, const Headers &headers, const std::string &body, const std::string &content_type, UploadProgress progress = nullptr);
   Result Post(const std::string &path, const Headers &headers, size_t content_length, ContentProvider content_provider, const std::string &content_type, UploadProgress progress = nullptr);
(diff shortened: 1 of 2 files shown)


yhirose:
@clarkok thanks for the pull request. But I am not going to approve it since it create another two POST specific implementations. Could you please rename `send_with_content_provider` to `send_with_content_provider_and_receiver`, and add your receiver support to it? Also could you add the same methods to PUT and PATCH for consistency? Then, I'll start review your pull request. Thanks!

clarkok:
Thanks @yhirose . I've done the changes you mentioned. Would you mind taking a look?

httplib.h
```diff
@@ -3847,6 +3847,8 @@ inline int getaddrinfo_with_timeout(const char *node, const char *service,
 
   // Allocate on the heap, so the resolver thread can keep using the data.
   auto state = std::make_shared<GetAddrInfoState>();
+  state->node = node;
```
yhirose: Could you explain why these lines need to be added? These are already added at line 3842 and 3843.

clarkok: Sorry I was not planning to take this change in the PR, I'm not very familar with how PR works in Github.

However this is a separated issue I found during my Android project. In short, this doesn't work, the `node` on the right side of the assignment is actually referring to the member of `GetAddrInforState` which is empty, instead of the function parameter. Thus the `getaddrinfo` below always return `EAI_SERVICE` as both `node` and `service` are empty, and probably `hints` are not in expected state.

I've written a short piece of repro code in the below link for your reference.

https://godbolt.org/z/9sW9EjjYK

I'll revert this back and create a new PR for the issue.

yhirose: Thanks for the explanation!

clarkok: NP, BTW after this fix the library is working on Android flawlessly

yhirose:
@clarkok Thanks for the excellent contribution!

linbrandon4:
This is a pretty meaningful API expansion. Adding Client methods that support both a content provider (streaming upload) and a content receiver (streaming response) is really useful for large uploads/downloads where you don’t want everything buffered in memory. The overload set also looks thoughtfully designed (with/without known length, with headers, and optional progress callbacks).

I also like the direction from review: instead of adding more POST-specific implementations, folding it into the existing internal send path (renaming send_with_content_provider → send_with_content_provider_and_receiver) keeps the codebase from drifting into duplicated code. Extending the same pattern to PUT and PATCH is a good consistency move too — otherwise users end up with “POST can stream both ways but PUT can’t” type surprises.

A couple things I’d sanity-check when reviewing this kind of change:

Behavior parity: do POST/PUT/PATCH all handle the same edge cases (chunked vs content-length, redirects if supported, connection reuse, error returns) the same way as before?

Receiver timing: make sure the receiver gets called in the right spot (after headers, during body streaming), and that early aborts / receiver returning false are handled cleanly (connection close vs keep-alive).

Progress callbacks: confirm progress is consistent with the new data flow (upload progress tied to provider writes, download progress tied to receiver reads).

No hidden breaking changes: adding overloads is great, but watch for ambiguous overload resolution in C++ if signatures are too close (especially with defaults / nullptr progress args).

The side thread about the resolver lines is also a good example of keeping scope tight — good call to revert that and put it in a separate PR. Overall, big diff but it’s additive and clearly improves streaming ergonomics, and the internal refactor request from yhirose should help keep maintenance manageable.

PR #2273, opened 2025-11-14
Fix struct member initialization issue in getaddrinfo_with_timeout 

clarkok:
The current member initialization for `struct GetAddrInfoState` does not work as the `node` on the right side of the assignment refers to the member `node` of `struct GetAddrInfoState`, which is empty for debug build and uninitialized for optimized build, instead of the `node` the function parameter.

This results in the following `getaddrinfo` getting called with unexpected parameters, thus it is yielding unexpected result. I found this issue on Android, and probably this also applies to platforms other than those in the `#if`.

Also this change fixes a potential memory leak. When timeout is hit on line 3857, and the `getaddrinfo` succeeds in the end, the `state->info` will be assigned to the pointer of a new `struct addrinfo` allocated by `getaddrinfo`, while no one will free that. This change adds a destructor on `GetAddrInfoState` to manage the lifecycle of that.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -3809,22 +3809,30 @@ inline int getaddrinfo_with_timeout(const char *node, const char *service,
   // Fallback implementation using thread-based timeout for other Unix systems
 
   struct GetAddrInfoState {
+    ~GetAddrInfoState() {
+      if (info) { freeaddrinfo(info); }
+    }
+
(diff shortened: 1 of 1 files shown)


yhirose:
@clarkok thanks for the bug fix!

Issue #2269, opened 2025-11-11
Generator-like Client response


ngxson:
Hi,

I would request a new feature in the library: allowing to read data from `httplib::Client` in a generator-like way.

The problem is that currently, to read data progressively, we will need to use callback function. An example can be found in this section: https://github.com/yhirose/cpp-httplib?tab=readme-ov-file#receive-content-with-a-content-receiver-1

```cpp
std::string body;

auto res = cli.Get("/large-data",
  [&](const char *data, size_t data_length) {
    body.append(data, data_length);
    return true;
  });
```

What I propose is to instead, allow using `res` as a generator, for example:

```cpp
std::string body;

auto res = cli.Get("/large-data");
while (res->has_data()) {
  body.append(res->data(), res->len());
  res->next();
}
```

This will be useful for the use case where we need to pipe the response back to `set_chunked_content_provider` in `httplib::Server` (for example, doing a reverse proxy)

ngxson:
> This will be useful for the use case where we need to pipe the response back to `set_chunked_content_provider` in `httplib::Server` (for example, doing a reverse proxy)

Otherwise, if there is a way to just give the `Result` returned by `Client::Get()` back to server's `Response`, I think it can also be a nice solution

yhirose:
@ngxson, thanks for the suggestion! It looks like it would need a lot of changes to httplib.h. I'll check how hard it is to implement when I have time.

yhirose:
@ngxson I implemented your suggestion at #2281. If you have any additional comments, please let me know. Thanks!
https://github.com/yhirose/cpp-httplib/blob/master/README.md#streaming-api
https://github.com/yhirose/cpp-httplib/blob/master/README-stream.md

PR #2281, opened 2025-11-27
New Stream API Support

yhirose:
This PR is trying to implement #2269. It requires C++20 compiler or later which supports C++ coroutines.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,7 +16,11 @@ example/benchmark
 example/redirect
 !example/redirect.*
 example/ssecli
+!example/ssecli.*
+example/ssecli-stream
+!example/ssecli-stream.*
 example/ssesvr
+!example/ssesvr.*
 example/upload
 !example/upload.*
 example/one_time_request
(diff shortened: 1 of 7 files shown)


yhirose:
@ngxson the only thing of this new stream API is that it requires C++20 which supports C++ coroutines.

Here is [README](https://github.com/yhirose/cpp-httplib/blob/generator/README-stream.md).

yhirose:
@ngxson I discarded my first implementation for C++20, then implemented the stream API which works with C++11 according to your request at #2269.

ngxson:
Thanks, indeed C++11 or 17 is better, but not necessarily as we already moved cpp-httplib to a dedicated unit in llama.cpp. We have some other vendors compiled with C++17 too, while the main project still uses 11, so it's probably fine even if internally you want to use C++20 features

ngxson:
Just read through the documentation, this is more than enough for me. Thanks again 😄

PR #2283, opened 2025-11-30
Optimize ThreadPool and MatcherBase constructors

Skylion007:
Add a missing reserve and missing std::move to each ctor respectively. The latter should really be caught by a clang-tidy perf linter.


yhirose:
@Skylion007 thanks for your contribution!

PR #2291, opened 2025-12-08
Add C++ modules support

mikomikotaishi:
This pull request adds support for C++20 modules through CMake. It is enabled by the `HTTPLIB_BUILD_MODULES` option (which requires `HTTPLIB_COMPILE` to be enabled, though it probably doesn't have to - I only forced this requirement because it seems to make the most sense to force the library to compile if modules are to be compiled).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -250,6 +250,22 @@ if(HTTPLIB_COMPILE)
 		message(FATAL_ERROR "Failed when trying to split cpp-httplib with the Python script.\n${_httplib_split_error}")
 	endif()
 
+	# If building modules, also generate the module file
+	if(HTTPLIB_BUILD_MODULES)
+		# Put the generate_module script into the build dir
+		configure_file(generate_module.py "${CMAKE_CURRENT_BINARY_DIR}/generate_module.py"
+			COPYONLY
+		)
+		# Generate the module file
+		execute_process(COMMAND ${Python3_EXECUTABLE} "${CMAKE_CURRENT_BINARY_DIR}/generate_module.py"
+			WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
+			ERROR_VARIABLE _httplib_module_error
+		)
+		if(_httplib_module_error)
+			message(FATAL_ERROR "Failed when trying to generate cpp-httplib module with the Python script.\n${_httplib_module_error}")
+		endif()
+	endif()
+
 	# split.py puts output in "out"
 	set(_httplib_build_includedir "${CMAKE_CURRENT_BINARY_DIR}/out")
 	add_library(${PROJECT_NAME} ${HTTPLIB_LIB_TYPE} "${_httplib_build_includedir}/httplib.cc")
--- a/split.py
+++ b/split.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 
-"""This script splits httplib.h into .h, .cc, and .cppm parts."""
+"""This script splits httplib.h into .h and .cc parts."""
 
 import os
 import sys
@@ -48,40 +48,13 @@ def main() -> None:
 
         os.makedirs(args.out, exist_ok=True)
 
-        # Find the Headers and Declaration comment markers
-        headers_start: int = -1
-        declaration_start: int = -1
-        for i, line in enumerate(lines):
-            if ' * Headers' in line:
-                headers_start = i - 1  # Include the /* line
-            elif ' * Declaration' in line:
-                declaration_start = i - 1  # Stop before the /* line
-                break
-
         in_implementation: bool = False
         cc_out: str = args.out + source_name
-        cppm_out: str = args.out + f"/{lib_name}.cppm"
         
-        with open(h_out, 'w') as fh, open(cc_out, 'w') as fc, open(cppm_out, 'w') as fm:
+        with open(h_out, 'w') as fh, open(cc_out, 'w') as fc:
             # Write source file
             fc.write("#include \"httplib.h\"\n")
             fc.write("namespace httplib {\n")
-            
-            # Write module file
-            fm.write("module;\n\n")
-            
-            # Write global module fragment (from Headers to Declaration comment)
-            # Filter out 'using' declarations to avoid conflicts
-            if headers_start >= 0 and declaration_start >= 0:
-                for i in range(headers_start, declaration_start):
-                    line: str = lines[i]
-                    if 'using' not in line:
-                        fm.write(line)
-            
-            fm.write("\nexport module httplib;\n\n")
-            fm.write("export extern \"C++\" {\n")
-            fm.write(f"{' ' * 4}#include \"httplib.h\"\n")
-            fm.write("}\n")
             
             # Process lines for header and source split
             for line in lines:
@@ -95,9 +68,9 @@ def main() -> None:
             
             fc.write("} // namespace httplib\n")
 
-        print(f"Wrote {h_out}, {cc_out}, and {cppm_out}")
+        print(f"Wrote {h_out} and {cc_out}")
     else:
-        print(f"{h_out}, {cc_out}, and module files are up to date")
+        print(f"{h_out} and {cc_out} are up to date")
 
 
 if __name__ == "__main__":
--- a/generate_module.py
+++ b/generate_module.py
@@ -1,0 +1,77 @@
+#!/usr/bin/env python3
+
+"""This script generates httplib.cppm module file from httplib.h."""
+
+import os
+import sys
+from argparse import ArgumentParser, Namespace
+from typing import List
+
+def main() -> None:
+    """Main entry point for the script."""
+
+    args_parser: ArgumentParser = ArgumentParser(description=__doc__)
+    args_parser.add_argument(
+        "-o", "--out", help="where to write the files (default: out)", default="out"
+    )
+    args: Namespace = args_parser.parse_args()
+
+    cur_dir: str = os.path.dirname(sys.argv[0])
+    if not cur_dir:
+        cur_dir = '.'
+    lib_name: str = "httplib"
+    header_name: str = f"/{lib_name}.h"
+    # get the input file
+    in_file: str = f"{cur_dir}{header_name}"
+    # get the output file
+    cppm_out: str = f"{args.out}/{lib_name}.cppm"
+
+    # if the modification time of the out file is after the in file,
+    # don't generate (as it is already finished)
+    do_generate: bool = True
+
+    if os.path.exists(cppm_out):
+        in_time: float = os.path.getmtime(in_file)
+        out_time: float = os.path.getmtime(cppm_out)
+        do_generate: bool = in_time > out_time
+
+    if do_generate:
+        with open(in_file) as f:
+            lines: List[str] = f.readlines()
+
+        os.makedirs(args.out, exist_ok=True)
+
+        # Find the Headers and Declaration comment markers
+        headers_start: int = -1
+        declaration_start: int = -1
+        for i, line in enumerate(lines):
+            if ' * Headers' in line:
+                headers_start = i - 1  # Include the /* line
+            elif ' * Declaration' in line:
+                declaration_start = i - 1  # Stop before the /* line
+                break
+
+        with open(cppm_out, 'w') as fm:
+            # Write module file
+            fm.write("module;\n\n")
+            
+            # Write global module fragment (from Headers to Declaration comment)
+            # Filter out 'using' declarations to avoid conflicts
+            if headers_start >= 0 and declaration_start >= 0:
+                for i in range(headers_start, declaration_start):
+                    line: str = lines[i]
+                    if 'using' not in line:
+                        fm.write(line)
+            
+            fm.write("\nexport module httplib;\n\n")
+            fm.write("export extern \"C++\" {\n")
+            fm.write(f"{' ' * 4}#include \"httplib.h\"\n")
+            fm.write("}\n")
+
+        print(f"Wrote {cppm_out}")
+    else:
+        print(f"{cppm_out} is up to date")
+
+
+if __name__ == "__main__":
+    main()


Spixmaster:
Great work. I also thought about this.

The approach is very similiar to what was done to https://github.com/nlohmann/json with https://github.com/nlohmann/json/pull/4799. Then, I noticed ...

Is this temporary approach with a file with `using ...` for the meantime while modules and no modules are supported?

This approach is not using modules natively but rather as an interface to the original way.

Does this method work without disadvantages?

mikomikotaishi:
This is the best way to my knowledge to support modules on top of a header-only or header/source library, allowing continued support for older versions while providing newer features as an option.

I'm not aware of any disadvantages to it besides a being additional translation unit to compile, but if I am wrong please correct me. The only glaring difference in API is that detail symbols are hidden as they are not exported, but in my opinion that's probably better not to expose detail symbols and flood IDE suggestions with implementation details.

Spixmaster:
What about compiled libraries? Is it possible to have the traditional method and modules installed in parallel?

I am thinking of repositories that ship compiled `.so` or `.a`.

This is relevant here, https://aur.archlinux.org/packages/cpp-httplib-compiled .

mikomikotaishi:
Yes I believe it's possible to use shared/static libraries with modules, all of my modular projects compiled to shared libraries that an executable consumes

yhirose:
@mikomikotaishi, thanks for the fine pull request! It's fantastic, but my concern is that someone needs to update `modules/httplib.cppm` whenever new external symbols are added or existing symbols are removed/renamed, and it could happen quite often. I am not planning to maintain this file. So if you cannot keep maintaining the file, I am probably not merge it since the file can be easily out-of-date. Is there a way to generate `modules/httplib.cppm` from `httplib.h` automatically?

@sum01 @jimmy-park @Tachi107 do you have any thought about this pull request?

mikomikotaishi:
I could create a Python script, or some other kinds of automated means of updating, which you could run every time it is updated. Until then I would be OK with maintaining this file, as it is a simple process. Such a script would probably comb through the file and add any symbols not part of a detail or internal namespace, or prefixed with an underscore, etc.

However, I am curious why it is not feasible to update the file manually. In case it isn't clear how, one can update the file by adding a `using httplib::NEW_SYMBOL_HERE;` into the `export namespace httlib { }` block, for each new symbol that is added. Do let me know if any more information is needed.

mikomikotaishi:
I have also seen some repositories use bots to push some commits too. Potentially one such bot could be set up to automatically populate the module with new changes each time there is a mismatch. I don't know anything about how to set this up, but I have seen this before and it could potentially be a solution (but I think the simplest one is just to run a Python script each time any update to the library happens).

mikomikotaishi:
Anyway, I think [this](https://github.com/stripe2933/imgui-module/blob/docking/generate.py) could be one such way of automatically updating the module.

yhirose:
@mikomikotaishi thanks for the additional explanation. I am ok with the following your suggestion.

> I have also seen some repositories use bots to push some commits too. Potentially one such bot could be set up to automatically populate the module with new changes each time there is a mismatch. I don't know anything about how to set this up, but I have seen this before and it could potentially be a solution.

We could automatically generate `modules/httplib.cppm` in a GitHub Actions script when `httplib.h` is pushed. (I can't accept the way to run a script to update the file manually, since I don't maintain any build systems in this repository. Please see https://github.com/yhirose/cpp-httplib/pull/955#issuecomment-855280013.)

mikomikotaishi:
OK, that makes sense to me. (I don't know anything about how to run GitHub Actions or write scripts for it however, so I'm afraid the most I can do is create a script for this.)

mikomikotaishi:
I'm not sure why there were failing workflows as I didn't change anything in the core library

mikomikotaishi:
Never mind, it seems the failing CI is happening upstream too.

split.py
yhirose, line 1: What is the purpose of the changes in this file? If it's only a matter of adding def main(), could you please revert this?

update_modules.py
yhirose, line 1: Could you please simply regenerate `httplib.cppm` from scratch with the latest `httplib.h`? It will make this script much simpler.

split.py
mikomikotaishi: Adding type annotations and using fstrings to clean up the code or make it easier to maintain

update_modules.py
mikomikotaishi: I think it's better to just search for changes to the script by just looking for declarations and removals in each change to `httplib.h`, especially due to all the possible edge cases

It seems a lot more complex to just regenerate the module every time

split.py
mikomikotaishi: Actually, is it necessary to ensure compatibility with Python 2? I noticed the script checks for whether the Python version is 3 or not. I presume most people nowadays are running 3 and wonder if this is necessary.

update_modules.py
mikomikotaishi: I am just curious though, is there a reason why updating the module file manually is out of the question? I don't imagine you will be adding hundreds of new symbols each update, and the process to update it is quite simple, if not menial when the number of new symbols is excessive.

.github/workflows/update-module-exports.yml
yhirose, line 1: Thanks for adding this GitHub Action. But after further consideration, I now think this process should be done in CMakeLists.txt. You could generate `modules/httplib.cppm` under this line https://github.com/yhirose/cpp-httplib/blob/87c2b4e584da64d80379e155480eeb954c7f9df5/CMakeLists.txt#L216

We actually do the similar to generate `httplib.h` and `httplib.cc` with `split.py` and put them in a distribution package. By doing this, we no longer need to keep `modules/httplib.cppm` in this repository. It will be created on the fly only when necessary.

@sum01 @jimmy-park Is my comment above correct?

split.py
yhirose: Thanks for the explanation. It's good code cleanup, but it should be done in a different pull request since it doesn't directly relate to this issue.

.github/workflows/update-module-exports.yml
mikomikotaishi: I think it would be difficult to parse an especially large file without syntax analysis or parsing library. I don't have a lot of experience with this sort of thing, and because the header contains things like method definitions outside of the class it seems to be a very complex task. Then including a C++ parser would require a dependency _even for a source generation script_ which would be additional bloat.

I guess this could be done by attaching to the header file that `split.py` generates, but I haven't actually seen what that looks like.

As for dynamically generating the module, this was raised before on the [Dear ImGui](https://github.com/ocornut/imgui) library which currently does something like this. I think this is a bad choice for module API (having a static file is obviously best for consumers to be able to just read it from a distance without additional interaction, i.e. a "what-you-see-is-what-you-get" sort of API), but ultimately as you are in charge I'll look into this if you prefer it.

update_modules.py
yhirose: As I already mentioned above, I don't maintain any build-system-related files. I just use my library as a single-header-only library. Other excellent contributors have kindly been maintaining the CMake and Meson build support for a long time. If you take the responsibility to update the file manually whenever httplib.h gets changed by contributors, that would be fantastic! 😄

split.py
mikomikotaishi: Do you think it's necessary to keep Python 2 support? If not, I'll remove the checks for it and move all of these changes to a new PR

update_modules.py
mikomikotaishi: I could maintain it for now, looking at what the `split.py` file generates it seems to be not too difficult to produce a module dynamically (of course, I have to look more closely into this first).

.github/workflows/update-module-exports.yml
mikomikotaishi: I'm looking more into this and I think if we want to allow CMake to compile the module if it's generated into an "out" directory, we have to force the out directory to be named "out" so that it can be written into the CMakeLists.txt script

split.py
yhirose: Sorry for the late reply. I have already approved the new PR and merged it. Thanks!

.github/workflows/update-module-exports.yml
mikomikotaishi: Anyway @yhirose do you have any opinion on the output directory having a hard-coded name?

yhirose: @sum01 @jimmy-park @Tachi107 could you please answer this question?

Tachi107:
If the file can be run during the build process (and if the output consists of machine-generated files, it should *only* run during build time), then the destination directory should be configurable (maybe defaulting to the the current working directory). Even better, if the output is a single file, the script should allow the user to specify the output file itself (full path).

This is because downstreams (like Debian, which is what I maintain the meson build scripts for) may have some requirements on where build products should be stored.

mikomikotaishi:
@Tachi107 CMake needs to know what the output directory is ahead of time to compile the module. How do you propose to solve this?

mikomikotaishi:
@yhirose I think there is one possible solution to allow both the directory to be user-specified while still supporting CMake module building, which is probably just to have the Python script generate the CMake file too. I don't know if this is too convoluted or awkward of a design though, so please do tell me your thoughts.

Tachi107:
Hi Miko,

On Tue Dec 9, 2025 at 10:46 PM CET, Miko wrote:
> @Tachi107 CMake needs to know what the output directory is ahead of time to compile the module. How do you propose to solve this?

You should use CMake's add_custom_command() function to invoke the 
script and pass it the output file path

https://cmake.org/cmake/help/latest/command/add_custom_command.html

Meson has a similar function, but I can do that myself after this gets 
merged.

modules/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
+add_library(httplib_module)
+
+target_sources(httplib_module
+    PUBLIC
+        FILE_SET CXX_MODULES FILES
+            httplib.cppm
+)
```
Tachi107: This requires CMake 3.28, but this project only requires CMake 3.0.

CMakeLists.txt
```diff
@@ -110,6 +110,7 @@ option(HTTPLIB_USE_CERTS_FROM_MACOSX_KEYCHAIN "Enable feature to load system cer
 option(HTTPLIB_USE_NON_BLOCKING_GETADDRINFO "Enables the non-blocking alternatives for getaddrinfo." ON)
 option(HTTPLIB_REQUIRE_ZSTD "Requires ZSTD to be found & linked, or fails build." OFF)
 option(HTTPLIB_USE_ZSTD_IF_AVAILABLE "Uses ZSTD (if available) to enable zstd support." ON)
+option(HTTPLIB_BUILD_MODULES "Build httplib modules (requires HTTPLIB_COMPILE to be ON)." OFF)
```
Tachi107: Why would this need to be an option? Does it add any other dependency? Cannot it be always be built, without an option?

update_modules.py
Tachi107, line 1: Which Python version does this script require? Is there a reason why, for example, `Optional[str]` is used instead of the built-in `str | None` introduced by [PEP 604](https://peps.python.org/pep-0604/)?

update_modules.py
```diff
@@ -0,0 +1,296 @@
...
+        result: CompletedProcess = subprocess.run(
+            ["git", "diff", base_ref, "--", file_path],
+            capture_output=True,
+            text=True,
+            check=True
+        )
+        return result.stdout
+    except CalledProcessError as e:
+        print(f"Error getting git diff: {e}", file=sys.stderr)
+        return None
```
Tachi107: Requiring a Git repository to built the modules is something many users would prefer to avoid. For example, users building from the release tarballs produced by GitHub will not have any Git history, and this would not work.

A way to resolve this would be to always generate the modules file from scratch, so that no diffing operations are required.

Tachi107 (changes requested):
Some feedback now that I have been able to look at the whole code :)

modules/CMakeLists.txt
mikomikotaishi: C++ modules require a later version of CMake, so activating modules should require the user has the correct version.

CMakeLists.txt
mikomikotaishi: I want to leave it as opt-in because C++ module support is compiler dependent. Most newer compiler versions offer it, but older ones do not. The build could break if we force it on older toolchains.

update_modules.py
mikomikotaishi: I am just used to writing `Optional[T]`, if `T | None` is better I can replace it

update_modules.py
mikomikotaishi: This was the suggestion @yhirose gave, and while I personally would prefer to have a static module for the purpose of having an obvious or clear module API (rather than behind a script) this is the approach I'm taking because the author does not maintain the module file or build system scripts.

That being said, this Python script is only supposed to be run by a GitHub Action script, so it does rely on the assumption of Git setup.

yhirose: @mikomikotaishi you probably misunderstood what I mentioned in the following comment. I said exactly the same thing as @Tachi107 mentioned above: "Could you please simply regenerate httplib.cppm from scratch with the latest httplib.h? It will make this script much simpler".
https://github.com/yhirose/cpp-httplib/pull/2291#discussion_r2600842632

mikomikotaishi: Yes, that is what I was saying, we would be generating it on demand.

update_modules.py
mikomikotaishi: Anyway I believe this requires a minimum of Python 3.5 now

update_modules.py
mikomikotaishi: There are some problems I am having with implementing this, for example having to insert the `#ifdef`s to add to the generated module all of the SSL classes. Do you have any suggestions for this?

yhirose: Sorry for the confusion... What I meant by 'from scratch' was to create httplib.cppm entirely, rather than just updating a portion of it. So we would like you not to use `get_git_diff`.

mikomikotaishi: Yes, I meant that I would be regenerating it entirely. My only problem is how to generate the symbols with the aforementioned.

sum01:
Is there a reason to try and add module support when the library itself is C++11?

Not to mention, Cmake v3.28 adds built-in module support, which isn't being used either.

Spixmaster:
This library being C++11 does not require programmes depending on it being C++11.

I support the modules.

mikomikotaishi:
Indeed, some people who use C++20 and onward may wish to consume the library with the module rather than the header. This is precisely what modules are designed to improve for compilations, as re-including the same large header in multiple locations massively increases build times whereas modules are built only once and then imported/linked.

sum01:
I understand you can use older standards on newer ones, but I mean trying to add support where it isn't "supported" seems unnecessary.

As for having to re-compile, why not use pre-compiled headers (Cmake v3.16 feature), or the compiled version of this library?

mikomikotaishi:
Precompiled headers are not a standard language feature, they are just a compiler trick. Modules provide a very specific API and very specific features for library encapsulation. For example, we may want to avoid including httplib's macros or avoid adding detail/implementation symbols into scope, which overwhelms autocomplete on IDEs with irrelevant symbols.

Even splitting the library into a header/source division still leaves a very large header that gets recompiled each time.

As for the question of "support", this specific wrapper style (a module which includes the header and then re-exports the symbols) is how most libraries built with C++11/14/17 provide module support (such as Boost, Poco, etc.) This is the simplest way to do it to my knowledge, and it's not a hastily tacked-on hack. You can see in other libraries this is how originally header-libraries are wrapped for modules.

mikomikotaishi:
Hi, sorry for the silence on this issue. I've been busy with an exam, but I'll return to this and send a solution today or tomorrow.

mikomikotaishi:
@yhirose Hi, I am sorry but I am completely lost as to how to actually generate a module correctly from the header. I was considering using `libclang` but I don't want to introduce a new dependency for a build script.

mikomikotaishi:
I personally would rather manually maintain the module file for the time being, but if someone more skilled at parsing C++ wants to take a crack at this please feel free.

yhirose:
@mikomikotaishi Thanks for your willingness. However, I am now thinking of closing this pull request for the maintenance reason I commented on before. I would like to confirm that you will commit to updating modules/httplib.cppm whenever new external symbols are added or existing symbols are removed/renamed by contributors. I am not going to wait for this update to happen. This update must be done by you before I release a new version. Otherwise, httplib.h will be shipped with an incompatible, out-of-date httplib.cppm. If you accept this responsibility from now on, I will reconsider it.

mikomikotaishi:
I will commit to this until I get a working script. (Seeing as you do ping contributors to request reviews, I hope it's not a steep ask for you to also ping me for symbol additions as well.) I'm sure it is possible - I have just had trouble with it. I can probably make a draft PR for that for the time being.

I am curious if you have any objection to me using `libclang` for this? It does introduce a new dependency but it can be made conditional. @Tachi107 seeing as you wrote this script originally, do you have any objections to that?

mikomikotaishi:
Actually, we could probably do this:
```cpp
module;

// include all included headers seen in <httplib.h>,
// to prevent them from being re-included in the export block
#include <...>

export module httplib;

export {
    #include "../httplib.h"
}
```
This way is far simpler, only problem is that this is sort of a hacky way to do it and also ends up exposing the `httplib::detail::*` symbols, which is preferably not exposed.

mikomikotaishi:
Until then, this is probably the best way to do it with no maintenance overhead, so I've gone ahead and pushed that, instead.

@yhirose could you re-review the current state and tell me your thoughts?

Tachi107:
On Mon Dec 29, 2025 at 4:40 AM CET, Miko wrote:
> I am curious if you have any objection to me using `libclang` for 
> this? It does introduce a new dependency but it can be made 
> conditional. @Tachi107 seeing as you wrote this script originally, do 
> you have any objections to that?

I did not wrote that script originally :)

But, for the record, adding one level of indentation for the whole 
split.py file just to have the code in a "main" function, with it being 
the only one, doesn't look like a cleanup ;)

Don't fall in the trap of following best pracises just for the sake of 
it.

On Mon Dec 29, 2025 at 4:46 AM CET, Miko wrote:
> Actually, we could probably do this:
> ```cpp
> module;
>
> // include all headers seen in <httplib.h>,
> // to prevent them from being re-included

Is this needed? Don't they have include guards already?

> export module httplib;
>
> export {
>     #include <httplib.h>
> }
> ```
> This way is far simpler, only problem is that this is sort of a hacky 
> way to do it and also ends up exposing the `httplib::detail::*` 
> symbols, which is preferably not exposed.

Couldn't you include the header produced by split.py, which contains 
less stuff?

Still, this long thread leaves me with a question: is all this mess 
worth it? How much does using C++20 modules instead of the header 
produced by split.py improve things? Because if the improvement is 
there it may make sense to pursue this. But if not, why bother?

mikomikotaishi:
Yes, we could probably use that header instead, and it probably is better to do that if we want to install the modules component from a package.

As for why I am even proposing adding modules, it is because it is cheaper to use them by building once and then importing, than including multiple times and then rebuilding the contents each time (even after splitting the contents the header still has a couple thousand lines I believe). As another person mentioned, one could use precompiled headers, but they are not an official language feature.

Also, because those headers have include guards I am including them first in the global module fragment before including httplib in the export block, to ensure that module httplib doesn't export a symbols like std::* symbols or structs from C POSIX headers, etc.

Tachi107:
On Mon Dec 29, 2025 at 1:17 PM CET, Miko wrote:
> As for why I am even proposing adding modules, it is because it is 
> cheaper to use them by building once and then importing, than 
> including multiple times and then rebuilding the contents each time. 

I'm aware of that. It is generally true. But in practise, how much does 
this improve the building time for this library?

mikomikotaishi:
I have never benchmarked it. But I can say confidently that expecting a compilation time improvement is reasonable in importing it 100 times over including it 100 times.

Even then, I don't think the current solution is a huge mess. I haven't changed the library itself or introduced peculiar artefacts into the header, the most extent this PR would have is
* one module file
* one CMake file
* eventually, a Python script to generate the contents correctly

yhirose:
If the following simple approach works, I’m fine with it. I don’t mind exposing the `httplib::detail::*` symbols, since `#including httplib.h` exposes the detail functions anyway. Could you please make the change? After that, I’ll take a look at it for review.

```cpp
module;

// include all included headers seen in <httplib.h>,
// to prevent them from being re-included in the export block
#include <...>

export module httplib;

export {
    #include "../httplib.h"
}
```

Also, please don’t forget to remove any files that are no longer needed, revert the changes in README.md, and add any necessary comments only to the CMake files. (I intentionally avoid mentioning any build systems in the README, as they are optional and not officially supported by cpp-httplib itself, but are kindly maintained by excellent contributors. Thanks for your understanding.)

mikomikotaishi:
I did make this change, but we have to re-open the PR for it to appear here. I will update the README as requested.

mikomikotaishi:
I've removed the changes to the README, but I did notice that the README mentions how to use the script to split the program into `.h`/`.cc` file, where would it be appropriate to document that the library contains support for modules features?

mikomikotaishi:
What if we put all detail namespaces inside an `#ifndef HTTPLIB_MODULE_HIDE_DETAILS` in the header, and defined that macro at the top of the global module fragment, so that the header exposes details but not the module? Essentially what I'm proposing:

`httplib.h`:
```cpp
namespace httplib {
    // ...

    #ifndef HTTPLIB_MODULE_HIDE_DETAILS
    namespace detail {
        // details here
    }
    #endif

    // ...
}
```

`httplib.cppm`:
```cpp
module;

#define HTTPLIB_MODULE_HIDE_DETAILS

// all httplib includes
#include <...>

export module httplib;

export {
    // Because HTTPLIB_MODULE_HIDE_DETAILS is defined, httplib::detail::* symbols aren't exported
    #include "../httplib.h"
}
```

@yhirose do let me know what you think, I think this wouldn't any more difficult to maintain moving forward, we would just have to ensure that `#ifndef HTTPLIB_MODULE_HIDE_DETAILS` appears before (or inside) every `namespace detail` declaration.

ChuanqiXu9:
https://chuanqixu9.github.io/c++/2025/12/30/C++20-Modules-Best-Practices.en.html#the-export-extern-c-style and https://chuanqixu9.github.io/c++/2025/12/30/C++20-Modules-Best-Practices.en.html#abi-breaking-style may be helpful for you.

mikomikotaishi:
Thanks, will review tomorrow

yhirose:
@mikomikotaishi Please don't change httplib.h or README.md at all. Instead, just focus on creating the simpler httplib.cppm, making the smallest possible changes and adding only the minimum necessary usage to the CMake-related files. Thank you.

mikomikotaishi:
Then the PR is basically in a finished state.

modules/CMakeLists.txt
Tachi107: If so, the proper minimum required version and/or policy behaviour should be declared in the relevant portions of the code.

I don’t use or maintain the CMake code of this project, so I’ll defer to @sum01’s feedback.

Tachi107:
> where would it be appropriate to document that the library contains support for modules features?

I think the build script itself is fine. Meson has `meson_options.txt`, which is self-documenting; CMake users can just grep for the available `options()`, or use some command I don't remember (or the GUI) to have CMake list them. As another example, the project supports system-wide installation, but since this is a build-system-specific feature its "documentation" is, well, the build system itself.

modules/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+add_library(httplib::module ALIAS httplib_module)
+
+# Installation
+install(TARGETS httplib_module
+    EXPORT ${PROJECT_NAME}Targets
+    LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
+    ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
+    RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
+    FILE_SET CXX_MODULES DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/httplib/modules
+)
```
Tachi107: I'm a bit confused: what are you installing here? The `httplib.cppm` file, or the built module interface?

In any case, creating a new `httplib_module` target feels wrong. The module should be added to the already-existing target, so that consumers of `httplib::httplib` automatically get the modules without having to chose one target or the others.

I know I asked already, but I guess this installes the `.cppm` file, and not the built module. It'd be nice to also install the built module interface, via the `CXX_MODULES_BMI` argument of the [`install(TARGETS)`](https://cmake.org/cmake/help/latest/command/install.html#targets) function.

Also, since this code will only run on recent CMake versions, all the various `ARCHIVE/LIBRARY/... DESTINATION` ceremony is unnecessary; they all have reasonable default values (and already use `GNUInstallDirs`).

Lastly, it might be easier to do all this in the top-level CMakeLists.txt file. Since pretty much all CMake code is already there (apart from tests), these 20-ish lines of code might as well be put there too.

modules/httplib.cppm
```diff
@@ -0,0 +1,142 @@
...
+//
+//  Copyright (c) 2025 Yuji Hirose. All rights reserved.
+//  MIT License
+//
+
+module;
+
+/*
+ * Include all headers in httplib.h, to prevent them from being re-exported
+ * in the export block.
```
Tachi107: This will easily get out-of-sync with the main header file. Can't these includes and defines be sourced and generated from the httplib.h file directly, during the configure or build step?

modules/httplib.cppm
```diff
@@ -0,0 +1,142 @@
...
+#endif
+
+export module httplib;
+
+/*
+ * Directly export the symbols in httplib.h. This is not the most elegant approach,
+ * but it will have to do until we figure out how to use a script to do it.
+ */
+export {
+    #include "../httplib.h"
```
Tachi107: As mentioned earlier, would it make sense to change this to the split `.h`, with less implementation details exposed, or does it require the full header?

In case the smaller header can be used instead, it might make sense to make this file a template `httplib.cppm.in` and configure it from CMake by including the header produced my `execute_process()` (which should really be changed to `add_custom_command(OUTPUT)`, but that's a separate issue) via `@VAR@` substitutions

mikomikotaishi: I would like to do this at some point with the smaller header, seeing as the the library has to be compiled anyway if the module has to be compiled.

modules/httplib.cppm
mikomikotaishi: I'll do this when I work on using the splitting script to use the reduced header.

modules/httplib.cppm
yhirose: I don't prefer including such a lengthy list of `#include ...`, because it's not easily maintainable... Is the following enough in this file?

```cpp
export module httplib;

export {
    #include "../httplib.h"
}
```

mikomikotaishi: The headers above were included first because it activates the header guard preventing it from being pulled in when the actual `httplib.h` header is included in the `export` block. If we didn't do that, the module would export stuff like `std::*` symbols or other headers pulled in.

This can just be copied in when we generate modules with Python instead

yhirose: This is too complex and appears difficult to maintain. I'm hesitant to merge this code unless it can be simplified and the code duplication is resolved. Thanks for your understanding.

mikomikotaishi: Weren't you originally content with a Python script to generate the module on demand?

yhirose: That’s fine. However, I wasn’t satisfied with the update_modules.py script at all. The script should generate the entire httplib.cppm from scratch, rather than updating the existing file using git diff. This would make the script much simpler.

Also, I would prefer not to commit httplib.cppm to this repository. It should be treated the same as httplib.h and httplib.cc, which are generated by split.py. In other words, httplib.cppm is just a build artifact.

mikomikotaishi: @yhirose It's now possible just doing `cmake -B build -S . -G Ninja -DHTTPLIB_COMPILE=ON -DHTTPLIB_BUILD_MODULES=ON`. I hope this is sufficiently simple.

Spixmaster:
Might I ask a question? Is the currennt approach good for programme packages? Think of https://aur.archlinux.org/packages?O=0&K=cpp-httplib .

Can users who have installed a package use modules and not with one package installed? This is very important because if one software depending on cpp-httplib uses modules and the other not one will need to have two different packages whose files would likely collide.

Alternatively formulated, can no modules and modules exist in parallel?

mikomikotaishi:
> Alternatively formulated, can no modules and modules exist in parallel?

I'm not sure if I entirely understand what you mean. Are you asking if it's possible to `#include <httplib/httplib.h>` and `import httplib;` in the same project? I believe the answer is yes.

mikomikotaishi:
@yhirose Hi, could you please review it again and tell me your thoughts on the current solution? I think this is very simple and should be to your liking.

yhirose:
@mikomikotaishi could you test it on three platforms (macos, ubuntu, Windows)?

mikomikotaishi:
> @mikomikotaishi could you test it on three platforms (macos, ubuntu, Windows)?

Things are building correctly for me and I'm able to use the module on MSVC, GCC and Clang, so I think things are good

yhirose:
Could you show me how you compile it with clang++, g++, and vc++? I'll try on my machines when I have time.

mikomikotaishi:
I'm relying on CMake to build it, because compiling libraries by manually calling the compiler rather than including it by header or using a build system is difficult and something I have little experience with. If you are interested in what I did:

`cmake -B build -S . -G Ninja -DHTTPLIB_COMPILE=ON -DHTTPLIB_BUILD_MODULES=ON`

This runs the `split.py` script and creates the files `httplib.h`, `httplib.cc`, and `httplib.cppm`. (Could we replace `.cc` with `.cpp` so that it aligns with `.cppm`? I'll leave this decision up to you, I personally prefer `.cpp` for the aforementioned reason but it's nothing dealbreaking, I don't particularly mind either way.) It configures to use Ninja as the build system, and indicates to the build system we wish to use modules. Of course, the compiler must support modules; see [arewemodulesyet/tools](https://arewemodulesyet.org/tools/) for versions.

`cmake --build build`

This performs the actual build. All of the compiler configuration is handled by the generated build system files.

CMakeLists.txt
```diff
@@ -247,6 +248,16 @@ if(HTTPLIB_COMPILE)
 			$<BUILD_INTERFACE:${_httplib_build_includedir}/httplib.h>
 			$<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}/httplib.h>
 	)
+	
+	# Add C++20 module support if requested
+	if(HTTPLIB_BUILD_MODULES)
+		target_sources(${PROJECT_NAME}
+			PUBLIC
+				FILE_SET CXX_MODULES FILES
```
Tachi107: The concerns about the CMake policy/minimum required version haven't been resolved still

CMakeLists.txt
```diff
@@ -365,6 +384,12 @@ if(HTTPLIB_INSTALL)
 	include(CPack)
 endif()
 
+if(HTTPLIB_BUILD_MODULES)
+	if(NOT HTTPLIB_COMPILE)
```
Tachi107:
```suggestion
if(HTTPLIB_BUILD_MODULES AND NOT HTTPLIB_COMPILE)
```

split.py
```diff
@@ -25,14 +24,14 @@ def main() -> None:
...
     header_name: str = f"/{lib_name}.h"
     source_name: str = f"/{lib_name}.{args.extension}"
     # get the input file
-    in_file: str = cur_dir + header_name
+    in_file: str = f"{cur_dir}{header_name}"
     # get the output file
-    h_out: str = args.out + header_name
-    cc_out: str = args.out + source_name
+    h_out: str = f"{args.out}{header_name}"
+    cc_out: str = f"{args.out}{source_name}"
```
Tachi107, line 34: For the sake of making reviews easier, please try to make your diff as small as possible, and leave unneeded changes like these ones outside this pull request (which is already taking a long time to get reviewed)

Tachi107 (changes requested):
Some more comments :)

Tachi107:
> (Could we replace `.cc` with `.cpp` so that it aligns with `.cppm`? I'll leave this decision up to you, I personally prefer `.cpp` for the aforementioned reason but it's nothing dealbreaking, I don't particularly mind either way.)

This would create issues for people who have hardcoded `.cc` in their build scripts, without any really advantage apart from a cosmetic difference :/

split.py
mikomikotaishi: Moving forward that's fine, but I do think there are tangible benefits to using fstrings over string concatenation (looking around online I often see such claims of performance benefits), and I do find it to be easier to read.

CMakeLists.txt
mikomikotaishi: OK, I've forced `HTTPLIB_BUILD_MODULES` to only be legal when CMake is 3.28 or higher.

Tachi107:
On Thu Jan 15, 2026 at 2:15 AM CET, Miko wrote:
> Moving forward that's fine, but I do think there are tangible benefits 
> to using fstrings over string concatenation (looking around online 
> I often see such claims of performance benefits), and I do find it to 
> be easier to read.

Generally speaking, I do agree with that. But also, this is really 
a quick&dirty script which hasn't been touched much during the years; 
moreover, f-strings are not supported in older versions of Python, which 
may create troubles for people trying to build this in old environments 
such as Debian 9, which is still supported (until 2027) via the ELTS 
service.

I don't think we should prevent updating our code to break compatibility 
with operating systems released in 2017 :)
But it's something which is always nice to keep in mind when making 
small changes that, while good practice, don't need to be done.

Tachi107:
On Thu Jan 15, 2026 at 2:22 AM CET, Miko wrote:
> OK, I've forced `HTTPLIB_BUILD_MODULES` to only be legal when CMake is 3.28 or higher.

Thanks, but this is not what I meant :/

My comment was about the suspicion that the CMake build script should 
opt in to some new policy behaviour with either cmake_policy or 
cmake_minimum_required, but I'm just not sure. I'll wait for someone 
actually maintaining the CMake support for cpp-httplib to comment on 
this matter. Or, if you're sure it's not needed, just tell me!

In any case, those checks aren't really needed because the build would 
fail anyway upon invocation of an unrecognized CMake command by older 
CMake installations.

mikomikotaishi:
OK, I'll instead put it behind an include call

mikomikotaishi:
@yhirose Anything else in need of addressing?

Tachi107:
> OK, I'll instead put it behind an include call

I think I haven't been understood :/

Please look at the documentation of `cmake_policy` and `cmake_minimum_required` to get what I meant :)

Spixmaster:
Cannot you (@Tachi107) just give a diff instead of letting him guess what you want? This is not constructive.

@mikomikotaishi, is very responsive and puts a lot of effort into it for more then five weeks.

Tachi107:
On Wed Jan 21, 2026 at 9:34 PM CET, Matheus Gabriel Werny wrote:
> Cannot you ***@***.***) just give a diff instead of letting him guess what you want? This is not constructive.

I would! As mentioned before, I am myself not sure this is needed, and 
has already asked people maintaning the CMake build system to comment. 
While I have some experience with CMake myself, here in the cpp-httplib 
project I only maintain the Meson build scripts. Hence, what I'm doing 
here is just providing some feedback, but someone else should have the 
final say.

mikomikotaishi:
Hi, I'm sorry but I don't have very much knowledge of CMake and its features or how to correctly apply the desired solution. I would have thought an include would be sufficient to prevent version incompatibilities, but if not a diff would be highly appreciated, I am willing to take suggestions of anyone more familiar with CMake than I am.

sum01:
I'd guess he was talking about adding `cmake_minimum_required(3.28)` into the modules file and setting [CMP0155](https://cmake.org/cmake/help/latest/policy/CMP0155.html#policy:CMP0155) to the new policy?

Tachi107:
> I'd guess he was talking about adding cmake_minimum_required(3.28) into the modules file and setting CMP0155 to the new policy?
> 

Yes, something like that. Note though that setting the minimum version to 3.28 already implies using the new policy behaviour.

mikomikotaishi:
@yhirose Could you please review the current changes and give your feedback on the current state?

yhirose:
Thank you for all your hard work! I just tried to build with g++-15 installed from homebrew on macos. I made the following test file.

```cpp
// test-cpp-modules.cc
import httplib;
using namespace httplib;

int main() {
  Server svr;
  svr.Get("/hi", [](const Request & /*req*/, Response &res) {
  });
  svr.listen("0.0.0.0", 8080);
}
```

Then, I did the following:

```bash
> g++-15 -std=c++20 -fmodules-ts -x c++ -c httplib.cppm
> g++-15 -std=c++20 -fmodules-ts test-cpp-modules.cc httplib.o -o test-cpp-modules
```

But when I run the second command to make the executable file, the compiler gave me errors below...

```
[mikomikotaishi-master] ~/Projects/cpp-httplib/test$ g++-15 -std=c++20 -fmodules-ts test-cpp-modules.cc httplib.o -o test-cpp-modules
In file included from /opt/homebrew/Cellar/gcc/15.2.0/include/c++/15/functional:61,
                 from httplib.cppm:106,
of module httplib, imported at test-cpp-modules.cc:5:
/opt/homebrew/Cellar/gcc/15.2.0/include/c++/15/bits/std_function.h: In instantiation of 'static bool std::_Function_handler<_Res(_ArgTypes ...), _Functor>::_M_manager(std::_Any_data&, const std::_Any_data&, std::_Manager_operation) [with _Res = void; _Functor = main()::<lambda(const httplib::Request@httplib&, httplib::Response@httplib&)>; _ArgTypes = {const httplib::Request@httplib&, httplib::Response@httplib&}]':
/opt/homebrew/Cellar/gcc/15.2.0/include/c++/15/bits/std_function.h:454:21:   required from 'std::function<_Res(_ArgTypes ...)>::function(_Functor&&) [with _Functor = main()::<lambda(const httplib::Request@httplib&, httplib::Response@httplib&)>; _Constraints = void; _Res = void; _ArgTypes = {const httplib::Request@httplib&, httplib::Response@httplib&}]'
  454 |               _M_manager = &_My_handler::_M_manager;
      |                            ^~~~~~~~~~~~~~~~~~~~~~~~
test-cpp-modules.cc:11:18:   required from here
   11 |   svr.Get("/hi", [](const Request & /*req*/, Response &res) {
      |                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   12 |     res.set_content("Hello World!", "text/plain");
      |     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   13 |   });
      |   ~
/opt/homebrew/Cellar/gcc/15.2.0/include/c++/15/bits/std_function.h:276:53: error: must '#include <typeinfo>' before using 'typeid'
   69 |    */
  +++ |+#include <typeinfo>
   70 |   template<typename _Tp>
......
  276 |             __dest._M_access<const type_info*>() = &typeid(_Functor);
      |                                                     ^~~~~~~~~~~~~~~~
/opt/homebrew/Cellar/gcc/15.2.0/include/c++/15/bits/std_function.h: In instantiation of 'static void std::_Function_base::_Base_manager<_Functor>::_M_create(std::_Any_data&, _Fn&&, std::true_type) [with _Fn = main()::<lambda(const httplib::Request@httplib&, httplib::Response@httplib&)>; _Functor = main()::<lambda(const httplib::Request@httplib&, httplib::Response@httplib&)>; std::true_type = std::true_type]':
/opt/homebrew/Cellar/gcc/15.2.0/include/c++/15/bits/std_function.h:217:15:   required from 'static void std::_Function_base::_Base_manager<_Functor>::_M_init_functor(std::_Any_data&, _Fn&&) [with _Fn = main()::<lambda(const httplib::Request@httplib&, httplib::Response@httplib&)>; _Functor = main()::<lambda(const httplib::Request@httplib&, httplib::Response@httplib&)>]'
  217 |             _M_create(__functor, std::forward<_Fn>(__f), _Local_storage());
      |             ~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/opt/homebrew/Cellar/gcc/15.2.0/include/c++/15/bits/std_function.h:451:36:   required from 'std::function<_Res(_ArgTypes ...)>::function(_Functor&&) [with _Functor = main()::<lambda(const httplib::Request@httplib&, httplib::Response@httplib&)>; _Constraints = void; _Res = void; _ArgTypes = {const httplib::Request@httplib&, httplib::Response@httplib&}]'
  451 |               _My_handler::_M_init_functor(_M_functor,
      |               ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~
  452 |                                            std::forward<_Functor>(__f));
      |                                            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
test-cpp-modules.cc:11:18:   required from here
   11 |   svr.Get("/hi", [](const Request & /*req*/, Response &res) {
      |                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   12 |     res.set_content("Hello World!", "text/plain");
      |     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   13 |   });
      |   ~
/opt/homebrew/Cellar/gcc/15.2.0/include/c++/15/bits/std_function.h:154:13: error: no matching function for call to 'operator new(sizetype, void*)'
  154 |             ::new (__dest._M_access()) _Functor(std::forward<_Fn>(__f));
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/opt/homebrew/Cellar/gcc/15.2.0/include/c++/15/bits/std_function.h:154:13: note: there are 2 candidates
In file included from /opt/homebrew/Cellar/gcc/15.2.0/include/c++/15/bits/stl_iterator.h:75,
                 from /opt/homebrew/Cellar/gcc/15.2.0/include/c++/15/bits/stl_algobase.h:67,
                 from /opt/homebrew/Cellar/gcc/15.2.0/include/c++/15/algorithm:62,
                 from httplib.cppm:94,
of module httplib, imported at test-cpp-modules.cc:5:
/opt/homebrew/Cellar/gcc/15.2.0/include/c++/15/new:137:26: note: candidate 1: 'void* operator new(std::size_t)'
  137 | _GLIBCXX_NODISCARD void* operator new(std::size_t)
      |                          ^~~~~~~~
/opt/homebrew/Cellar/gcc/15.2.0/include/c++/15/new:137:26: note: candidate expects 1 argument, 2 provided
/opt/homebrew/Cellar/gcc/15.2.0/include/c++/15/new:168:26: note: candidate 2: 'void* operator new(std::size_t, std::align_val_t)'
  168 | _GLIBCXX_NODISCARD void* operator new(std::size_t, std::align_val_t)
      |                          ^~~~~~~~
/opt/homebrew/Cellar/gcc/15.2.0/include/c++/15/new:168:26: note: no known conversion for argument 2 from 'void*' to 'std::align_val_t'
```

Do you have any clue?

mikomikotaishi:
It looks like the problem is that GCC has a broken module implementation. Adding
```cpp
#include <new>
#include <typeinfo>
```
inside your `test-cpp-modules.cc` file fixes it on GCC.

In my experience there is no need to include those headers on Clang; that's why I'm convinced it's a GCC bug.

yhirose:
If I put the two lines before `import httplib`, I got the followings:

```
Undefined symbols for architecture arm64:
  "httplib::detail::BufferStream@httplib::read(char*, unsigned long)", referenced from:
      vtable for httplib::detail::BufferStream@httplib in httplib.o
  "httplib::detail::BufferStream@httplib::write(char const*, unsigned long)", referenced from:
      vtable for httplib::detail::BufferStream@httplib in httplib.o
  "httplib::detail::close_socket@httplib(int)", referenced from:
      httplib::ClientConnection@httplib::~ClientConnection() in httplib.o
      httplib::ClientConnection@httplib::~ClientConnection() in httplib.o
  "httplib::detail::nocompressor@httplib::compress(char const*, unsigned long, bool, std::function<bool (char const*, unsigned long)>)", referenced from:
      vtable for httplib::detail::nocompressor@httplib in httplib.o
  "httplib::ClientImpl@httplib::StreamHandle::read(char*, unsigned long)", referenced from:
      httplib::stream::Result@httplib::next() in httplib.o
  "httplib::ClientImpl@httplib::shutdown_ssl(httplib::ClientImpl@httplib::Socket&, bool)", referenced from:
      vtable for httplib::ClientImpl@httplib in httplib.o
  "httplib::ClientImpl@httplib::process_socket(httplib::ClientImpl@httplib::Socket const&, std::chrono::time_point<std::chrono::_V2::steady_clock, std::chrono::duration<long long, std::ratio<1l, 1000000000l>>>, std::function<bool (httplib::Stream@httplib&)>)", referenced from:
      vtable for httplib::ClientImpl@httplib in httplib.o
  "httplib::ClientImpl@httplib::ensure_socket_connection(httplib::ClientImpl@httplib::Socket&, httplib::Error@httplib&)", referenced from:
      vtable for httplib::ClientImpl@httplib in httplib.o
  "httplib::ClientImpl@httplib::create_and_connect_socket(httplib::ClientImpl@httplib::Socket&, httplib::Error@httplib&)", referenced from:
      vtable for httplib::ClientImpl@httplib in httplib.o
  "httplib::ClientImpl@httplib::~ClientImpl()", referenced from:
      vtable for httplib::ClientImpl@httplib in httplib.o
  "httplib::ClientImpl@httplib::~ClientImpl()", referenced from:
      vtable for httplib::ClientImpl@httplib in httplib.o
  "httplib::Client@httplib::open_stream(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>> const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>> const&, std::multimap<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>, std::less<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>>, std::allocator<std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>> const, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>>>> const&, std::unordered_multimap<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>, httplib::detail::case_ignore::hash@httplib, httplib::detail::case_ignore::equal_to@httplib, std::allocator<std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>> const, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>>>> const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>> const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>> const&)", referenced from:
      httplib::stream::Result@httplib httplib::stream::Get@httplib<httplib::Client@httplib>(httplib::Client@httplib&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>> const&, std::unordered_multimap<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>, httplib::detail::case_ignore::hash@httplib, httplib::detail::case_ignore::equal_to@httplib, std::allocator<std::pair<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>> const, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>>>> const&, unsigned long) in httplib.o
  "httplib::Client@httplib::stop()", referenced from:
      httplib::sse::SSEClient@httplib::stop() in httplib.o
  "httplib::Server@httplib::process_and_close_socket(int)", referenced from:
      vtable for httplib::Server@httplib in httplib.o
  "httplib::Server@httplib::Get(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>> const&, std::function<void (httplib::Request@httplib const&, httplib::Response@httplib&)>)", referenced from:
      _main in cc5namgh.o
  "httplib::Server@httplib::listen(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>> const&, int, int)", referenced from:
      _main in cc5namgh.o
  "httplib::Server@httplib::Server()", referenced from:
      _main in cc5namgh.o
  "httplib::Server@httplib::~Server()", referenced from:
      vtable for httplib::Server@httplib in httplib.o
  "httplib::Server@httplib::~Server()", referenced from:
      _main in cc5namgh.o
      _main in cc5namgh.o
      vtable for httplib::Server@httplib in httplib.o
  "httplib::Response@httplib::set_content(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>&&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>> const&)", referenced from:
      main::'lambda'(httplib::Request@httplib const&, httplib::Response@httplib&)::operator()(httplib::Request@httplib const&, httplib::Response@httplib&) const in cc5namgh.o
  "httplib::detail::BufferStream@httplib::is_readable() const", referenced from:
      vtable for httplib::detail::BufferStream@httplib in httplib.o
  "httplib::detail::BufferStream@httplib::wait_readable() const", referenced from:
      vtable for httplib::detail::BufferStream@httplib in httplib.o
  "httplib::detail::BufferStream@httplib::wait_writable() const", referenced from:
      vtable for httplib::detail::BufferStream@httplib in httplib.o
  "httplib::detail::BufferStream@httplib::get_local_ip_and_port(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>&, int&) const", referenced from:
      vtable for httplib::detail::BufferStream@httplib in httplib.o
  "httplib::detail::BufferStream@httplib::get_remote_ip_and_port(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>>&, int&) const", referenced from:
      vtable for httplib::detail::BufferStream@httplib in httplib.o
  "httplib::detail::BufferStream@httplib::socket() const", referenced from:
      vtable for httplib::detail::BufferStream@httplib in httplib.o
  "httplib::detail::BufferStream@httplib::duration() const", referenced from:
      vtable for httplib::detail::BufferStream@httplib in httplib.o
  "httplib::detail::RegexMatcher@httplib::match(httplib::Request@httplib&) const", referenced from:
      vtable for httplib::detail::RegexMatcher@httplib in httplib.o
  "httplib::detail::PathParamsMatcher@httplib::match(httplib::Request@httplib&) const", referenced from:
      vtable for httplib::detail::PathParamsMatcher@httplib in httplib.o
  "httplib::ClientImpl@httplib::is_ssl() const", referenced from:
      vtable for httplib::ClientImpl@httplib in httplib.o
  "httplib::ClientImpl@httplib::is_valid() const", referenced from:
      vtable for httplib::ClientImpl@httplib in httplib.o
  "httplib::Server@httplib::is_valid() const", referenced from:
      vtable for httplib::Server@httplib in httplib.o
  "httplib::Response@httplib::has_header(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>> const&) const", referenced from:
      httplib::stream::Result@httplib::has_header(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>> const&) const in httplib.o
  "httplib::Response@httplib::get_header_value(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>> const&, char const*, unsigned long) const", referenced from:
      httplib::stream::Result@httplib::get_header_value(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char>> const&, char const*) const in httplib.o
ld: symbol(s) not found for architecture arm64
collect2: error: ld returned 1 exit status
```

When the lines are added after `import httplib;`, I got this error. So the workaround didn't work...

```
In file included from /opt/homebrew/Cellar/gcc/15.2.0/include/c++/15/new:43,
                 from test-cpp-modules.cc:7:
/opt/homebrew/Cellar/gcc/15.2.0/include/c++/15/bits/exception.h:64:36: internal compiler error: in finish_member_declaration, at cp/semantics.cc:4210
   64 |     exception() _GLIBCXX_NOTHROW { }
      |                                    ^
Please submit a full bug report, with preprocessed source (by using -freport-bug).
See <https://github.com/Homebrew/homebrew-core/issues> for instructions.
```

yhirose:
Also I confirmed that the latest Apple Clang in Xcode 26.2 doesn't work. So I don't know how I can apply this C++ module support on my mac...

mikomikotaishi:
Did you make sure to link back to the source file, which contains the implementations of most of these listed symbols? I had a similar error until I realised I had not done the aforementioned step.

yhirose:
Could you please try the following with g++-15 on your machine to see if you get the same error or not? If you get the same error, could you show me how you can fix the problem? Thanks.

> g++-15 -std=c++20 -fmodules-ts -x c++ -c httplib.cppm
> g++-15 -std=c++20 -fmodules-ts test-cpp-modules.cc httplib.o -o test-cpp-modules

```cpp
// test-cpp-modules.cc
import httplib;

#include <new>
#include <typeinfo>

using namespace httplib;

int main() {
  Server svr;
  svr.Get("/hi", [](const Request & /*req*/, Response &res) {
  });
  svr.listen("0.0.0.0", 8080);
}
```

mikomikotaishi:
This seemed to work for me:
```sh
$ g++ -std=c++20 -c httplib.cc -o httplib.o
$ g++ -std=c++20 -fmodules-ts -x c++ -c httplib.cppm -o httplib_module.o
$ g++ -std=c++20 -fmodules-ts test.cpp httplib.o httplib_module.o -o test
```
Also, if you are getting a compiler crash or internal error on GCC, move the `#include`s before `import httplib;`.

yhirose:
Sorry, but the following still gives me the same exact errors regardless whether `#include` is located before or after `import httplib;`. Did you really try what I asked you to do?

```
g++-15 -std=c++20 -c httplib.cc -o httplib.o
g++-15 -std=c++20 -fmodules-ts -x c++ -c httplib.cppm -o httplib_module.o
g++-15 -std=c++20 -fmodules-ts test-cpp-modules.cc httplib.o httplib_module.o -o test-cpp-modules
```

mikomikotaishi:
Have you cloned the most recent revision? Specifically, the generated module should have `export extern "C++"`:
```cpp
export extern "C++" {
    #include "httplib.h"
}
```
Lacking `extern "C++"` seems to cause a similar error to what you showed, but when it is present it builds successfully. It is specifically for marking the module as an interface to the header, i.e. symbols in `module httplib` refer to the same symbols you would get from including `<httplib.h>`

yhirose:
I am using the latest `split.py` in this pull request to generate `httplib.cppm`, and it includes `extern "C++"` at the end.

```cpp
...

export module httplib;

export {
    #include "httplib.h"
}
```

mikomikotaishi:
I'm confused what you mean, you say it includes `extern "C++"` but the snippet you posted doesn't have `extern "C++"`.

yhirose:
You are right. I'll clone the latest. Sorry for the confusion.

yhirose:
I finally got it to build with C++ modules. Thanks for your help. (`#include <new>` and `#include <typeinfo>` should be added before `import httplib;`.)

```cpp
#include <new>
#include <typeinfo>

import httplib;

using namespace httplib;

int main() {
  Server svr;

  svr.Get("/hi", [](const Request & /*req*/, Response &res) {
  });

  svr.listen("0.0.0.0", 8080);
}
```

mikomikotaishi:
Fantastic!

Is there anything left to do before we can finish this PR?

yhirose:
I just took a look at the code, and I realized that the C++ module generation isn't related to split.py. It would be better to move this functionality to a separate file, such as gen_cpp_module.py. So, could you please move the module generation code to gen_cpp_module.py and revert the changes to split.py? That's all!

mikomikotaishi:
But I think a lot of the logic that is used to construct the module is dependent on the code splitting script, and it's simpler to ensure the module generation takes place after the header is split into header and source, etc. I just personally don't think that there is much to gain from dividing the logic, especially if module generation is only allowed when the library is compiled, etc.

yhirose:
It's fine with me if `gen_cpp_module.py` gets longer. My main intention is to keep this additional feature separate from the official files that I maintain, such as `split.py`. Thanks for your understanding.

mikomikotaishi:
@yhirose I split the script into two files and tested it again with both manually calling the script and doing it from CMake, and it looks good to me.

split.py
```diff
@@ -1,21 +1,20 @@
 #!/usr/bin/env python3
 
-"""This script splits httplib.h into .h and .cc parts."""
+"""This script splits httplib.h into .h, .cc, and .cppm parts."""
```
yhirose: Could you remove `.cppm`?

mikomikotaishi: Fixed it, apologies for not noticing this sooner.

yhirose:
Looks good to me. Thanks for all your hard work!

PR #2295, opened 2025-12-09
Update split.py file

mikomikotaishi:
This pull request updates the `split.py` file as requested, to begin preparations for generating a module file from a header (#2291).


yhirose:
Thanks for the contribution!

PR #2297, opened 2025-12-09
Use std::move for content_provider in adapter

Skylion007:
Missing std::move in ContentProvider ctor. No reason to copy the content for the adapter when the move ctor will do.


yhirose:
Thanks!

PR #2298, opened 2025-12-09
Use std::move for boundary in set_boundary method

Skylion007:
Add missing std::move. `boundary` is bound by name inside the function scope so becomes an lvalue. Needs std::move to explicitly have it be an xvalue again.


yhirose:
Thanks!

PR #2300, opened 2025-12-10
Replace httpbin.org with httpcan.org

jfsimoneau:
Because of the recent httpbin.org outages that are failing the proxy tests, I replaced the calls to a rust-based clone called httpcan.org. Another contender is the Go-based httpbingo.org service, but they do not support Brotli or SHA-512 digests. The only difference is in the auth digest, they return the algorithm as part of the response, and I couldn't make auth-int work for some reason.

### DIFF
--- a/test/test_proxy.cc
+++ b/test/test_proxy.cc
@@ -182,7 +182,6 @@ template <typename T> void DigestAuthTestFromHTTPWatch(T &cli) {
         "/digest-auth/auth/hello/world/MD5",
         "/digest-auth/auth/hello/world/SHA-256",
         "/digest-auth/auth/hello/world/SHA-512",
-        // "/digest-auth/auth-int/hello/world/MD5",
     };
 
     cli.set_digest_auth("hello", "world");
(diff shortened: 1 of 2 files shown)


yhirose:
Could you please add more information in the description section?

yhirose:
Great! Could you also take a look at `DigestAuthTest.FromHTTPWatch_Online` error? Thanks a lot!

test/test_proxy.cc
```diff
@@ -182,15 +182,18 @@ template <typename T> void DigestAuthTestFromHTTPWatch(T &cli) {
         "/digest-auth/auth/hello/world/MD5",
         "/digest-auth/auth/hello/world/SHA-256",
         "/digest-auth/auth/hello/world/SHA-512",
-        "/digest-auth/auth-int/hello/world/MD5",
+        // "/digest-auth/auth-int/hello/world/MD5",
```
yhirose: Is there any reason why this line is disabled?

jfsimoneau: I couldn't make `auth-int` work at all. Even `curl` fails, so I suppose this is a limitation of `httpcan.org`.

jfsimoneau:
> Great! Could you also take a look at `DigestAuthTest.FromHTTPWatch_Online` error? Thanks a lot!

Sure. It seems to be the same limitation the original `httpbin.org` limitation where the username is ignored in the digest comparison. `httpcan.org` doesn't have that limitation, but does not support `auth-int`. I will use `httpcan.org` for the bad username test but keep `nghttp2.org` for the rest of the tests.

yhirose:
Thanks! I have merged it.

PR #2306, opened 2025-12-11
Use move semantics for auth key and value

Skylion007:


yhirose:
Thanks for the improvements!

PR #2319, opened 2026-01-05
Prevent redefinition of ssize_t

seragh:
On Windows cpp-httplib defines ssize_t, therefore applications needing to define ssize_t for their own needs or are using libraries that do require a means to avoid a possible incompatible redefinition.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -205,7 +205,10 @@
 
 #pragma comment(lib, "ws2_32.lib")
 
+#ifndef _SSIZE_T_DEFINED
 using ssize_t = __int64;
+#define _SSIZE_T_DEFINED
+#endif
 #endif // _MSC_VER
 
 #ifndef S_ISREG


httplib.h
```diff
@@ -205,7 +205,11 @@
 
 #pragma comment(lib, "ws2_32.lib")
 
+#ifndef HAVE_SSIZE_T
```
yhirose: Could you please change it to `_SSIZE_T_DEFINED`, as this area is Windows- specific?

seragh: `HAVE_SSIZE_T` is used by other libs for this very purpose, see wx/types.h for example. Tho for my purpose `_SSIZE_T_DEFINED` would work as well if you have a strong preference.

yhirose: Yes, I would like to use `_SSIZE_T_DEFINED` instead.

seragh: Changed to `_SSIZE_T_DEFINED`

yhirose:
@seragh thanks for your contribution!

PR #2320, opened 2026-01-08
Fix path params with empty values

smarttowel:
This PR fix path params with empty values.

Before:
`/path?key1&key2` will become `/path?key1=&key2=`

Now we just ignore empty values and leave keys only.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -6178,8 +6178,10 @@ inline std::string params_to_query_str(const Params &params) {
   for (auto it = params.begin(); it != params.end(); ++it) {
     if (it != params.begin()) { query += '&'; }
     query += encode_query_component(it->first);
-    query += '=';
-    query += encode_query_component(it->second);
+    if (!it->second.empty()) {
+      query += '=';
+      query += encode_query_component(it->second);
+    }
   }
   return query;
 }


yhirose:
@smarttowel thanks for the pull request. Here is the review from Copilot.

## Copilot

I'll help you review this pull request. Let me gather more information about it. Looking at this PR, I can see a potential issue with the proposed fix. Let me explain: 

## The Problem

The PR description states:
- **Before**: `/path?key1&key2` becomes `/path?key1=&key2=`
- **After**: Keys with empty values are left without the `=` sign

## The Issue

**This change may not be correct** for the following reason:

In URL query strings, there's an important distinction:
- `? key1&key2` - keys without values (no `=` sign)
- `?key1=&key2=` - keys with **empty string** values (has `=` sign)

These two formats have **different semantic meanings**:
1. `key1` with no `=` typically means the parameter is present as a flag/boolean
2. `key1=` with `=` means the parameter has an explicitly empty string value

## The Code Change

The PR modifies the `params_to_query_str` function to skip adding `=` and the value when `it->second.empty()` is true. However, this means:
- If someone intentionally sets a parameter to an empty string (`"key" -> ""`), the resulting query string will be `?key` instead of `?key=`
- This changes the semantics and could break applications expecting `key=` format

## Recommendation

The current behavior (adding `=` even for empty values) is likely **correct** according to standard URL encoding practices.  If the goal is to support parameters without values, the library should distinguish between: 
- Parameters with no value (don't add to Params at all, or use a special marker)
- Parameters with empty string values (current behavior is correct)

pavmk:
Hi @yhirose,

I found this new behavior when migrate to latest version of cpp-httplib. As I understand, keys without `=` is not good practice, but acceptable by specs. Anyway, it should be interpreted by server.

yhirose:
You are right. I checked with JavaScript on browser and Python. Both put `=` for an empty string. I'll follow them. Thanks!

image
image

Issue #2321, opened 2026-01-09
Tests fail to build with gcc 16


opoplawski:
Fedora rawhide has updated to gcc 16.  We are now seeing:
```
cd /builddir/build/BUILD/cpp-httplib-0.26.0-build/cpp-httplib-0.26.0/redhat-linux-build/test/fuzzing && /usr/bin/g++ -DCPPHTTPLIB_BROTLI_SUPPORT -DCPPHTTPLIB_OPENSSL_SUPPORT -DCPPHTTPLIB_USE_NON_BLOCKING_GETADDRINFO -DCPPHTTPLIB_ZLIB_SUPPORT -isystem /builddir/build/BUILD/cpp-httplib-0.26.0-build/cpp-httplib-0.26.0 -O2 -flto=auto -ffat-lto-objects -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-U_FORTIFY_SOURCE,-D_FORTIFY_SOURCE=3 -Wp,-D_GLIBCXX_ASSERTIONS -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -fstack-protector-strong -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1  -m64 -march=x86-64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -mtls-dialect=gnu2 -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -MD -MT test/fuzzing/CMakeFiles/httplib-test-fuzz.dir/standalone_fuzz_target_runner.cpp.o -MF CMakeFiles/httplib-test-fuzz.dir/standalone_fuzz_target_runner.cpp.o.d -o CMakeFiles/httplib-test-fuzz.dir/standalone_fuzz_target_runner.cpp.o -c /builddir/build/BUILD/cpp-httplib-0.26.0-build/cpp-httplib-0.26.0/test/fuzzing/standalone_fuzz_target_runner.cpp
/builddir/build/BUILD/cpp-httplib-0.26.0-build/cpp-httplib-0.26.0/test/test.cc: In member function ‘virtual void EncodeQueryParamTest_TestUTF8Characters_Test::TestBody()’:
[...]

PR #2326, opened 2026-01-12
Stop using u8 string literals for more recent compilers

pemensik:
GCC 15 will fail with u8 string literals.

Alternative to #2321.

### DIFF
--- a/test/test.cc
+++ b/test/test.cc
@@ -42,6 +42,12 @@
 #define SERVER_ENCRYPTED_PRIVATE_KEY_FILE "./key_encrypted.pem"
 #define SERVER_ENCRYPTED_PRIVATE_KEY_PASS "test123!"
 
+#if __cplusplus >= 201402L
+#define U8(str) str
+#else
+#define U8(str) u8##str
+#endif
+
 using namespace std;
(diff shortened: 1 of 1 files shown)


yhirose:
@pemensik  Thank you for the pull request. However, this issue was clearly described and resolved in #2321, which detailed the technical cause and included reproducible build errors. The solution proposed and accepted in that issue is sufficient for GCC 16 compatibility.

Since this pull request (#2326) overlaps with the changes already made in #2321, I do not see a need to merge it additionally and will be rejecting this PR.

Thank you very much for your suggestion and cooperation.

pemensik:
Hmm, I created alternative one, because I think reinterpret casts are unnecessary and ugly. Also, it is still not merged. I am not sure then how accepted it is considered to be. But okay, I admit that patch makes the test compiling again, I have tested it first.

yhirose:
@pemensik I have actually implemented this in a different way at https://github.com/yhirose/cpp-httplib/commit/aaf7ada8f1cb5acbbc7aa8bf57997ff1684a073a. This is a mix of these two. Thanks!

Issue #2217, opened 2025-08-22
SSL verification failure on redirect


JakubHazik:
**Problem**
When using the `SSLClient`  class, I call `set_ca_cert_store` to configure a custom certificate store.
However, if a request is redirected to another domain, the library internally creates a new `SSLClient` instance, and the configured certificate store is not propagated to the new client.

This causes SSL verification issues after redirects.

The same issue can also be observed when using the `Client `class if the request is redirected multiple times (e.g., two or more redirects).

You can use this test for reproduce
```cpp
TEST(SSLClientRedirectTest, CertFile) {
  SSLServer ssl_svr1(SERVER_CERT2_FILE, SERVER_PRIVATE_KEY_FILE);
  ASSERT_TRUE(ssl_svr1.is_valid());
  ssl_svr1.Get("/index", [&](const Request &, Response &res) {
    res.set_redirect("https://127.0.0.1:1235/index");
    ssl_svr1.stop();
  });

  SSLServer ssl_svr2(SERVER_CERT2_FILE, SERVER_PRIVATE_KEY_FILE);
  ASSERT_TRUE(ssl_svr2.is_valid());
  ssl_svr2.Get("/index", [&](const Request &, Response &res) {
    res.set_content("test", "text/plain");
    ssl_svr2.stop();
  });

  thread t = thread([&]() { ASSERT_TRUE(ssl_svr1.listen("127.0.0.1", PORT)); });
  thread t2 = thread([&]() { ASSERT_TRUE(ssl_svr2.listen("127.0.0.1", 1235)); });
  auto se = detail::scope_exit([&] {
    t2.join();
    t.join();
    ASSERT_FALSE(ssl_svr1.is_running());
  });

  ssl_svr1.wait_until_ready();
  ssl_svr2.wait_until_ready();

  SSLClient cli("127.0.0.1", PORT);
  std::string cert;
  read_file(SERVER_CERT2_FILE, cert);
  cli.load_ca_cert_store(cert.c_str(), cert.size());
  cli.enable_server_certificate_verification(true);
  cli.set_follow_location(true);
  cli.set_connection_timeout(30);

  auto res = cli.Get("/index");
  ASSERT_TRUE(res);
  ASSERT_EQ(StatusCode::OK_200, res->status);
}
```

PR #2335, opened 2026-01-22
Fix set_ca_cert_store() to skip system certs like set_ca_cert_path()

solarispika:
## Problem

`set_ca_cert_path()` and `set_ca_cert_store()` behave inconsistently regarding system certificate loading:

| Method | Skips system certs |
|--------|-------------------|
| `set_ca_cert_path()` | Yes |
| `set_ca_cert_store()` | No |

Both APIs conceptually do the same thing: "use these CA certs for verification." However, when using `set_ca_cert_store()` for certificate pinning, system certs are still loaded in `load_certs()`, defeating the purpose of pinning.

## Root Cause

In `load_certs()`:

```cpp
if (!ca_cert_file_path_.empty()) {
  // load from file - skips else branch
} else if (!ca_cert_dir_path_.empty()) {
  // load from dir - skips else branch
} else {
  // load system certs - ALWAYS runs when using set_ca_cert_store()
}
```

`set_ca_cert_path()` sets `ca_cert_file_path_`, taking the first branch and skipping system cert loading.

`set_ca_cert_store()` sets `ca_cert_store_` (since #2217 fix), but `load_certs()` doesn't check it, so it falls through to the else branch where system certs are added to the user's custom store.

## Fix

Check `ca_cert_store_` in `load_certs()`:

```cpp
    } else if (!ca_cert_store_) {  // Was: } else {
      // load system certs
    }
```

This makes `set_ca_cert_store()` behave consistently with `set_ca_cert_path()`.

## Test

Added `SSLClientTest.SetCaCertStoreSkipsSystemCerts_Online` to verify system certs are not loaded when custom store is set.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -13093,7 +13093,7 @@ inline bool SSLClient::load_certs() {
         last_openssl_error_ = ERR_get_error();
         ret = false;
       }
-    } else {
+    } else if (!ca_cert_store_) {
       auto loaded = false;
 #ifdef _WIN32
       loaded =
--- a/test/test.cc
+++ b/test/test.cc
@@ -9682,6 +9682,27 @@ TEST(SSLClientRedirectTest, CertFile) {
   ASSERT_EQ(StatusCode::OK_200, res->status);
 }
 
+// Test that set_ca_cert_store() skips system certs (consistent with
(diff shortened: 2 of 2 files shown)


yhirose:
@solarispika thanks for finding the problem!

PR #2336, opened 2026-01-22
timegm api absent in AIX

mehendarkarprajwal:
The timegm api is not part of AIX libc. Hence we want to use the alternative available ie. mktime


yhirose:
@mehendarkarprajwal thanks for the contribution!

PR #2339, opened 2026-01-23
fix: allow non -chunked streams w/o Content-Length header

stefan-tb:
Plain (non chunked) HTTP bodies from server responses without a `Content-Length` header can't be read at client side.

Example:
```c
auto stream = cli.open_stream("GET", "/stream-path");
auto n = stream.read(buffer, sizeof(buffer)); // always returns 0
```

That's because in the case of the missing header, the method `ClientImpl::open_stream` doesn't set `BodyReader::content_length` and the value remains entirely untouched (default value 0):

https://github.com/yhirose/cpp-httplib/blob/fbec2a3466efa5fb89527195fd103f2f116d98f5/httplib.h#L10448-L10453

Therefore consecutive tests in `BodyReader::read`, like this one:

https://github.com/yhirose/cpp-httplib/blob/fbec2a3466efa5fb89527195fd103f2f116d98f5/httplib.h#L8128-L8131

will immediately set `eof `and return 0 before even trying to read any actual data.

This PR is introducing another field `bool has_content_length` (`std::optional` is not C++11 as far as I know) to track if `BodyReader::content_length` has been set. If not, the related tests are skipped. Please note that the `eol` flag will still be properly set when a read from the underlying stream returns 0.

I am new to this code base, this is a best effort commit. I tried to keep things as local as possible. Sorry in advance if I am violating style / approaches.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -1475,6 +1475,7 @@ struct ChunkedDecoder;
 
 struct BodyReader {
   Stream *stream = nullptr;
+  bool has_content_length = false;
   size_t content_length = 0;
   size_t bytes_read = 0;
   bool chunked = false;
@@ -5707,6 +5708,7 @@ inline bool read_content_with_length(Stream &strm, size_t len,
 
   detail::BodyReader br;
   br.stream = &strm;
+  br.has_content_length = true;
   br.content_length = len;
   br.chunked = false;
   br.bytes_read = 0;
@@ -8125,13 +8127,13 @@ inline ssize_t detail::BodyReader::read(char *buf, size_t len) {
 
   if (!chunked) {
     // Content-Length based reading
-    if (bytes_read >= content_length) {
+    if (has_content_length && bytes_read >= content_length) {
       eof = true;
       return 0;
     }
 
     auto remaining = content_length - bytes_read;
-    auto to_read = (std::min)(len, remaining);
+    auto to_read = has_content_length ? (std::min)(len, remaining) : len;
     auto n = stream->read(buf, to_read);
 
     if (n < 0) {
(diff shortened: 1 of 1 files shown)


yhirose:
@stefan-tb Thanks for the pull request. Could you please add the unit test in `test/test.cc` that I always ask contributors to include? Thanks!

stefan-tb:
@yhirose: wow. 🥇 for your response time. Honestly.
I will. I need to look into them. Might take a week.

Thanks

yhirose:
@stefan-tb I just had the Copilot review this pull request, and here is the summary. Could you take a look at this potential problem, too? 

> This PR introduces a potential denial-of-service (DoS) vulnerability on the client side. When an HTTP response lacks both the Content-Length header and Transfer-Encoding: chunked, the client will now read the body until the connection is closed, without any built-in limit. As a result, if a malicious server sends an unbounded or extremely large body, the client may consume excessive memory or disk resources and cannot protect itself by default. Unlike the server side, which enforces a maximum payload size (`set_payload_max_length`), there is no similar safeguard for client responses, making this change dangerous without additional resource limits.

I have already fixed the same issue on the server side before. You can see the code in the following locations.

**Payload Max Length ExceedLimit:**
https://github.com/yhirose/cpp-httplib/blob/fbec2a3466efa5fb89527195fd103f2f116d98f5/httplib.h#L9126-L9131
https://github.com/yhirose/cpp-httplib/blob/fbec2a3466efa5fb89527195fd103f2f116d98f5/test/test.cc#L8244-L8507

**Decompressed Size Exceeds Limit:**
https://github.com/yhirose/cpp-httplib/blob/fbec2a3466efa5fb89527195fd103f2f116d98f5/httplib.h#L9016-L9023
https://github.com/yhirose/cpp-httplib/blob/fbec2a3466efa5fb89527195fd103f2f116d98f5/test/test.cc#L14240

Thanks!

stefan-tb:
My changes only apply to non-chunked, client-side streaming scenarios. Calling `cli.Get("/")` is doing its entirely own thing (to my understanding from Debugging my entire way my thru it) and reads "only" `res.body.max_size()` in case of a missing header, which is effectively int64.MaxValue and therefore quite large.

PR #2342, opened 2026-02-02
Abstract TLS API support (Resolve #2309)

### LINKED ISSUE #2309: Optional mbedTLS support for embedded / cross-compiled environments
Hi @yhirose,

First of all, thank you for `cpp-httplib`.
Its clean design, single-header approach, and minimal dependencies make it a joy to use.

I’m currently working on an embedded / cross-compiled project (e.g. ARM Linux),
where **[mbedTLS](https://github.com/Mbed-TLS/mbedtls) is often the preferred or only available TLS library** due to its
smaller footprint and common adoption in embedded toolchains.

At the moment, `cpp-httplib` enables HTTPS via OpenSSL (and compatible implementations),
which works very well on desktop and server platforms.
Would you consider adding **optional mbedTLS support** as an alternative TLS backend,
enabled only when explicitly requested?

I completely understand that supporting multiple TLS backends increases complexity
and maintenance cost, and that this may be outside the project’s current scope.
Even a minimal client-side HTTPS integration would already be very valuable
for embedded use cases.

If this is something you’d be open to in principle, I’d be happy to help with testing
or provide feedback from an embedded-user perspective.

Thanks again for your excellent work.

yhirose:
Abstract TLS API support. Currently, the following backends are supported.
- OpenSSL
- MbedTLS

### DIFF
--- a/test/test.cc
+++ b/test/test.cc
@@ -9985,6 +9985,7 @@ TEST(SSLClientRedirectTest, CertFile) {
 }
 #endif
 
+#ifdef CPPHTTPLIB_SSL_ENABLED
 // Test that set_ca_cert_store() skips system certs (consistent with
 // set_ca_cert_path behavior). When a custom cert store is set, only those certs
 // should be trusted - system certs should NOT be loaded.
@@ -10994,6 +10995,7 @@ TEST(RedirectTest, RedirectToUrlWithQueryParameters) {
     EXPECT_EQ("val&key2=val2", res->body);
   }
 }
+#endif
 
 TEST(RedirectTest, RedirectToUrlWithPlusInQueryParameters) {
   Server svr;


Spixmaster:
@Tachi107, I think that this requires a small update of `CMakeLists.txt`. Probably a new component like `MbedTLS` to set the macro `CPPHTTPLIB_MBEDTLS_SUPPORT`.

Sets macro `CPPHTTPLIB_OPENSSL_SUPPORT`.

```cmake
find_package("httplib" REQUIRED COMPONENTS "Brotli" "OpenSSL" "ZLIB" "zstd")
```

Needs to set macro `CPPHTTPLIB_MBEDTLS_SUPPORT`.

```cmake
find_package("httplib" REQUIRED COMPONENTS "Brotli" "MbedTLS" "ZLIB" "zstd")
```

PR #2347, opened 2026-02-08
Fix out of bounds read in etag comparison

emmericp:
string_view is easier to use correctly here, for std::equal we'd need an explicit extra parameter for the end of the second range for correctness.


yhirose:
@emmericp thanks for the pull request. However, this library supports C++11 compilers which don't have std::string_view yet.

yhirose:
I'll try to implement it without std::string_view. Thanks!

Issue #2345, opened 2026-02-06
CMake component `MbedTLS`


Spixmaster:
I think that [this pull request](https://github.com/yhirose/cpp-httplib/pull/2342) requires a small update of `CMakeLists.txt`. Probably a new component like `MbedTLS` to set the macro `CPPHTTPLIB_MBEDTLS_SUPPORT`.

Overall conidition via `CPPHTTPLIB_SSL_ENABLED`.

Sets macro `CPPHTTPLIB_OPENSSL_SUPPORT`.

```cmake
find_package("httplib" REQUIRED COMPONENTS "Brotli" "OpenSSL" "ZLIB" "zstd")
```

Needs to set macro `CPPHTTPLIB_MBEDTLS_SUPPORT`.

```cmake
find_package("httplib" REQUIRED COMPONENTS "Brotli" "MbedTLS" "ZLIB" "zstd")
```

yhirose:
@Spixmaster could you please send a pull request?

Spixmaster:
Why was this closed? CMake and Meson are different meta build tools or am I wrong?

Spixmaster:
@Tachi107, where to add the component? I do not find `..._FIND_COMPONENTS`, https://cmake.org/cmake/help/latest/guide/importing-exporting/index.html#adding-components.

DarthGandalf:
Yeah, I only updated meson, not cmake

yhirose:
Sorry for the misunderstanding. I would appreciate it if someone could handle the CMake update.

yhirose:
Since the corresponding pull request (#2360) has been created, further discussions should take place there.

PR #2350, opened 2026-02-10
Add support for mbedtls to meson

DarthGandalf:
Related: #2345

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -39,11 +39,29 @@ endif
 deps = [dependency('threads')]
 args = []
 
-openssl_dep = dependency('openssl', version: '>=3.0.0', required: get_option('openssl'))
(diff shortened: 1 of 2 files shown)


DarthGandalf:
should I make it automatically search for mbedtls if openssl isn't available?

yhirose:
@DarthGandalf thanks for the pull request.
@Tachi107 @sum01 @jimmy-park could you review this pull request when you have time?
Thanks all!

Tachi107:
One small nitpick before reviewing the code: could you name the option as tls instead of ssl? It’s the term nowadays used to refer to the standard. Thanks!

yhirose:
Thanks for the fine contribution!

PR #2354, opened 2026-02-10
Fix compilation on BoringSSL by replacing ASN1_TIME_to_tm

angt:
BoringSSL doesn't expose `ASN1_TIME_to_tm`.
This patch switches to using `ASN1_TIME_diff` to calculate `time_t`. This is supported by OpenSSL, LibreSSL, and BoringSSL, and also avoids the platform-specific `timegm` vs `_mkgmtime` logic.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -15457,26 +15457,18 @@ inline bool get_cert_validity(cert_t cert, time_t &not_before,
 
   ASN1_TIME *epoch = ASN1_TIME_new();
   if (!epoch) return false;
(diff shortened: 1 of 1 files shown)


yhirose:
@angt, thanks for the patch! Could you please use `detail::scope_exit` to call `ASN1_TIME_free`, instead of handling it manually? You can see similar patterns in httplib.h for releasing resources safely.

yhirose:
@angt thanks for your contribution!

ksokolovskyi:
@angt thanks for the fix!

PR #2355, opened 2026-02-11
Correct sign comparison error with sk_X509_OBJECT_num

strangewiz:
In some build configurations, sk_X509_OBJECT_num (from BoringSSL) returns a size_t. Comparing this directly against a signed int loop counter triggers -Werror,-Wsign-compare.

Instead, move the count into a local int variable so the compiler uses implicit conversion to ensure type consistency during the loop comparison.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -15596,7 +15596,7 @@ inline std::vector<std::string> get_ca_names(ctx_t ctx) {
   if (!objs) { return names; }
 
   auto count = sk_X509_OBJECT_num(objs);
-  for (int i = 0; i < count; i++) {
+  for (decltype(count) i = 0; i < count; i++) {
     auto obj = sk_X509_OBJECT_value(objs, i);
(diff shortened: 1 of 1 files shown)


httplib.h
```diff
@@ -15593,7 +15595,7 @@ inline std::vector<std::string> get_ca_names(ctx_t ctx) {
   auto objs = X509_STORE_get0_objects(store);
   if (!objs) { return names; }
 
-  int count = sk_X509_OBJECT_num(objs);
+  auto count = sk_X509_OBJECT_num(objs);
   for (int i = 0; i < count; i++) {
```
yhirose: Should `int` be `decltype(count)`?

strangewiz: It should, thanks!

yhirose:
@strangewiz thanks for your contribution!

ksokolovskyi:
@strangewiz thanks for the fix!

PR #2356, opened 2026-02-11
Setup Gentoo+Meson CI

DarthGandalf:
The goal was to add a CI with Meson, and without access to internet, because Gentoo emerge runs tests in sandbox, skipping tests which are explicitly marked as online.
But on github-hosted runners, there's no good way to disable network, emulate that by emptying resolv.conf

With such workaround the "Gentoo" part is not really relevant anymore, the same meson+resolv.conf can be done in Ubuntu. But this does increase coverage a bit, so why not.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -200,3 +200,37 @@ jobs:
     env:
       VCPKG_ROOT: "C:/vcpkg"
       VCPKG_BINARY_SOURCES: "clear;x-gha,readwrite"
+
+  gentoo:
+    runs-on: ubuntu-latest
+    container:
+      image: gentoo/stage3:latest
+    if: >
+      (github.event_name == 'push') ||
+      (github.event_name == 'pull_request'  &&
+       github.event.pull_request.head.repo.full_name != github.event.pull_request.base.repo.full_name) ||
+      (github.event_name == 'workflow_dispatch' && github.event.inputs.test_linux == 'true')
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Sync portage tree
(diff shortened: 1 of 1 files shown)


yhirose:
I made #2358 which handles the offline test with openssl, mbedtls, no_tls on ubuntu.

DarthGandalf:
Do I understand correctly that you just filtered which tests to run?

The point here is to ensure that tests which require network, all have the suffix. I don't see you disabling network in any way?

yhirose:
You’re right. I forgot to disable the network. Do you notice any missing pieces? Thanks!

```yaml
      - name: disable network
        run: sudo sh -c 'echo > /etc/resolv.conf'
```

DarthGandalf:
I suppose that should work, thanks. Not sure what's the value of the matrix `[openssl, mbedtls, no-tls]` though: if the purpose of the test is to check names of tests, running one of them should probably be enough; the rest is just subset of tests running in the other ubuntu CI

What about having meson in one of jobs? That one would benefit from a matrix actually

yhirose:
There are non `_Online` tests for both `CPPHTTPLIB_SSL_ENABLED` and when `CPPHTTPLIB_SSL_ENABLED` is not defined. So we need to test with both `no-tls` and either `openssl` or `mbedtls`. Since `mbedtls` was just introduced and `openssl` is more widely used, I'll stick with `openssl` and `no-tls`. Thanks for your insight!

yhirose:
The only issue is that this test_offline adds 10 more minutes to the GitHub CI. The CI tests already take quite a long time, so I'm wondering if it's really worth including this check in this way. If there's a more efficient approach, I'd prefer to use that instead.

PR #2358, opened 2026-02-11
Offline test (Resolve #2356)

yhirose:
Offline test on Ubuntu (OpenSSL, no TLS)


yhirose:
See comments in https://github.com/yhirose/cpp-httplib/pull/2356.

yhirose:
The only issue is that this test_offline adds 10 more minutes to the GitHub CI. The CI tests already take quite a long time, so I'm wondering if it's really worth including this check in this way. If there's a more efficient approach, I'd prefer to use that instead.

PR #2359, opened 2026-02-11
Bypass new 100MB payload limit when a content_receiver is used

angt:
The new hard payload limit broke the `llama.cpp` downloader and while this is expected for a breaking change, I believe it is going to break a lot of code where this limit is not really helpful and hard to handle correctly when we use our own callback to handle the download. 

Let me know if you think there is a better way to handle this.

With this patch, `llama.cpp` successfully downloads models like before.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -9158,6 +9158,8 @@ inline bool SSLSocketStream::wait_writable() const {
 }
 
 inline ssize_t SSLSocketStream::read(char *ptr, size_t size) {
+  constexpr size_t max_int = static_cast<size_t>((std::numeric_limits<int>::max)());
+  if (size > max_int) size = max_int;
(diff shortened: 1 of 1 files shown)


yhirose:
@angt, thanks for the excellent feedback! I made a similar change in a slightly different way. I introduced has_payload_max_length_ in the Client. When a user explicitly sets the value, httplib will still honor the payload limit even when content_receiver is used. Otherwise, the payload limit will not be enforced. As for the clamping the size value, this is not correct for MbedTLS. So I moved it to the OpenSSL version of `tls::read`. Please let me know if you find any remaining problems.

yhirose:
I've just released v0.32.0. I hope it works well with llama.cpp.
https://github.com/yhirose/cpp-httplib/releases/tag/v0.32.0

PR #2365, opened 2026-02-13
Remove macOS select() fallback

angt:
macOS has supported `poll` for a long time now, so there's no need for the specific `select` code paths.

With this commit, we can successfully build on visionOS.


yhirose:
Thank you for simplifying this, @angt!

Issue #2366, opened 2026-02-13
std::this_thread::sleep() is unsafe on older MSVC


ian-at-des:
**Note:** this issue only affects MSVC STL prior to v17.9 (2022). See [microsoft/STL#718](https://github.com/microsoft/STL/issues/718#issuecomment-3385231473) for details (thanks @Curve).

# Background
The c++ standard [recommends](https://en.cppreference.com/w/cpp/thread/sleep_for.html), but doesn't require, that `std::this_thread::sleep_for()` (and related functions like `sleep_until()`) use a steady clock i.e. one that is monotonic. Unfortunately MSVC implements `this_thread::sleep_for()` using the system clock. This is a [known limitation](https://developercommunity.visualstudio.com/t/bogus-stdthis-threadsleep-for-implementation/58530) but fixing it would require a breaking ABI change which Microsoft do not currently plan to do.

If a thread calls `this_thread::sleep_for()` and the system clock changes (e.g. by the user or by the NTP sync service) then the thread may either:
- Wake up early (not generally an issue)
- Never wake up (generally undesirable behaviour)

The Win32 `::Sleep()` function does not use the system clock and so is not vulnerable to system time changes.

# Bug
Several places call `this_thread::sleep_for()` (e.g. `ClientImpl::write_request()`). Any place that calls `this_thread::sleep_for()` may cause the thread to hang indefinitely, requiring the user to kill/restart the application.

# Fix
- Provide a wrapper `httplib::detail::sleep_for()` that uses `::Sleep()` on MSVC, and `this_thread::sleep_for()` on other platforms.
- Replace uses of `this_thread::sleep_for()` with `detail::sleep_for()`.

I'm happy to provide an implementation, just let me know if you'd prefer it posted as a comment or via a pull request.

# Outstanding Questions
Looking at where `this_thread::sleep_for()` is called, httplib seems to have two use cases:
- `sleep_for(1..100ms)` -> this looks like a genuine "suspend the thread and come back in a while"
- `sleep_for(1..10us)` -> this looks more like "yield my time slice but come back quickly".

[...]

PR #2372, opened 2026-02-16
Fix older MSVC issue where some operations could block indefintately

ian-at-des:
See comment in new `httplib::detail::sleep_for()` function, and issue #2366 for details.


ian-at-des:
Closing as per https://github.com/yhirose/cpp-httplib/issues/2373#issuecomment-3920354212 which also applies to https://github.com/yhirose/cpp-httplib/issues/2366

Issue #2188, opened 2025-07-22
Request for mbedtls Support as a Alternative to OpenSSL


zhlinh:
I hope [mbedtls](https://github.com/Mbed-TLS/mbedtls) can be supported as a alternative to OpenSSL, as it allows for a smaller package size.

yhirose:
@zhlinh thanks for the request. I don't have time to work on that. Could you please send me a pull request for this support? Thanks!

Issue #1142, opened 2021-12-24
Support mbedtls instead of OpenSSL


mrexodia:
OpenSSL is a very heavy dependency and mbedtls is fast to compile and easy to embed in your application.

I looked at the code a year ago or so, but it doesn’t seem like the SSL parts are designed to be replaced easily unfortunately…

A similar idea would be to support the OS’ native SSL implementation so that you can do HTTPS without any dependencies. This should be possible at least on Windows but I’m pretty sure on Mac too.

Also: do you have GitHub sponsors?

yhirose:
@mrexodia, thanks for the feedback. It's a good idea, but I simply don't have time to work on it. Of course, a pull request is always welcome!

mrexodia:
That looks very promising, I will definitely give it a try!

lifw555:
Has there been any progress?
thx

PR #2389, opened 2026-03-08
Add TLSe support

nitanmarcel:
This PR adds support for the [TLSe](https://github.com/eduardsui/tlse) library by @eduardsui.

Tested only on Linux and Android at this moment, both systems work fine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,7 @@ cpp-httplib supports multiple TLS backends through an abstraction layer:
 | OpenSSL | `CPPHTTPLIB_OPENSSL_SUPPORT` | `libssl`, `libcrypto` |
 | Mbed TLS | `CPPHTTPLIB_MBEDTLS_SUPPORT` | `libmbedtls`, `libmbedx509`, `libmbedcrypto` |
 | wolfSSL | `CPPHTTPLIB_WOLFSSL_SUPPORT` | `libwolfssl` |
+| TLSe | `CPPHTTPLIB_TLSE_SUPPORT` | `tlse.c` + `libtomcrypt.c` (compiled as C) |
 
 > [!NOTE]
 > OpenSSL 3.0 or later is required. Please see [this page](https://www.openssl.org/policies/releasestrat.html) for more information.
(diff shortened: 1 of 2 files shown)


Spixmaster:
Honestly, can we stop with adding support for multiple TLS libraries. It does not benefit the project anymore and makes maintenance a lot harder with no positive outcome.

I understood the profit by supporting Mbed TLS (https://github.com/yhirose/cpp-httplib/issues/1142, https://github.com/yhirose/cpp-httplib/issues/2188, https://github.com/yhirose/cpp-httplib/issues/2309) but the rest just increases maintenance.

nitanmarcel:
Ok then. Maybe, and if I can give a suggestion, there's a way to allow users to implement their ssl libraries using some callbacks cpp-httplib provides? Just an idea, no clue if it's actually possible without a big refactoring

Spixmaster:
Just to be clear, I have no authority to decide anything. I am just worried that this shifts into a wrong direction.

It is rather uncommon to have mutually exclusive dependencies. I have heard of it once in an issue somewhere else than in this project.

yhirose:
@nitanmarcel, thank you for the contribution and for the effort put into this PR!

Unfortunately, I'm not able to merge support for TLSe into cpp-httplib. As a security-critical component, I need to limit TLS backend support to well-established libraries with a strong track record of active maintenance, continuous security auditing, and broad industry adoption — such as OpenSSL, MbedTLS, and wolfSSL. TLSe, while an interesting single-file implementation, doesn't meet that bar at this time.

That said, I'd like to share some positive news and thoughts:

I recently implemented a TLS abstraction API in cpp-httplib, which is what now allows support for multiple backends beyond OpenSSL (MbedTLS and wolfSSL are already supported through this). This actually makes the idea raised in comment #2 — *"Is there a way for users to implement their own SSL library?"* — a very compelling direction. I think exposing the abstraction API and common helper functions to users would be a great next step, allowing anyone to plug in a custom TLS backend (including TLSe, if that's what they need) without requiring it to be bundled into the library itself.

If this PR were to be revised with that in mind — leveraging the existing common functions rather than reimplementing them — it could also serve as a nice reference implementation for how a custom TLS backend can be built on top of cpp-httplib's abstraction layer.

I have higher-priority projects to focus on for the time being, so I won't be able to act on this anytime soon — but I'll keep it in mind as a worthwhile direction for the future.

Thanks again for the proposal — the underlying idea is genuinely valuable, even if I can't merge it in its current form.

PR #2390, opened 2026-03-08
[cmake] Allow using pre-existing zstd target if it exists 

crueter:
adds support for pre-existing `zstd::libzstd` which is useful for
projects that bundle their own zstd in a way that doesn't get caught by
`CONFIG`

Signed-off-by: crueter <crueter@eden-emu.dev>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -242,27 +242,33 @@ endif()
 # zstd < 1.5.6 does not provide the CMake imported target `zstd::libzstd`.
 # Older versions must be consumed via their pkg-config file.
 if(HTTPLIB_REQUIRE_ZSTD)
-	find_package(zstd 1.5.6 CONFIG)
-	if(NOT zstd_FOUND)
-		find_package(PkgConfig REQUIRED)
-		pkg_check_modules(zstd REQUIRED IMPORTED_TARGET libzstd)
-		add_library(zstd::libzstd ALIAS PkgConfig::zstd)
+	if (NOT TARGET zstd::libzstd)
+		find_package(zstd 1.5.6 CONFIG)
+		if(NOT zstd_FOUND)
+			find_package(PkgConfig REQUIRED)
+			pkg_check_modules(zstd REQUIRED IMPORTED_TARGET libzstd)
+			add_library(zstd::libzstd ALIAS PkgConfig::zstd)
+		endif()
 	endif()
 	set(HTTPLIB_IS_USING_ZSTD TRUE)
 elseif(HTTPLIB_USE_ZSTD_IF_AVAILABLE)
-	find_package(zstd 1.5.6 CONFIG QUIET)
-	if(NOT zstd_FOUND)
-		find_package(PkgConfig QUIET)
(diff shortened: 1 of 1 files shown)


yhirose:
@sum01 @jimmy-park @Tachi107 could you please check the pull request? If it looks good, I'll merge it. Thanks!

CMakeLists.txt
```diff
@@ -241,28 +241,34 @@ endif()
...
-			if(TARGET PkgConfig::zstd)
+if (HTTPLIB_REQUIRE_ZSTD)
+	set(HTTPLIB_ZSTD_REQUESTED ON)
+	set(HTTPLIB_ZSTD_REQUIRED REQUIRED)
+elseif (HTTPLIB_USE_ZSTD_IF_AVAILABLE)
+	set(HTTPLIB_ZSTD_REQUESTED ON)
+	set(HTTPLIB_ZSTD_REQUIRED QUIET)
+endif()
+
+if (HTTPLIB_ZSTD_REQUESTED)
```
sum01: You should probably just wrap the whole block in an `if(TARGET zstd::libztsd)` instead of adding more vars.

yhirose:
@crueter any response?

CMakeLists.txt
crueter: What do you mean? These vars are only used in the case where the target does not already exist.

sum01: Then wrap it with an `if(NOT TARGET zstd::libzstd)`

yhirose: Thanks for the contribution and the review discussion!

@crueter I looked at how the other dependencies (OpenSSL, wolfSSL, MbedTLS, ZLIB, Brotli) are handled in `CMakeLists.txt`, and they all follow the same two-branch pattern:

```
if(HTTPLIB_REQUIRE_XXX)
    find_package(XXX REQUIRED)
    set(HTTPLIB_IS_USING_XXX TRUE)
elseif(HTTPLIB_USE_XXX_IF_AVAILABLE)
    find_package(XXX QUIET)
    set(HTTPLIB_IS_USING_XXX ${XXX_FOUND})
endif()
```

The `HTTPLIB_ZSTD_REQUESTED` / `HTTPLIB_ZSTD_REQUIRED` variable approach in this PR would make zstd the only dependency that deviates from this established pattern.

I'd prefer to keep consistency with the other dependencies. Could you rework this to keep the existing `if(REQUIRE) / elseif(IF_AVAILABLE)` two-branch structure, and add the pre-existing target support with an `if(NOT TARGET zstd::libzstd)` guard inside each branch, as @sum01 suggested?

Yes, the pkg-config fallback lines will remain duplicated across the two branches, but that's a small cost for keeping the CMake structure consistent and easy to follow.

crueter: Sure, I can do that in a bit

crueter: Fixed. I think the overall `available`/`requested` logic should be reworked for *everything* eventually, but that's for a later me or someone else to figure out

yhirose:
Thanks all!

Issue #2392, opened 2026-03-10
Cmake error with modules support


v-nam:
Greetings!

I use CPM to build the project with the following settings:

```
CPMAddPackage(
    NAME httplib
    VERSION 0.37.0 
    URL https://github.com/yhirose/cpp-httplib/archive/refs/tags/v0.37.0.tar.gz
    URL_HASH SHA256=4663e5596ad182b18c86ca5c552f74a4b256b4c31042699dcb52b6ca02f797e5
    OPTIONS "CPPHTTPLIB_OPENSSL_SUPPORT ON"
    OPTIONS "HTTPLIB_COMPILE ON"
    OPTIONS "HTTPLIB_BUILD_MODULES ON"
)
```

When I configure a project with option **HTTPLIB_BUILD_MODULES ON**, the following message appears:
...
[cmake] CMake Error at /home/beny/.cache/CPM/httplib/2cbd/cmake/modules.cmake:11 (target_sources):
[cmake]   File:
[cmake] 
[cmake]     /home/aaa/projects/app/build/debug/_deps/httplib-build/out/httplib.cppm
[cmake] 
[cmake]   must be in one of the file set's base directories:
[cmake] 
[cmake]     /home/aaa/.cache/CPM/httplib/2cbd
[cmake] Call Stack (most recent call first):
[cmake]   /home/aaa/.cache/CPM/httplib/2cbd/CMakeLists.txt:327 (include)
[cmake] 
[cmake] 
[cmake] CMake Error at /home/aaa/.cache/CPM/httplib/2cbd/cmake/modules.cmake:11 (target_sources):
[cmake]   File:
[cmake] 
[cmake]     /home/aaa/projects/lv-admin/build/debug/_deps/httplib-build/out/httplib.cppm
[cmake] 
[cmake]   must be in one of the file set's base directories:
[cmake] 
[cmake]     /home/aaa/.cache/CPM/httplib/2cbd
[cmake] Call Stack (most recent call first):
[cmake]   /home/aaa/.cache/CPM/httplib/2cbd/CMakeLists.txt:327 (include)
[cmake] 
[cmake] 
[cmake] CMake Error at /home/aaa/.cache/CPM/httplib/2cbd/cmake/modules.cmake:11 (target_sources):
[cmake]   File:
[cmake] 
[cmake]     /home/aaa/projects/app/build/debug/_deps/httplib-build/out/httplib.cppm
[cmake] 
[...]

PR #2393, opened 2026-03-10
[cmake] Update modules.cmake to fix cmake error

v-nam:
Update modules.cmake to fix cmake error

### DIFF
--- a/cmake/modules.cmake
+++ b/cmake/modules.cmake
@@ -12,7 +12,7 @@ if(HTTPLIB_BUILD_MODULES)
 		PUBLIC
 			FILE_SET CXX_MODULES 
 			BASE_DIRS 
-				${_httplib_build_includedir}
+				"${_httplib_build_includedir}"
 			FILES
 				"${_httplib_build_includedir}/httplib.cppm"
 	)


v-nam:
Check again please

yhirose:
@sum01 @jimmy-park @Tachi107 could you please review it?

Tachi107:
Hi, this seems to be related to the modules code added by @mikomikotaishi. I have no idea what this is supposed to fix :)

mikomikotaishi:
Could you give a more clear description of what the error was in the old code?

v-nam:
I added a lines 'BASE_DIRS ${_httplib_build_includedir}' and the second commit added quotes around the variable ${_httplib_build_includedir}

mikomikotaishi:
Could you explain what the problem was before this change though? I am a little confused as to what this is intended to solve.

v-nam:
Initially, the problem was when working with the module. When eliminating it, errors appeared when passing tests. I added quotes around the cmake variable and after that all the tests were passed successfully.

DarthGandalf:
Please paste the errors. You're saying they appeared. What errors?

v-nam:
It all started with [this](https://github.com/yhirose/cpp-httplib/issues/2392). Then I tried to make a comit, but it didn't pass the [tests](https://github.com/yhirose/cpp-httplib/actions/runs/22907886229). After that, I found and corrected the error associated with cmake syntax - I did not put quotes. That's all.

DarthGandalf:
Right... For future, a link such as #2392 should be mentioned directly in the PR description. Without that, it's unclear what the PR is trying to do.

v-nam:
I will mean, thanks

sum01 (commented):
I don't use CPM so I don't know how relevant the fix is, but it seems fine.

yhirose:
Thanks all!

PR #2401, opened 2026-03-20
Adding configurable TCP no delay to websocket

snibo-scale:
Adding function to allow disabling of Nagle's algorithm for WebSocket clients


yhirose:
@snibo-scale thank you!

PR #2406, opened 2026-03-24
Simplify OpenSSL settings in the guide

DarthGandalf:
cpp-httplib's CMakeLists.txt already enables everything needed, including for MacOS

### DIFF
--- a/docs-src/pages/en/llm-app/ch04-model-management.md
+++ b/docs-src/pages/en/llm-app/ch04-model-management.md
@@ -65,7 +65,7 @@ set(HTTPLIB_REQUIRE_OPENSSL true)
 FetchContent_Declare(httplib ...)
 ```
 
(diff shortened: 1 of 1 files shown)


yhirose:
@DarthGandalf Thank you for the suggestion! While HTTPLIB_REQUIRE_OPENSSL does simplify things, this document is a tutorial chapter aimed at helping readers understand what's happening under the hood. Keeping the explicit find_package, target_link_libraries, and target_compile_definitions calls teaches readers the standard CMake patterns for enabling OpenSSL — knowledge that transfers to other projects too. The macOS framework linking is also worth showing explicitly so readers understand why it's needed when they encounter certificate issues. For these reasons, I'd prefer to keep the current verbose form.

Also, please note that the docs have multiple language versions (currently English and Japanese) that need to stay in sync. Any changes to the English pages would require corresponding updates to the other language versions as well.

PR #2407, opened 2026-03-24
[Issue#2400] Fix for Segmentation Fault in Debug Mode While Trying to Join Server Listen Thread

dimy-x:
[Issue#2400] Fix for Segmentation Fault in Debug Mode While Trying to Join Server Listen Thread

MinGW 64 bit 13.1.0 (gcc-13.1.0 x86_64-1310-posix-seh-msvcrt-rt_v11-rev1) / Qt Creator 16.0.2 / Qt 6.8.3

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -5060,7 +5060,7 @@ inline bool parse_trailers(stream_line_reader &line_reader, Headers &dest,
   // chunked transfer coding data without the final CRLF.
 
   // RFC 7230 Section 4.1.2 - Headers prohibited in trailers
-  thread_local case_ignore::unordered_set<std::string> prohibited_trailers = {
+  static const case_ignore::unordered_set<std::string> prohibited_trailers = {
       "transfer-encoding",
       "content-length",
       "host",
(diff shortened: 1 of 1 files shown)


yhirose:
@dimy-x Thank you for the pull request. However, I can’t merge it because there’s a specific reason why thread_local was chosen. Your change will create the thread-safe issue. (Also [MinGW isn't supported](https://github.com/yhirose/cpp-httplib?tab=readme-ov-file#windows).)

You can find more background information here:
https://github.com/yhirose/cpp-httplib/issues?q=thread_local

dimy-x:
Hi Yuji,

Actually, what I've modified from thread_local to 'static const' should be fine (I haven't replaced all occurencies of thread_local in the code, because some occurencies should not be replaced), but what I've replaced with 'static const' should be thread safe and not do any harm, because it is 'const' and we don't modify it so no race conditions or synchronization issues. Could you please add my modifications into the code ? I could do the pull request again if you need it..

Kind regards,
Dimy

dimy-x:
Hi Yuji,

Should I do the pull request again ? Do you agree to add it ? (take a look at my previous comment please)

Kind regards,
dimy-x

yhirose:
@dimy-x sorry, but I am not able to accept this pull request.
https://github.com/yhirose/cpp-httplib/issues/2097
https://github.com/yhirose/cpp-httplib/commit/87a5ae64a416d03021d990c962805d3f0549a5a8

PR #2408, opened 2026-03-25
Enhancement and fixes based on static code analysis.

SiarheiHerasiuta:
[Enhancement] Warning level changed for examples from 3 to 4.
[Fix] Moved buffers allocation to heap memory instead stack.
[Fix] Added zero initialization on some variables.


yhirose:
@SiarheiHerasiuta Thank you! However, I am not planning to merge this pull request, since “[Fix] Moved buffers allocation to heap memory instead of stack” is not acceptable for performance reasons.

PR #2411, opened 2026-03-27
Export set_socket_opt in detail namespace

angt:
This allows users to set custom socket options without reimplementing platform specific wrappers

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -1565,6 +1565,8 @@ int close_socket(socket_t sock);
 
 ssize_t write_headers(Stream &strm, const Headers &headers);
 
+bool set_socket_opt(socket_t sock, int level, int optname, int optval);
+
 bool set_socket_opt_time(socket_t sock, int level, int optname, time_t sec,
                          time_t usec);
 


yhirose:
@angt does the following method meets your needs?

```cpp
  Server &set_socket_options(SocketOptions socket_options);
```

angt:
> @angt does the following method meets your needs?
> 
> ```c++
>   Server &set_socket_options(SocketOptions socket_options);
> ```

I use it to customize the setsockopt function yes, but I need to handle the Windows case myself and replicate the exact same `set_socket_opt` function that’s already present in the `cpp` file but not exported.

yhirose:
I intentionally put `set_socket_opt` in the detailed namespace which is not meant to be part of the public API. However, if it would be useful, I don’t mind moving it to the public interface. Before doing so, though, I’d like to understand the full context of why the existing public method (`set_socket_options`) didn’t work in the particular situation.

Could you please try to resolve the issue you’re experiencing using the public method `set_socket_options`? If that doesn’t work, I’ll take a closer look. Thanks!

angt:
This was just to allow the user to have a simple code like this:
```
srv->set_socket_options([](socket_t sock) {
    set_socket_opt(sock, SOL_SOCKET, SO_REUSEADDR, 1);
});
```
instead of
```
srv->set_socket_options([](socket_t sock) {
    int opt = 1;
#ifdef _WIN32
    const char *optval = (const char *)&opt;
#else
    const void *optval = &opt;
#endif
    setsockopt(sock, SOL_SOCKET, SO_REUSEADDR, optval, sizeof(opt));
});
```

yhirose:
@angt I made the function public.

PR #2412, opened 2026-03-28
Add two new fuzzers

DavidKorczynski:
The goal is to increase code coverage by way of OSS-Fuzz. A recent code coverage report is available at
https://storage.googleapis.com/oss-fuzz-coverage/cpp-httplib/reports/20260326/linux/report.html


yhirose:
@DavidKorczynski thanks!

PR #2425, opened 2026-04-12
fix: stop using PATH_MAX

Tachi107:
`PATH_MAX` is tricky, leads to buggy code, and isn't even defined on all platforms. For instance, GNU Hurd does not have it, and using it makes cpp-httplib incompatible with that OS.

Passing `nullptr` to `realpath()` is easier, and makes all the `PATH_MAX`-related problems go away.

For more reasoning and examples on avoiding `PATH_MAX`, see <https://github.com/HowardHinnant/date/commit/ac0c58d5da602e6387589d849f1d8d523091f4f0>

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -4949,9 +4949,10 @@ inline bool canonicalize_path(const char *path, std::string &resolved) {
   if (_fullpath(buf, path, _MAX_PATH) == nullptr) { return false; }
   resolved = buf;
 #else
-  char buf[PATH_MAX];
-  if (realpath(path, buf) == nullptr) { return false; }
+  char *buf = realpath(path, nullptr);
+  if (buf == nullptr) { return false; }
   resolved = buf;
+  std::free(buf);
 #endif
   return true;
 }
--- a/test/test.cc
+++ b/test/test.cc
@@ -17576,9 +17576,10 @@ TEST(SymlinkTest, SymlinkEscapeFromBaseDirectory) {
   }
 
   // Create symlink using absolute path so it resolves correctly
-  char abs_secret[PATH_MAX];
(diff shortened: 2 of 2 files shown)


yhirose:
@Tachi107 thanks! I fixed in a slightly different way. (Use the stack memory as much as possible.)

Tachi107:
Hi,

On Sun Apr 12, 2026 at 11:26 PM CEST, yhirose wrote:
> I fixed in a slightly different way. (Use the stack memory as much as possible.)

While your change makes the code compile when PATH_MAX isn't defined, it 
sadly doesn't solve all the other points that I brought up before. For 
instance, the code will break with paths longer than PATH_MAX.

Using stack memory unfortunately isn't possible in this case. You cannot 
know in advance how long a path may be.

PATH_MAX, per POSIX, is the "maximum number of bytes the implementation 
stores as a pathname in a user-supplied buffer", not the maximum number 
of bytes of any given path. If a path is longer, things will break.

https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/limits.h.html

Moreover, the link above also says:

> It should be noted, however, that many of the listed limits are not 
> invariant, and at runtime, the value of the limit may differ from 
> those given in this header, for the following reasons:
>
> - The limit is pathname-dependent.
>
> - The limit differs between the compile and runtime machines.
>
> - The limit has been changed at runtime by an application (see setrlimit()).
>
> For these reasons, an application can use the fpathconf(), 
> getrlimit(), pathconf(), and sysconf() functions to determine the 
> actual value of a limit at runtime.

In short: compile-time limits are often not what you think they are.

Please consider dropping the #if defined(PATH_MAX) branch, and using the 
realpath(path, nullptr) variant on all platforms.

PR #2426, opened 2026-04-12
fix: cast len to 64 bits before right shift in ws

Tachi107:
Fixes WebSocketIntegrationTest.LargeMessage and WebSocketIntegrationTest.MaxPayloadAtLimit on i386


yhirose:
@Tachi107 thanks a lot to fix the UB bug.

PR #2430, opened 2026-04-13
fix #2429

### LINKED ISSUE #2429: tls::read() returns -1 instead of 0 on clean SSL peer close, breaking Connection: close responses without Content-Length
Hi!

When upgrading from cpp-httplib 0.27 to 0.42, HTTPS responses that use Connection: close without a Content-Length header fail with Error::Read. I believe that the root cause is that the OpenSSL tls::read() abstraction converts SSL_read() returning 0 (clean peer shutdown) into -1, making it indistinguishable from an actual error.

Reproduction
Server: Any HTTPS server that responds with:

Connection: close
No Content-Length header
No Transfer-Encoding: chunked
A valid response body

I believe this should be a valid HTTP/1.1 response format per RFC 9112 -- "A response that has neither chunked transfer coding nor Content-Length is terminated by closure of the connection and, if the header section was received intact, is considered complete unless an error was indicated by the underlying connection." 

I found the following change to fix my issue:

``` cpp
  err.code = impl::map_ssl_error(ssl_err, err.sys_errno);
  if (err.code == ErrorCode::PeerClosed) { return 0; } // Gracefully handle the peer closed state.
  if (err.code == ErrorCode::Fatal) { err.backend_code = ERR_get_error(); }
```

Kukodam:
When SSL_read() returns 0 after receiving a close_notify alert, tls::read() incorrectly returned -1, making clean peer shutdown indistinguishable from a read error. This broke HTTP responses using Connection: close without Content-Length, where the connection close signals end-of-body. Return 0 for PeerClosed to preserve POSIX read() semantics.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -16766,6 +16766,9 @@ inline ssize_t read(session_t session, void *buf, size_t len, TlsError &err) {
 
   auto ssl_err = SSL_get_error(ssl, ret);
   err.code = impl::map_ssl_error(ssl_err, err.sys_errno);
+  if (err.code == ErrorCode::PeerClosed) {
+    return 0;
+  } // Gracefully handle the peer closed state.
   if (err.code == ErrorCode::Fatal) { err.backend_code = ERR_get_error(); }
   return -1;
 }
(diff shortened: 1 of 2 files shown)


yhirose:
@Kukodam thanks for your fine contribution!

PR #2437, opened 2026-04-28
Add client fuzzing harness

DavidKorczynski:
Cover client request processing logic. The goal is to enable this running on OSS-Fuzz.


yhirose:
@DavidKorczynski thanks!

PR #2451, opened 2026-05-14
Specifying "Server::stop()" as noexcept

NsPro04:
The current implementation of `Server::stop()` doesn't throw an exception, so why not specify this explicitly?

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -1597,7 +1597,7 @@ private:
   std::regex regex_;
 };
 
-int close_socket(socket_t sock);
+int close_socket(socket_t sock) noexcept;
 
 ssize_t write_headers(Stream &strm, const Headers &headers);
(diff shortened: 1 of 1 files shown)


yhirose:
@NsPro04 thanks for the contribution, but this change creates errors...

NsPro04:
I accidentally missed “noexcept” in the declaration. Fixed it.
Also remove the duplicate declaration.

NsPro04:
Afaict, the "BaseAuthTest.FromHTTPWatch_Online" and "KeepAliveTest.SSLWithDigest" tests failed, but... that's strange; I don't think my changes had anything to do with it, since everything else is working fine.

UPD: And now everything is fine

yhirose:
Thank you!

PR #2453, opened 2026-05-22
Fix `zstd` detection in installed `httplibConfig.cmake`

sakurai-ryuhei:
PR #2334 fixed `zstd` detection in [`CMakeLists.txt`](https://github.com/yhirose/cpp-httplib/blob/b1cc8095a8feca1b99e296ef78eddd312e80d2c6/CMakeLists.txt#L258), but the installed package config still uses the previous logic in `cmake/httplibConfig.cmake.in`.

As a result, `find_package(httplib REQUIRED)` can still fail after installation when `find_package(zstd)` succeeds without providing a usable `zstd::libzstd` target.

This PR updates the installed config to mirror the `zstd` detection strategy already used in the build-tree path.

# Reproducer

1. Install `cpp-httplib` with default options.
2. In another cmake project, call:

```cmake
find_package(httplib REQUIRED)
```

3. On Ubuntu 24.04, configuration can still fail with:

```
CMake Error at /usr/local/httplib/lib/cmake/httplib/httplibTargets.cmake:61 (set_target_properties):
  The link interface of target "httplib::httplib" contains:

    zstd::libzstd

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.

Call Stack (most recent call first):
  /usr/local/httplib/lib/cmake/httplib/httplibConfig.cmake:113 (include)
  ...
```

# Environment

At the time of verification, Ubuntu 24.04 LTS (noble) shows `zstd 1.5.5+dfsg2-2build1` on the official package page: https://packages.ubuntu.com/noble/zstd


yhirose:
@crueter, @DarthGandalf, could you please review this pull request? I don’t have enough expertise on this matter. Thanks!

crueter (approved):
pkg-config is a blight upon my soul.

LGTM, but man, some of this find logic needs to be reworked... or just wait until 24.04 is no longer a supported target :P

DarthGandalf:
pkg-config here is not a problem, but the way how cmake uses either it or not it, resulting in inconsistent dependencies.

DarthGandalf (approved):

yhirose:
@sakurai-ryuhei thanks for your contribution!

sakurai-ryuhei:
@yhirose @crueter @DarthGandalf
Thank you for merging this so quickly!

PR #2454, opened 2026-05-22
Fix SecTrustCopyAnchorCertificates removed in iOS 16+/macOS 13+

gillesdegottex:


yhirose:
@gillesdegottex thanks for the report! I fixed this issue in a different way.

gillesdegottex:
@yhirose Nice! Thanks for the fix!

PR #2463, opened 2026-06-03
add set_hostname_addr_map to WebSocketClient

Hyukya:
WebSockets also need the set_hostname_addr_map function when connecting.

### DIFF
--- a/test/test.cc
+++ b/test/test.cc
@@ -17684,6 +17684,9 @@ TEST(WebSocketTest, SpecifyServerIPAddress_RealHostname) {
 
   ws::WebSocketClient client("ws://localhost:" + std::to_string(port) + "/ws");
   client.set_hostname_addr_map({{"localhost", wrong_ip}});
+  client.set_connection_timeout(1);
+  client.set_read_timeout(1);
+  client.set_write_timeout(1);
 
   EXPECT_FALSE(client.connect());
   EXPECT_FALSE(client.is_open());
--- a/test/test.cc
+++ b/test/test.cc
@@ -17658,8 +17658,7 @@ TEST(WebSocketTest, SpecifyServerIPAddress_AnotherHostname) {
   auto another_host = "example.com";
   auto wrong_ip = "192.0.2.1";
 
-  ws::WebSocketClient client(
-      "ws://localhost:" + std::to_string(port) + "/ws");
+  ws::WebSocketClient client("ws://localhost:" + std::to_string(port) + "/ws");
   client.set_hostname_addr_map({{another_host, wrong_ip}});
 
   ASSERT_TRUE(client.connect());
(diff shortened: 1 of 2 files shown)


yhirose:
@Hyukya could you please add at least one unit test case in `test/test.cc` to verify your change?

Hyukya:
@yhirose I added unit tests for this.

test/test.cc
```diff
@@ -17644,6 +17644,56 @@ TEST(WebSocketTest, ComplexPath) {
...
+  svr.WebSocket("/ws", [](const Request &, ws::WebSocket &ws) {
+    std::string msg;
+    while (ws.read(msg)) {}
+  });
+
+  auto port = svr.bind_to_any_port(HOST);
+  std::thread t([&]() { svr.listen_after_bind(); });
+  svr.wait_until_ready();
+
+  auto wrong_ip = "0.0.0.0";
```
yhirose: Implementation looks good — faithfully mirrors `ClientImpl::set_hostname_addr_map` 👍

One issue in this test: `0.0.0.0` will likely fail on the Linux CI. There, `connect(2)` to `0.0.0.0` is treated as `127.0.0.1`, so it reaches the server bound above → `connect()` succeeds and
`EXPECT_FALSE` fails. It only passes on macOS/BSD because they reject `connect()` to `0.0.0.0`.

Suggest a non-routable RFC 5737 address (never aliases to loopback) plus a short timeout, like `ConnectionErrorTest.InvalidPort`:

```cpp
  auto unreachable_ip = "192.0.2.1"; // RFC 5737, never aliases to loopback

  ws::WebSocketClient client(
      "ws://localhost:" + std::to_string(port) + "/ws");
  client.set_hostname_addr_map({{"localhost", unreachable_ip}});
  client.set_connection_timeout(1, 0); // bound the wait; default is 300s
```

yhirose:
@Hyukya also test/style-check fails.

test/test.cc
Hyukya: I fully understand your feedback.
I will apply an address from the `192.0.2.0/24` range.

I have one additional question.
When implementing this unit test, I referred to `SpecifyServerIPAddressTest.AnotherHostname_Online` and `SpecifyServerIPAddressTest.RealHostname_Online`.
In those tests, `wrong_ip` is set to `0.0.0.0`.
However, I believe that using an IP from the `192.0.2.0/24` range would be more appropriate there as well.
Am I correct?
In those tests, there is no localhost server running, so even if `0.0.0.0` is treated as `127.0.0.1`, the connection still fails, which makes the tests appear valid.
However, if a localhost server exists in the test environment, it seems that the test results could be affected.

Hyukya:
I have made the changes. Could you please review them?

test/test.cc
yhirose: Yes, I think your understanding is correct. Even if the current _Online tests happen to pass with `0.0.0.0` because no localhost server is running, `0.0.0.0` is still platform-dependent and not ideal as an unreachable test address. Using `192.0.2.0/24` there as well would make the tests more robust and the intent clearer. I'll take care of it later. Thanks!

yhirose:
@Hyukya thanks for your contribution!

Hyukya:
@AhnLab-OSS @AhnLab-OSSG

PR #2465, opened 2026-06-03
Ignore range requests for unknown-length streaming responses

florianfischerx:
Clients like ffmpeg/browser-like clients may send `Range: bytes=0-` for live multipart streams. A concrete example is a live MJPEG stream served as multipart/x-mixed-replace.

In such a scenario the resource is not meaningfully byte-range addressable: there is no known total length and no stable "complete representation" to range over. 
What happens currently, is that the range validation will fail and result in  HTTP 416 (Range Not Satisfiable).

The cleaner response shape is a normal streaming response (since 206 doesnt really make much sense) and ignoring the range erquest in this particular case, e.g.:

```
HTTP/1.1 200 OK
Content-Type: multipart/x-mixed-replace; boundary=frame
Cache-Control: no-store
etc
```

This should be permissible and correct, because HTTP range handling [1] is optional as a request modifier, i.e. it is legal to simply respond with 200 OK and the complete representation. 206 is only appropriate when the server honors the range request.

[1] https://httpwg.org/specs/rfc7233.html#rfc.section.3

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -8362,6 +8362,14 @@ inline void coalesce_ranges(Ranges &ranges, size_t content_length) {
 
 inline bool range_error(Request &req, Response &res) {
   if (!req.ranges.empty() && 200 <= res.status && res.status < 300) {
+    if (res.body.empty() && res.content_provider_ && res.content_length_ == 0) {
+      req.ranges.clear();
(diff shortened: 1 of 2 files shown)


yhirose:
@florianfischerx thank you!

PR #2469, opened 2026-06-09
cast to unsigned char before ctype calls in is_hex and is_token_char

metsw24-max:
**Negative char reaches ctype in parsing helpers**

`is_hex`, `is_token_char` and `is_multipart_boundary_chars_valid` hand a raw `char` to `isdigit`/`isalnum`, so a byte above 0x7F sign-extends to a negative `int` and the ctype call becomes undefined (the MSVC debug CRT asserts on it). These helpers run on untrusted input: `is_hex` through request-target percent-decoding and the chunked chunk-size line, `is_token_char` through header field-name validation. Casting to `unsigned char` first brings them in line with the other guarded ctype sites and `case_ignore::to_lower`; ASCII handling is unchanged, so this should read as a portability fix rather than a behaviour change.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -4386,7 +4386,7 @@ inline bool set_socket_opt_time(socket_t sock, int level, int optname,
 }
 
 inline bool is_hex(char c, int &v) {
-  if (isdigit(c)) {
+  if (isdigit(static_cast<unsigned char>(c))) {
     v = c - '0';
     return true;
   } else if ('A' <= c && c <= 'F') {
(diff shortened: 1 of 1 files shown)


yhirose (approved):
LGTM. Correct portability fix — passing a raw `char` to `isdigit`/`isalnum` is UB when `char` is signed and a byte exceeds 0x7F (and trips the MSVC debug CRT assert). All three call sites handle untrusted input, and these are the only remaining raw-`char` ctype sites in the file; the rest already cast or use `unsigned char`. ASCII behavior is unchanged.

yhirose:
@metsw24-max thank you!

PR #2472, opened 2026-06-10
use strict hex parsing in decode_query_component

metsw24-max:
decode_query_component decodes each percent escape by handing a two character substring to std::strtoul, which tolerates a leading sign or whitespace, so input that is not valid hex still parses. reached from untrusted query strings through parse_query_text, "%-0" decodes to a NUL byte, "%-1" to 0xFF and "%+5" or "% 5" to 0x05, whereas decode_uri_component and decode_path_component reject the same sequences via detail::from_hex_to_i and leave them literal. switching this helper to from_hex_to_i brings the three decoders back into agreement and drops the throwaway substring allocation; well formed escapes are untouched, so it should read as a strictness fix rather than a behaviour change.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -9734,11 +9734,9 @@ inline std::string decode_query_component(const std::string &component,
 
   for (size_t i = 0; i < component.size(); i++) {
     if (component[i] == '%' && i + 2 < component.size()) {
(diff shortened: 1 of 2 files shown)


yhirose:
@metsw24-max thank you for your fine contribution!

PR #2475, opened 2026-06-13
make from_chars(double) parsing locale-independent

metsw24-max:
**Locale-dependent quality value parsing**

`detail::from_chars(double)` hands off to `std::strtod`, which reads the decimal point from the global C locale, whereas the integer overload beside it is written to stay locale-independent. Once an embedding process has run `setlocale(LC_ALL, "")` into a locale that uses ',' (de, fr and others), an untrusted `Accept`/`Accept-Encoding` header such as `gzip;q=0.8` is parsed as `q=0` because `strtod` stops at the '.', so the weight collapses to "not acceptable" and the server quietly drops gzip and skews content negotiation. The fix translates '.' to the active locale's separator before the call, restoring the locale-independent intent without pulling in a different parser. The regression test pins q-value ordering under a comma-decimal locale and skips where none is installed.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -312,6 +312,7 @@ using socket_t = int;
 #include <cctype>
 #include <chrono>
 #include <climits>
+#include <clocale>
 #include <condition_variable>
 #include <cstdlib>
 #include <cstring>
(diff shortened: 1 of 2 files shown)


yhirose:
@metsw24-max thanks for the pull request! I implemented it in a better way.

PR #2476, opened 2026-06-14
skip CN fallback for IP hosts in mbedTLS and wolfSSL verify_hostname

metsw24-max:
**IP-literal hosts can be matched against the certificate CN**

On the Mbed TLS and wolfSSL backends `tls::verify_hostname` checks the iPAddress SANs for an IP host and then, when none match, falls back to comparing the IP against the certificate Common Name. RFC 9110 only allows an IP identity to come from an iPAddress SAN, and the OpenSSL backend already behaves this way (`X509_check_ip` never consults the CN), so a certificate that carries the address in its CN but omits a matching SAN is wrongly accepted here. Gated the CN fallback on the host not being an IP literal in both backends, which should bring them in line with OpenSSL and with the RFC 2818 note already in `verify_host`.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -18600,10 +18600,12 @@ inline bool verify_hostname(cert_t cert, const char *hostname) {
     san = san->next;
   }
 
-  // Fallback: Check Common Name (CN) in subject
+  // Fallback: Check Common Name (CN) in subject. Skipped for IP-literal hosts:
+  // an IP identity is only valid via an iPAddress SAN, never the CN (RFC 9110;
+  // the OpenSSL backend's X509_check_ip behaves the same way).
   char cn[256];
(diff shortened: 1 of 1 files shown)


yhirose:
@metsw24-max thanks for the report. I fixed it in a better way at https://github.com/yhirose/cpp-httplib/commit/ba390f23994cad46ebcd1497414d22a2ebf736de

yhirose:
@metsw24-max, by the way, from now on, please make sure that any pull request you send includes at least one unit test for each change. Thanks for your understanding!

metsw24-max:
Makes sense. Folding in IPv6 and actually matching the v6 SANs is the more complete fix, and dropping the dead verify_host* code is a nice tidy-up. Noted on the tests as well; I'll include a unit test with each change from here on.

PR #2477, opened 2026-06-19
use an unsigned accumulator in base64_encode

metsw24-max:
base64_encode folds each input byte into a signed int and shifts it left by eight every round; once four bytes are in, the top bit is set and the next shift left-shifts a negative int, which is undefined behaviour (ubsan reports a left shift of a negative value, and the same shift recurs in the trailing six-bit fixup). the helper runs on the server's websocket accept-key path over the sha1 digest and in the basic-auth header builder, so it is reachable from request handling. making the accumulator a uint32_t removes the ub and keeps the output byte-for-byte identical, since every emitted six-bit group is masked with 0x3f and is independent of the accumulator's signedness. the added test pins the rfc 4648 vectors plus a high-byte input that keeps the top bit set across rounds.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -4665,7 +4665,9 @@ inline std::string base64_encode(const std::string &in) {
   std::string out;
   out.reserve(in.size());
 
-  auto val = 0;
+  // Unsigned: once four bytes are folded in the top bit is set, so the next
+  // `val << 8` would left-shift a negative int (undefined behaviour).
+  uint32_t val = 0;
   auto valb = -6;
 
   for (auto c : in) {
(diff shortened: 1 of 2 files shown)


yhirose:
Looks like there are quite a few errors...

metsw24-max:
Those were all the same failure: the new test calls `detail::base64_encode`, but in the split build `split.py` strips `inline` and moves the definition into httplib.cc, so it isn't declared in the public header. I've forward-declared it in the test the same way `getaddrinfo_with_timeout` already is, so it resolves in both the header-only and split builds. Confirmed both compile and the test passes locally.

yhirose:
@metsw24-max thanks for your contribution!

PR #2478, opened 2026-06-21
escape crlf and quotes in multipart form-data field headers

metsw24-max:
**Part-header injection in multipart serialisation**

`serialize_multipart_formdata_item_begin` writes each part's `Content-Disposition` and `Content-Type` lines by concatenating `item.name`, `item.filename` and `item.content_type` straight into the header, with no escaping. A name or filename carrying a CR/LF therefore injects further part headers (and a body), and one carrying a `"` closes the quoted-string early. When an application passes a value it received from elsewhere (a forwarded upload filename being the common case) the caller cannot prevent this from the outside, so the escaping belongs in the writer.

A filename of `evil\r\nContent-Type: text/x-injected\r\n\r\nSMUGGLED` currently produces a part with an attacker-chosen `Content-Type` and body; `cli.Post("/x", items)` is enough to reproduce it.

The fix escapes CR, LF and `"` to `%0D`/`%0A`/`%22` (RFC 7578 §5.1.1, matching the WHATWG form-data escaping browsers use) inside the serialiser, so every path that goes through it is covered. Well-formed values are unchanged. I have leant towards encoding rather than rejecting to keep the existing string-returning API and behaviour stable, though rejecting outright would also be defensible if you would prefer that.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -8314,18 +8314,37 @@ inline bool is_multipart_boundary_chars_valid(const std::string &boundary) {
   return valid;
 }
 
+// RFC 7578 §5.1.1: escape CR, LF and the double quote so a caller-supplied
+// field name or filename cannot break out of the quoted-string or inject extra
+// part headers. Mirrors the WHATWG form-data escaping used by browsers.
+inline std::string escape_multipart_field(const std::string &s) {
+  std::string r;
+  r.reserve(s.size());
+  for (auto c : s) {
+    switch (c) {
+    case '\r': r += "%0D"; break;
+    case '\n': r += "%0A"; break;
+    case '"': r += "%22"; break;
+    default: r += c; break;
+    }
+  }
+  return r;
+}
+
 template <typename T>
 inline std::string
 serialize_multipart_formdata_item_begin(const T &item,
                                         const std::string &boundary) {
   std::string body = "--" + boundary + "\r\n";
-  body += "Content-Disposition: form-data; name=\"" + item.name + "\"";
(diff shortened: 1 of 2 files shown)


yhirose:
@metsw24-max, thanks for the pul request. Could you fix the CI errors?

Also escaping " in content_type is overly aggressive. The Content-Type: header value is not inside a quoted-string context, so a legitimate value like multipart/mixed; boundary="xyz" gets corrupted into boundary=%22xyz%22. For content_type, only CR/LF need to be neutralized. The " escaping should be limited to name and filename.

yhirose:
Following up on my comment above — I went ahead and addressed this on master myself, since I wanted to get the fix in promptly:

- 8149bb3 escapes `"`, CR and LF in `name` and `filename` (WHATWG form-data escaping: `%22`/`%0D`/`%0A`), as in your PR.
- f5c8c98 handles `content_type` separately, escaping only CR and LF and leaving `"` intact for the reason mentioned above (values like `multipart/mixed; boundary="xyz"` must survive unchanged).

The tests on master also exercise the full round trip through the server-side parser to confirm no headers can be injected.

Closing this as superseded by those commits. Thanks again for finding and reporting the injection vector!

metsw24-max:
Sounds good, and you're right about content_type — it isn't in a quoted-string context, so leaving " intact there (multipart/mixed; boundary="xyz" and the like) is the correct call. The CR/LF-only split for content_type plus the round-trip parser tests look spot on. Glad the injection vector is closed off.

PR #2479, opened 2026-06-23
Send query string verbatim when path encoding is disabled

sabify:
set_path_encode(false) only suppressed encoding of the path part. The query part was always run through normalize_query_string(), which decodes then re-encodes each key/value pair regardless of the flag.

That round-trip is lossy for pre-encoded payloads: re-encoding emits sub-delimiters literally (%2C->",", %24->"$", %3B->";", ...) and turns %20 into "+", which a strict RFC 3986 server decodes back as "+" (0x2B) rather than a space (0x20), corrupting binary query data.

Honor path_encode_ for the query as well: when disabled, append the caller-supplied query verbatim. Add a regression test asserting on the raw request target, since the server decodes "+" as space and would otherwise mask the difference.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -13657,9 +13657,18 @@ inline bool ClientImpl::write_request(Stream &strm, Request &req,
 
     if (!query_part.empty()) {
       // Normalize the query string (decode then re-encode) while preserving
-      // the original parameter order.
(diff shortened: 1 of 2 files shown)


yhirose:
@sabify thank you for your contribution!

PR #2483, opened 2026-07-01
Fix locale-dependent alpha-numeric checks

### LINKED ISSUE #2482: encode_query_component depends on the current locale on Mac
`httplib::encode_query_component` encodes `Ÿ` as `xc5%B8` on Mac, when the locale is set to `en_US.UTF-8`. It should be encoded as `%C5%B8` (this works correctly on Linux with the same locale).

The reason this happens is that `std::isalnum` is dependent on the locale (https://en.cppreference.com/cpp/string/byte/isalnum) and on Mac it returns a truthy value for the byte `\xC5`. Minimal `std::isalnum` test that works on Mac and Linux: https://github.com/ungive/httplib-bug-encode-query-component-isalnum.

This leads to the following code to incorrectly encode the UTF-8 multibyte character:

https://github.com/yhirose/cpp-httplib/blob/9d159bb412393eb7bc821fb193999cc44ef32636/httplib.h#L9730-L9732

Output on Mac:

```
% ./isalnum_test 
Locale: en_US.UTF-8

0x41  isalnum=1  isalpha=1  isdigit=0
0x35  isalnum=1  isalpha=0  isdigit=1
0xC5  isalnum=1  isalpha=1  isdigit=0
0xB8  isalnum=0  isalpha=0  isdigit=0

UTF-8 bytes of Ÿ:
0xC5
0xB8
```

Output on Linux:

```
Locale: en_US.UTF-8

0x41  isalnum=8  isalpha=1024  isdigit=0
0x35  isalnum=8  isalpha=0  isdigit=1
0xC5  isalnum=0  isalpha=0  isdigit=0
0xB8  isalnum=0  isalpha=0  isdigit=0

UTF-8 bytes of Ÿ:
0xC5
0xB8
```

ungive:
std::isalnum depends on the current locale. This becomes apparent on Mac, where e.g. the byte \xC5 may be treated as alpha-numeric, even though it is not.

Fixes #2482.

Note that I naively replaced all uses of `std::isalnum` with `detail::is_ascii_alnum`. If this is not correct in any place, let me know!

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -8302,11 +8302,17 @@ inline std::string make_multipart_data_boundary() {
   return "--cpp-httplib-multipart-data-" + detail::random_string(16);
 }
 
+inline bool is_ascii_alnum(unsigned char c) {
+  return (c >= 'A' && c <= 'Z') || (c >= 'a' && c <= 'z') ||
+         (c >= '0' && c <= '9');
+}
+
 inline bool is_multipart_boundary_chars_valid(const std::string &boundary) {
   auto valid = true;
   for (size_t i = 0; i < boundary.size(); i++) {
     auto c = boundary[i];
-    if (!std::isalnum(static_cast<unsigned char>(c)) && c != '-' && c != '_') {
+    if (!is_ascii_alnum(static_cast<unsigned char>(c)) && c != '-' &&
+        c != '_') {
       valid = false;
       break;
     }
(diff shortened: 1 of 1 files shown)


yhirose:
@ungive could you fix the style error and add at least one unit test in test/test.cc? Thanks!

yhirose:
@ungive Thank you for the report and the PR — the locale dependency you identified was real and worth fixing. It's the same bug class as #2475, where `std::strtod` consulted the global C locale inside `from_chars`, so I wanted to address both with one consistent mechanism.

Rather than wait on the CI fixes here, I went ahead and implemented this on master in af6ed68, extending your approach a bit:

- `is_ascii_alnum` is now split into `is_ascii_digit` / `is_ascii_alpha` / `is_ascii_alnum` and placed above the split border, so both the header-area code and the implementation area can share them.
- I was careful to eliminate every duplicate of this logic rather than just the `std::isalnum` sites your PR covered: all `std::isdigit` uses and the hand-rolled `'0' <= c && c <= '9'` range checks scattered across the codebase — including the digit loops in the locale-independent `from_chars` from #2475, URL parsing, and `parse_ipv4` — now go through the same single set of helpers.
- With those replaced, nothing uses `<cctype>` anymore, so the include is gone entirely — removing this whole class of locale surprises at the root.
- Added the tests this needed: an exhaustive check that all 256 byte values classify by the ASCII grammar alone, a regression check for the `0xC5` byte you reported, and an end-to-end check that non-ASCII bytes are always percent-encoded in URIs.

Closing this as superseded by that commit. Thanks again for tracking this down!

PR #2484, opened 2026-07-01
meson: fix build failure on glibc >= 2.34 without standalone libanl

setarcos:
On glibc >= 2.34, getaddrinfo_a is no longer provided by a standalone libanl shared library — it is built directly into libc.  Newer architectures (e.g. loongarch64, riscv64) have never shipped a separate libanl, causing the meson build to fail with:

    C++ shared or static library 'anl' not found

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -107,7 +107,12 @@ if host_machine.system() == 'windows'
 elif host_machine.system() == 'darwin'
   async_ns_dep = dependency('appleframeworks', modules: ['CFNetwork', 'CoreFoundation'], required: async_ns_opt)
 else
-  async_ns_dep = cxx.find_library('anl', required: async_ns_opt)
+  has_gai_a = cxx.has_function('getaddrinfo_a', args: '-D_GNU_SOURCE')
+  if has_gai_a
+    async_ns_dep = declare_dependency()
+  else
+    async_ns_dep = cxx.find_library('anl', required: async_ns_opt)
+  endif
 endif
 
 if async_ns_dep.found()


yhirose:
@Tachi107 could you please review this pull request for me?

meson.build
```diff
@@ -107,7 +107,11 @@ if host_machine.system() == 'windows'
 elif host_machine.system() == 'darwin'
   async_ns_dep = dependency('appleframeworks', modules: ['CFNetwork', 'CoreFoundation'], required: async_ns_opt)
 else
-  async_ns_dep = cxx.find_library('anl', required: async_ns_opt)
+  async_ns_dep = cxx.find_library('anl', required: false)
+  if not async_ns_dep.found()
+    # glibc >= 2.34 has getaddrinfo_a built into libc, no separate libanl needed
+    async_ns_dep = cxx.has_function('getaddrinfo_a') ? declare_dependency() : dependency('', required: async_ns_opt)
+  endif
```
Tachi107: I think it'd make more sense to invert the logic. i.e. look for getaddrinfo_a (since it's what we need), and if absent, require libanl.

Also, `dependency('', found: false)` is the usual way of having an "useless" dependency, as per [meson docs](https://mesonbuild.com/Reference-manual_functions_dependency.html).

```suggestion
  has_gai_a = cxx.has_function('getaddrinfo_a', args: '-D_GNU_SOURCE')
  if not has_gai_a
    async_ns_dep = cxx.find_library('anl', required: async_ns_opt)
  else
    async_ns_dep = dependency('', found: false)
  fi
```

Tachi107:
@yhirose I think it'd make sense to add me as a "code owner" for the various `meson.build` files, so that I get automatically pinged for patches targetting them. I have never used this GitHub feature though, so can't suggest how to do that :)

meson.build
setarcos: Thanks for the suggestion — inverting the logic does read much better. One small note: Instead `dependency('', found: false)`, I used `declare_dependency()` instead, which creates an empty dependency that reports as found (so the -DCPPHTTPLIB_USE_NON_BLOCKING_GETADDRINFO flag still gets added when getaddrinfo_a is available in libc).

yhirose:
@setarcos @Tachi107 thanks all!

PR #2497, opened 2026-07-15
Limit the number of header lines per multipart form-data part

superm1:
The multipart/form-data parser bounded the length of each individual header line via `CPPHTTPLIB_HEADER_MAX_LENGTH` but did not cap how many header lines a single part could contain. Within the overall payload limit, a client could pack a large number of small header lines into one part, each of which is parsed twice, inflating CPU usage.

This adds a per-part header-line counter that resets when a new entry begins and aborts parsing once it reaches `CPPHTTPLIB_HEADER_MAX_COUNT`, mirroring the limit already enforced by `read_headers()` for request headers.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -8095,6 +8095,13 @@ public:
             break;
           }
 
+          // Check header count limit
+          if (header_count_ >= CPPHTTPLIB_HEADER_MAX_COUNT) {
+            is_valid_ = false;
+            return false;
+          }
+          header_count_++;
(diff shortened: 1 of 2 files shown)


yhirose:
@superm1 thanks for the PR. Could you please add at least one unit test?

superm1:
Added!

yhirose:
@superm1 thanks for this contribution!

PR #2498, opened 2026-07-17
32-bit container CI

daveisfera:
I'm using `cpp-httplib` on an embedded device that uses 32-bit, so this PR improves testing on 32-bit platforms to help ensure it's reliable there

### DIFF
--- a/.github/workflows/test-win32.yml
+++ b/.github/workflows/test-win32.yml
@@ -1,4 +1,4 @@
-name: 32-bit Build Test
+name: Win32 Build Test
 
 on:
   push:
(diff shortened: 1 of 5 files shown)


yhirose:
@daveisfera, thank for the PR, but this PR bundles too many separate concerns to accept as-is. The one piece I'd welcome is adding UBSan to the existing i386 job that runs on the amd64 hardware I already have, so I can actually reproduce and fix anything it catches. I can't commit to armhf real-device testing or promoting 32-bit to a supported tier, because I have no 32-bit hardware and wouldn't be able to support failures I can't reproduce. If you'd like, please open a small PR that just adds -fsanitize=undefined to the i386 matrix. (The parse_http_date Y2038 change is interesting but affects all platforms that needs its own PR.)

yhirose:
You can find more information about 32-bit platform support at the beginning of the README.

daveisfera:
I'm using it on a 32-bit ARM device and would like to help improve the support for that use case. You should be able to use containers to emulate the 32-bit support in many cases. There are issues with QEMU not always playing nice in those cases, but it is a more robust approach and is something that can be done. I'd be glad to help with that to ensure that the coverage and support remains robust.

daveisfera:
Never mind. The rejection prompted me to look into `libcurl` as an option and I can get a similar sized build when I strip it down, so I'll just run with that and you don't have to worry about 32-bit support.

PR #2500, opened 2026-07-18
Match chunked as the final transfer coding, order-independently

yhirose:
Based on #2487 by @metsw24-max — reworked implementation, so opening separately to make the diff reviewable.

## Problem

`is_chunked_transfer_encoding` compared the whole `Transfer-Encoding` field value against `"chunked"`, so a message whose final coding is chunked but which names another coding first (`"gzip, chunked"`, valid under RFC 9112 §6.1 where chunked must be the final coding) was read as unframed. On a keep-alive server the body is then left in the socket and parsed as the next request (smuggling). `open_stream` repeated the check case-sensitively with a raw `==`, desyncing the client stream the same way.

## Change

- Match the **last coding token** of the `Transfer-Encoding` value case-insensitively instead of comparing the whole field, and route `open_stream` through the same helper so server and streaming client agree.
- Scan **all** `Transfer-Encoding` lines rather than only index 0. `Headers` is an `unordered_multimap` whose duplicate-key iteration order is not portable (this is what broke CI on the original PR: libstdc++ and libc++ disagree), so the final coding of a multi-line field cannot be determined reliably. In that ambiguous case we **fail safe** — any multi-line field that names `chunked` is treated as chunked. A mis-parse only closes the connection, whereas the opposite error enables smuggling.

## Tests

`ChunkedTransferEncodingTest` covers single-coding (case-insensitive), `chunked` as the final coding of a list, non-final `chunked`, absent/empty, and **order-independent** multi-line cases. Builds and passes on both the header-only and split (`test_split`) builds.

## Difference from #2487

#2487 read only the first `Transfer-Encoding` line via `get_header_value(..., 0)` and dropped its multi-line test because of the `unordered_multimap` ordering issue. This version handles the multiple-line case deterministically by failing safe, so no ordering assumption is needed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -7386,8 +7386,46 @@ inline ReadContentResult read_content_chunked(Stream &strm, T &x,
 }
 
 inline bool is_chunked_transfer_encoding(const Headers &headers) {
-  return case_ignore::equal(
-      get_header_value(headers, "Transfer-Encoding", "", 0), "chunked");
+  // RFC 9112 6.1: a message is framed with the chunked coding when "chunked"
+  // is the final transfer coding. A single field value may list several
+  // codings ("gzip, chunked"), and the list may be split across multiple
+  // Transfer-Encoding header lines (RFC 9110 5.3). Match the last coding token
+  // case-insensitively rather than comparing the whole value against "chunked".
+  //
+  // Security: reading a chunked message as unframed leaves its body in the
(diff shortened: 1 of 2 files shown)


yhirose:
@metsw24-max thanks for sending #2487. This PR is the revised version that I made with Opus. I think this is a more comprehensive solution. Please let me know if you have any question.

PR #2501, opened 2026-07-18
Reject CR/LF in the request target and fail the request cleanly

yhirose:
Supersedes #2495 (thanks @metsw24-max — the first two commits are theirs, unchanged).

## What the original PR does

The client builds each request line in `write_request_line` by concatenating the method, the target, and `" HTTP/1.1\r\n"`, with no check on the target. A target carrying CR/LF splits the request line and injects a header or a whole request. The neighbouring writers already guard this (`check_and_write_headers` over header values, `perform_websocket_handshake` over the path), but the ordinary request path was left out.

It becomes reachable when following a redirect with path encoding disabled: `ClientImpl::redirect` decodes the `Location` path through `decode_path_component`, so a wire-safe `%0D%0A` turns into a raw CR/LF, and under `set_path_encode(false)` that decoded target reaches `write_request_line` verbatim. The guard rejects a target that fails `is_field_value`, matching the other two writers.

## What this PR adds on top

`write_request_line` now returns `-1` on a rejected target, but **`ClientImpl::write_request` ignored that return value.** A rejected target was silently dropped, producing a request-line-less (headers-only) request — and the client reported it as a **successful** send (`res.error() == Success`). This is the primary path (all normal requests and the `CONNECT` target flow through `write_request`), i.e. exactly the path the original PR is about.

- Check the return value in `write_request` and fail with `Error::Write`, matching `ClientImpl::open_stream`, which already checks it.
- Add `RequestLineInjectionTest.ClientRejectsCRLFTargetEndToEnd`: with `set_path_encode(false)`, `cli.Get("/a\r\nInjected: pwned")` must return `false` with `Error::Write` rather than reaching the wire. Verified to fail without the call-site fix.

Note: the guard blocks CR/LF (the injection vector). `is_field_value` still permits interior SP/HTAB, consistent with the other two writers; normal encoding prevents raw spaces in targets.

Tested: full suite green (734 tests) in both header-only and split builds.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -7504,6 +7504,13 @@ bool read_content(Stream &strm, T &x, size_t payload_max_length, int &status,
 
 inline ssize_t write_request_line(Stream &strm, const std::string &method,
                                   const std::string &path) {
+  // A request target must not carry CR/LF (or other control octets); otherwise
+  // a value smuggled into it splits the request line and injects headers or a
+  // whole request. The same field-value check already guards header values in
+  // check_and_write_headers and the request target in
+  // perform_websocket_handshake; apply it here too.
+  if (!fields::is_field_value(path)) { return -1; }
+
   std::string s = method;
   s += ' ';
   s += path;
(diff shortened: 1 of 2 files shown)


yhirose:
@metsw24-max this is another PR that revised your PR #2495. I fixed an issue that the return code from `write_request_line` is ignored. Also it adds an end-to-end test.

PR #2503, opened 2026-07-20
scan X-Forwarded-For right-to-left in get_client_ip

metsw24-max:
**Client-spoofable source address in get_client_ip**

With `set_trusted_proxies` configured, `req.remote_addr` is derived by scanning `X-Forwarded-For` left to right and returning the entry just before the first trusted-proxy address (or the leftmost entry when none is listed). Each hop appends the address it received the request from, so the trustworthy entries sit on the right while the left is whatever the client chose to send. A client can therefore pin the result to any value with `X-Forwarded-For: <forged>, <trusted-proxy-ip>`: the real proxy still appends the client's true address on the right, but the left-to-right scan stops at the injected trusted-proxy token and returns `<forged>`. Because `remote_addr` tends to back IP allowlists, rate limits and audit logs, that is a source-address spoof.

The scan now runs right to left and returns the first address that is not a trusted proxy, i.e. the furthest hop still written by our own infrastructure. Well-formed proxy chains resolve to the same client as before; the only behavioural change is the "no trusted proxy in the header" fallback, which now prefers the rightmost address over the client-controlled leftmost one. The existing fallback test is updated to match and a new test covers the spoof case. The helper keeps its shape, so it should stay easy to follow.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -12420,25 +12420,25 @@ get_client_ip(const std::string &x_forwarded_for,
   // caller can fall back to the connection-level remote address.
   if (ip_list.empty()) { return std::string(); }
 
-  for (size_t i = 0; i < ip_list.size(); ++i) {
-    auto ip = ip_list[i];
+  // Each hop appends the address it received the request from, so the rightmost
+  // entries are the ones written by our own infrastructure while the leftmost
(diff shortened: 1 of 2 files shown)


yhirose:
@metsw24-max thanks for your contribution!

Issue #2450, opened 2026-05-13
Server closes keep-alive connection after DELETE request without Content-Length (drain bug)


Beewe:
# Disclaimer
This is my basically my first proper issue reporting to a repo but I tried to describe the problem as best as I could.

## Summary

When a `DELETE` request arrives **without** a `Content-Length` header on a keep-alive connection, I think the server's post-response body drain logic calls `read_content_without_length()`, which reads until EOF — consuming the next pipelined request's bytes and destroying the connection of the succeeding request.

## Environment

- cpp-httplib version: **0.42.0**
- OS: Debian 13 (Linux 6.x, x86_64)
- Compiler: GCC 14.2.0

## Minimal working example

**Server (C++):**

```cpp
#include "httplib.h"
#include <iostream>

int main() {
    httplib::Server server;

    server.Delete("/items/:id", [](const httplib::Request&, httplib::Response& res) {
        res.status = 204;
    });

    std::cout << "Listening on http://localhost:9999\n";
    server.listen("0.0.0.0", 9999);
}
```

**Client (Python):**

```python
import requests

session = requests.Session()

for i in range(1, 6):
    req = requests.Request("DELETE", "http://localhost:9999/items/1")
    prepared = session.prepare_request(req)
    prepared.headers.pop("Content-Length", None)
    try:
        resp = session.send(prepared)
        print(f"Request {i}: status={resp.status_code}")
    except Exception as e:
        print(f"Request {i}: ERROR — {e}")
```

**Output:**

```
Request 1: status=204
Request 2: ERROR — ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))
Request 3: status=204
Request 4: ERROR — ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))
Request 5: status=204
```

yhirose:
@Beewe thanks for the report!

PR #2504, opened 2026-07-20
Skip request body drain when connection will close

emreay-:
## Summary

Skip draining an unconsumed framed request body when the response has already committed the connection to close.

The post-response drain prevents unread body bytes from being parsed as the next request on a keep-alive connection. When the response contains `Connection: close`, there is no next request to protect.

## Problem

If a `ContentReader` aborts an unterminated chunked upload, the current drain waits for the terminal chunk after sending the response. A client that keeps producing chunks continually resets the read timeout, so:

- the announced connection close never occurs;
- the client receives no transport-level failure signal; and
- a server worker remains occupied discarding data.

This also means the server can send `Connection: close` while continuing to consume the request indefinitely.

## Fix

Use the finalized response header as the close decision:

```cpp
if (!req.body_consumed_ && detail::has_framed_body(req)) {
  if (res.get_header_value("Connection") == "close") {
    connection_closed = true;
  } else {
    // Drain before reusing the connection.
  }
}
```

`write_response_core` already sets this header for keep-alive exhaustion, request-directed closure, handler-directed closure, and error responses. Reusable connections retain the existing drain behavior.

## Relationship to #2450

This is a follow-up to #2450:

- #2450 prevented the drain from reading an unframed request, where it could consume the next request.
- This change prevents the drain from running when the connection will close, where no next request exists.

Both changes restrict the drain to cases where it can serve its request-smuggling-prevention purpose.

## Test

A raw-socket regression test sends an unterminated chunked request whose `ContentReader` rejects the first chunk. It verifies that the client receives the `409 Connection: close` response followed by EOF.

On unpatched `master`, the final `recv` times out because the server remains in the drain. With this change, it returns `0` immediately.

The new test and the existing #2450 regression both pass on Windows with MSVC.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -12715,13 +12715,18 @@ Server::process_request(Stream &strm, const std::string &remote_addr,
 
   // Drain any unconsumed framed body to prevent request smuggling on
   // keep-alive. Without framing there is no body to drain — reading would
-  // consume the next request (issue #2450).
+  // consume the next request (issue #2450). If the response has committed the
+  // connection to close, there is no next request to protect.
   if (!req.body_consumed_ && detail::has_framed_body(req)) {
-    int dummy_status;
-    if (!detail::read_content(
-            strm, req, payload_max_length_, dummy_status, nullptr,
-            [](const char *, size_t, size_t, size_t) { return true; }, false)) {
+    if (res.get_header_value("Connection") == "close") {
       connection_closed = true;
+    } else {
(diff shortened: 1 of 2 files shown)


yhirose:
@emreay- thanks for this PR. Could you please take a look at the `test/style-check` error?

yhirose:
@emreay- thanks for this fine contribution!

PR #2505, opened 2026-07-23
skip invalid fields in write_headers to prevent response splitting

metsw24-max:
The server serialises `res.headers` straight through `write_headers`, which formats each entry as `name: value` followed by CRLF without checking either part. The client side already guards this: `req.headers` go through `check_and_write_headers`, and the request line and websocket handshake run the same `is_field_value` check, so `write_headers` was the one output path left unguarded. `Response::headers` is a public field, and an application that reflects a request-derived value into it (a fairly common pattern) can carry a CR/LF into the response, splitting the header block and injecting further headers or a body. I noticed the gap while reading the recent chunked-trailer fix, which closed the same class for trailers but did not reach the ordinary header writer. The change skips any field whose name or value fails validation, which keeps the behaviour consistent with `set_header` rather than dropping the whole response. Fields the library emits itself all go through `set_header` and stay valid, so only directly-injected ones are affected. The regression test drives a handler that plants CR/LF in both a header name and value via `res.headers` and checks the split content never reaches the wire.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -7594,6 +7594,15 @@ inline ssize_t write_response_line(Stream &strm, int status) {
 inline ssize_t write_headers(Stream &strm, const Headers &headers) {
   ssize_t write_len = 0;
   for (const auto &x : headers) {
+    // Skip fields with invalid names or values to prevent response splitting
+    // via CR/LF injection, matching set_header(). The client validates request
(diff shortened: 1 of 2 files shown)


yhirose:
@metsw24-max thanks!

PR #2508, opened 2026-07-26
allow setting socket manually

sebasjm:
systemd services don't own the socket and should not shut it down

This PR add 2 features on **Server**:
 * _set_socket_ : to specify an already created socket instead of using bind
 * _set_shutdown_socket_ : to change the default behavior of always shutting down the current socket. Default to true.

Defaults makes it backward compatible.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -2013,6 +2013,7 @@ private:
   int address_family_ = AF_UNSPEC;
   bool tcp_nodelay_ = CPPHTTPLIB_TCP_NODELAY;
   bool ipv6_v6only_ = CPPHTTPLIB_IPV6_V6ONLY;
+  bool shutdown_socket = true;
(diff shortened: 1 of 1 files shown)


sebasjm:
More on systemd

https://www.freedesktop.org/software/systemd/man/latest/systemd.socket.html

> A service listening on an AF_UNIX socket may, but does not need to, call [close(2)](https://man7.org/linux/man-pages/man2/close.2.html) on the received socket before exiting. However, it must not unlink the socket from a file system. It should not invoke [shutdown(2)](https://man7.org/linux/man-pages/man2/shutdown.2.html) on sockets it got with Accept=no, but it may do so for sockets it got with Accept=yes set.

sebasjm:
I accidentally created from the wrong branch. I will create a new PR

PR #2510, opened 2026-07-28
only apply Range to a 206 response in write_content_with_provider

metsw24-max:
**Unvalidated Range reaches the content provider on non-2xx responses**

`write_content_with_provider` slices the representation whenever `req.ranges` is non-empty, but `apply_ranges` only slices for a 206 and `detail::range_error` only validates the ranges for a 2xx status, so a handler answering 403 through a content provider hands the client's raw offset straight down (ASAN: heap-buffer-overflow, READ of 201 bytes past a 7-byte buffer, which then goes out on the socket; the bounds asserts in `get_range_offset_and_length` are compiled out under NDEBUG). Gating on 206 to match `apply_ranges` also settles the 2xx-but-not-206 case, where the body written was shorter than the Content-Length already sent, and the added test covers both.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -11777,7 +11777,16 @@ Server::write_content_with_provider(Stream &strm, const Request &req,
   };
 
   if (res.content_length_ > 0) {
-    if (req.ranges.empty()) {
+    // Only a 206 response is served as a partial representation. That is the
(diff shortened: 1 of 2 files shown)


yhirose:
@metsw24-max thanks!

PR #2511, opened 2026-07-29
fail mmap::open when ::mmap returns MAP_FAILED

metsw24-max:
the posix branch of `detail::mmap::open` only tests `MAP_FAILED` as part of the empty-file case, so a failed mapping of a non-empty file leaves the sentinel in `addr_` and `open()` still returns true. `is_open()` compares `addr_` against nullptr alone, so callers pass the guard and `data()` hands back `(const char *)-1`, which `handle_file_request` and the `set_file_content` path in `process_request` both feed to a content provider as `data() + offset`, with `offset` coming from the client's Range header and the length bounded by the fstat size rather than by any mapping. the windows branch already closes and fails on every mapping error, so this mostly brings posix into line with it. mapping a directory is the easiest way to reach the failure, since `::open` and `fstat` succeed there and `::mmap` does not, and that is what the added test uses.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -5568,6 +5568,15 @@ inline bool mmap::open(const char *path) {
     is_open_empty_file = true;
     return false;
   }
+
+  // A failed mapping must not be left in `addr_`: `is_open()` only compares it
+  // against nullptr, so the MAP_FAILED sentinel would pass and `data()` would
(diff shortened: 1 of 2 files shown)


yhirose:
@metsw24-max thanks!

Issue #2452, opened 2026-05-18
feat: add set_socket() method for pre-bound sockets (systemd socket activation)


123Satyajeet123:
## Problem

There's no way to pass a pre-bound socket to httplib::Server. This is needed for systemd socket activation, where the init system opens the socket and passes it as fd 3.

## Proposed API

```cpp
bool Server::set_socket(socket_t sock) { svr_sock_ = sock; return true; }
```

One line. Allows external code to inject a socket before calling `listen_after_bind()`.

## Use case

systemd socket activation: service uses 0 RAM when idle, starts on first connection. Standard on Linux for on-demand services.

```cpp
const char *fds = getenv("LISTEN_FDS");
if (fds && atoi(fds) >= 1) {
    svr.set_socket(3); // systemd's pre-bound socket
    svr.listen_after_bind();
}
```

## Context

I submitted a PR to llama.cpp (which vendors cpp-httplib) and was told vendor changes must go upstream first: https://github.com/ggml-org/llama.cpp/pull/23214

Working implementation: https://github.com/123Satyajeet123/bitnet-server

yhirose:
@123Satyajeet123 Thank you for the request. However, this change is very risky and could cause various problems if used incorrectly.

The reason is that the semantic meaning of the fd passed by the user must match exactly what the cpp-httplib server instance expects. Also, related to this, it is a difficult issue to decide whether to transfer the fd’s lifetime to cpp-httplib or have the user handle the cleanup.

In short, there are many considerations behind this single line of code. Furthermore, all of these cases must be thoroughly verified through comprehensive unit tests. Therefore, I am not planning to work on this issue at this point. Thanks for your understanding!

PR #2512, opened 2026-07-30
allow setting socket manually

sebasjm:
systemd services don't own the socket and should not shut it down

This PR add 2 features on Server:

**set_socket** : to specify an already created socket instead of using bind
**set_shutdown_socket** : to change the default behavior of always shutting down the current socket. Default to true.
Defaults makes it backward compatible.
More on systemd

https://www.freedesktop.org/software/systemd/man/latest/systemd.socket.html

> A service listening on an AF_UNIX socket may, but does not need to, call close(2) on the received socket before exiting. However, it must not unlink the socket from a file system. It should not invoke shutdown(2) on sockets it got with Accept=no, but it may do so for sockets it got with Accept=yes set.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -1821,6 +1821,8 @@ public:
   Server &set_address_family(int family);
   Server &set_tcp_nodelay(bool on);
   Server &set_ipv6_v6only(bool on);
+  Server &set_shutdown_socket(bool shutdown_socket);
+  Server &set_socket(const socket_t socket);
   Server &set_socket_options(SocketOptions socket_options);
 
   Server &set_default_headers(Headers headers);
@@ -1893,6 +1895,7 @@ protected:
   time_t websocket_ping_interval_sec_ =
       CPPHTTPLIB_WEBSOCKET_PING_INTERVAL_SECOND;
   int websocket_max_missed_pongs_ = CPPHTTPLIB_WEBSOCKET_MAX_MISSED_PONGS;
+  bool shutdown_socket_ = true;
 
 private:
   using Handlers =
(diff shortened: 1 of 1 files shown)


yhirose:
@sebasjm Thank you for the PR, and for the pointer to the systemd documentation.

Unfortunately I have to decline this for the same reason I gave in #2452, where the same feature was requested: passing an externally created socket to `Server` requires the semantics of the fd to match exactly what the server instance expects, and it raises the difficult question of whether the fd's lifetime is transferred to cpp-httplib or stays with the caller. All of those cases would need to be covered by thorough unit tests before I could take it in, and there are none here.

One thing worth noting in case you carry this in your own fork: the systemd rule you quoted applies only to the *listening* socket obtained with `Accept=no`. The document explicitly allows `shutdown(2)` on the connected sockets. This PR applies `shutdown_socket_` to the accepted client sockets as well (in `Server::process_and_close_socket()`, `SSLServer::process_and_close_socket()`, and the resource-exhaustion path in `listen_internal()`), so `set_shutdown_socket(false)` changes the connection teardown behavior of every client connection, not just the inherited listening socket. Only the one in `Server::stop()` is needed for your use case.

Thanks again for your understanding!

sebasjm:
> The document explicitly allows shutdown(2) on the connected sockets

100% i fixed in my branch, thanks for pointing it up.

> All of those cases would need to be covered by thorough unit tests before I could take it in

I agree, if you tell me what kind of test is good enough I can work it out. Otherwise I can keep this for me.

> it raises the difficult question of whether the fd's lifetime is transferred to cpp-httplib or stays with the caller

Stays with the caller. 
You either get a new socket by `bind_internal` in that case is owned by the `Server` or `set_socket / set_shutdown_socket / listen_after_bind` and in that case is __not__ owned by the server.

TBH `set_socket` can also make `shutdown_socket_ = false` and `bind_internal` set `shutdown_socket_ = false`.

```cpp
inline Server &Server::set_socket(const socket_t n_sock) {
  svr_sock_ = n_sock;
  shutdown_socket_ = false;
  return *this;
}

inline bool Server::listen(const socket_t socket) {
  set_socket(socket); 
  return listen_internal();
}
```

PR #2514, opened 2026-08-01
websocket: rebuild handshake on Request (support Host override, align with HTTP client behavior)

Hyukya:
## Motivation

1. Like HTTP, WebSocket needed support for overriding Host and adding custom headers.
2. Keep WebSocket connection behavior as consistent as possible with the existing HTTP client.

## Changes

- `perform_websocket_handshake` now takes a `Request&` instead of building a raw
  string, using `write_request_line`/`check_and_write_headers` (same path as the
  HTTP client).
- Protocol-mandatory headers (Upgrade, Connection, Sec-WebSocket-Key/Version) are
  erased and re-emplaced right before sending, so caller-supplied values can't
  break the handshake.
- Added `WebSocketClient::prepare_default_headers`: injects only Host and
  User-Agent (body-related headers like Accept are skipped since a 101 response
  has no body).
- `connect()` now builds the Request (including headers) before the handshake
  instead of after, enabling Host and other header overrides.

## Compatibility

No public API changes. Existing call sites behave the same as before.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -3154,6 +3154,10 @@ private:
 std::string make_host_and_port_string(const std::string &host, int port,
                                       bool is_ssl);
 
+template <typename T>
+bool check_and_write_headers(Stream &strm, Headers &headers, T header_writer,
+                             Error &error);
+
 std::string trim_copy(const std::string &s);
 
 void divide(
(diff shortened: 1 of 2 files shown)


yhirose:
@Hyukya thanks for the excellent improvement!

Hyukya:
@AhnLab-OSS @AhnLab-OSSG

PR #2515, opened 2026-08-01
Allow addr_map_ values to be hostnames, not just IP literals

Hyukya:
## Motivation
- addr_map_ only supported IP literals as mapped values, used purely to bypass DNS resolution while `host_` stayed the identity for the Host header and SNI.
- Some setups need to redirect a hostname to another hostname (not just an IP), e.g. when the target is only reachable by name.
## Changes
- In `ClientImpl::create_client_socket` and `WebSocketClient::connect`, the addr_map_ lookup now checks whether the mapped value is an IP literal or a hostname, using `detail::is_ip_address`.
- IP literal values are still passed as the `ip` argument (unchanged `AI_NUMERICHOST` path); hostname values are now passed as `connect_host` for resolution.
- `host_` itself is never touched, so it keeps supplying the Host header and SNI in both cases.
- Moved `is_ip_address()` into the `detail` namespace so both client code paths can share it instead of duplicating the check.
- Added 3 tests: hostname-as-value redirection, IP-as-value redirection (regression), and empty-value rejection.
## Compatibility
- No public API changes. Existing call sites (IP-literal mappings) behave the same as before.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -2452,7 +2452,8 @@ protected:
   std::thread::id socket_requests_are_from_thread_ = std::thread::id();
   bool socket_should_be_closed_when_request_is_done_ = false;
 
-  // Hostname-IP map
+  // Hostname to connection target map. The value is an IP literal or another
+  // hostname; only the connection target changes, never the identity.
   std::map<std::string, std::string> addr_map_;
 
   // Default headers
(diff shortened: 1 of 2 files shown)


yhirose:
@Hyukya thank you for the contribution!

Hyukya:
@AhnLab-OSS @AhnLab-OSSG

PR #2516, opened 2026-08-01
Reject ambiguous request framing

carrerasdarren-cell:
## Summary

- Reject server requests that contain both `Content-Length` and
  `Transfer-Encoding`, including `Content-Length: 0`.
- Reject request `Transfer-Encoding` values whose final coding is not
  `chunked`.
- Prevent the client from automatically adding `Content-Length` when a caller
  explicitly supplies `Transfer-Encoding`.

## Why

RFC 9112 section 6.3 requires a server to return 400 and close the connection
when a request has `Transfer-Encoding` without `chunked` as the final coding.
It also identifies messages containing both framing headers as potential
request smuggling attempts.

Current `master` classifies `Transfer-Encoding: chunked, gzip` as not chunked,
but then treats the request as having no body. On a persistent connection,
bytes intended as that indeterminate body can consequently be parsed as a new
request. The existing `Content-Length` conflict check also uses a numeric
`> 0` test, so `Content-Length: 0` bypasses it.

A two-stage raw-socket proof sends the malformed outer request, waits for its
response, and then sends a request-shaped byte sequence. Before this patch,
both requests receive 200 responses and the hidden route executes once. With
the patch, the outer request receives 400, the connection closes, and the
hidden route is never executed.

This builds on #2500's final-coding detection. That change correctly returns
false for a non-final `chunked` token; this patch enforces the required server
error at the caller and keeps valid `gzip, chunked` behavior as a negative
control.

## Validation

- Full OpenSSL test suite: 755 tests passed across four shards.
- Header-only and split-library builds passed.
- Targeted tests passed in both OpenSSL and no-TLS configurations.
- The official server, client, and header-parser fuzz regression corpora
  replayed successfully.
- The two-stage proof passed under Linux AddressSanitizer and
  UndefinedBehaviorSanitizer.
- Changed lines pass clang-format and `git diff --check`.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -12603,11 +12603,13 @@ Server::process_request(Stream &strm, const std::string &remote_addr,
     return write_response(strm, close_connection, req, res);
   }
 
-  // RFC 9112 §6.3: Reject requests with both a non-zero Content-Length and
-  // any Transfer-Encoding to prevent request smuggling. Content-Length: 0 is
-  // tolerated for compatibility with existing clients.
-  if (req.get_header_value_u64("Content-Length") > 0 &&
-      req.has_header("Transfer-Encoding")) {
+  // RFC 9112 §6.3: ambiguous request framing can desynchronize this parser
+  // from an intermediary and enable request smuggling. Reject both framing
+  // headers together, even when Content-Length is zero, and reject a
+  // Transfer-Encoding whose final coding is not chunked.
+  if (req.has_header("Transfer-Encoding") &&
+      (req.has_header("Content-Length") ||
+       !detail::is_chunked_transfer_encoding(req.headers))) {
     connection_closed = true;
     res.status = StatusCode::BadRequest_400;
     return write_response(strm, close_connection, req, res);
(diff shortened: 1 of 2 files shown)


yhirose:
Thanks for the report and the analysis — the underlying issue is real, and I confirmed it: with `Transfer-Encoding: chunked, gzip` on a keep-alive connection, the server answers 200 and the remaining bytes are parsed as a second request. The location and the shape of your fix are correct.

I'm closing this PR because of what is bundled around that fix, not because of the fix itself.

1. Rejecting `Content-Length: 0` alongside `Transfer-Encoding`, and the two client-side changes that follow from it, are a separate hardening. They are defensible under RFC 9112, but they are not needed for the smuggling fix and they change behavior for existing clients, so they need their own PR and a release note.

2. That bundled change is what forced the modifications to existing tests (`CaseInsensitiveTransferEncoding`, `LargeChunkedPost`). Modifying existing tests is something I need called out explicitly in the description with the reasoning; this PR does not mention it at all.

3. Most importantly, it silently disables the regression tests for GHSA-h6wq-j5mv-f3q8. `expect_chunked_body_rejected()` still sends `Content-Length: 0` with `Transfer-Encoding: chunked`, so the new framing check now returns 400 during header processing and the chunk-size parser is never reached. `RejectsNegativeChunkSize` and `RejectsChunkSizeWithLeadingPlus` still pass, but they no longer test anything — I verified that a perfectly valid chunked body with `Content-Length: 0` also returns 400 under this patch. A green test suite is exactly what hid this.

I'll land the non-final-chunked rejection separately. It needs one condition change in `process_request()` and no changes to any existing test.

PR #2519, opened 2026-08-01
Close the bound socket in ~Server() and on bind_internal() error paths

yhirose:
Follow-up to #2517. Two paths still leaked a bound-and-listening socket after stop() learned to release it:

**`~Server()`** was defaulted, so a Server that bound and was then destroyed without stop() kept the descriptor and the port — and since `create_server_socket()` has already called `listen(2)`, the abandoned port kept accepting connections into the backlog and never answering them. The destructor now calls stop(). Destroying a Server whose accept loop is still running was always undefined behavior, so this only affects the bound-but-idle case.

**`bind_internal()`** had the same problem on its own error paths: a `getsockname()` failure or an unsupported address family returned -1 with the socket already bound and listening. Those two branches now close it directly.

`BindServerTest.DestructorClosesBoundSocket` covers the destructor path with the raw-`connect()` helper from #2517 (a `Client` request against the old code would hang instead of fail). The `bind_internal()` error paths are not covered, since `getsockname()` cannot readily be made to fail.

Verified with the full suite (743/743) on both the regular and split builds.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -11327,7 +11327,12 @@ inline Server::Server()
 #endif
 }
 
-inline Server::~Server() = default;
+inline Server::~Server() {
+  // Close the listening socket for callers that bound but never stopped the
+  // server. Destroying a Server whose accept loop is still running was always
+  // undefined behavior, so this only affects the bound-but-idle case.
+  stop();
+}
 
 inline std::unique_ptr<detail::MatcherBase>
 Server::make_matcher(const std::string &pattern) {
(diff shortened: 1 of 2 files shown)


yhirose:
After #2517, stop() releases the socket from any state, so the invariant users need — call stop() and everything is freed — already holds. The destructor change only benefits code that never calls stop() in a long-lived process, and doing it safely requires guarding against a still-running accept loop (closing the socket from the destructor would wake the parked loop on a destroyed object). Not worth the risk surface. The bind_internal() error paths are likewise recoverable via stop() and effectively unreachable in practice.

PR #2529, opened 2026-08-03
match mount points on a segment boundary in handle_file_request

metsw24-max:
**Mount point matched mid-segment in handle_file_request**

The mount point is compared as a plain byte prefix, so a mount of `/mount` also serves `/mountdir/...` out of the same directory; with the pairing the ServerTest fixture already uses (`/` on `./www`, `/mount` on `./www2`), `GET /mountdir/test.html` answers 200 with the contents of `./www2/dir/test.html`. Nothing escapes the base directory since `is_valid_path` still rejects traversal, but the mounted tree becomes reachable under URLs outside the prefix the mount declares, so a reverse proxy location, auth filter or cache rule keyed on `/mount/` quietly stops covering files it was meant to guard.

Matching now has to land on a segment boundary. A mount point that already ends in `/`, the root mount included, keeps its current behaviour, and `set_mount_point` guarantees the value is non-empty and starts with `/`.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -12448,8 +12448,14 @@ inline bool Server::read_content_core(
 
 inline bool Server::handle_file_request(Request &req, Response &res) {
   for (const auto &entry : base_dirs_) {
-    // Prefix match
-    if (!req.path.compare(0, entry.mount_point.size(), entry.mount_point)) {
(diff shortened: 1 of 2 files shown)


yhirose:
@metsw24-max thanks for the excellent contribution!

PR #2535, opened 2026-08-10
Idea to improve latency in Route Matching

fidgetaryan445:
Added another class for route matching to improve latency for simple string literal routes along with their benchmarks and tests

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,6 +20,7 @@
 	* HTTPLIB_INSTALL (default on)
 	* HTTPLIB_SHARED (default off) builds as a shared library (if HTTPLIB_COMPILE is ON)
 	* HTTPLIB_TEST (default off)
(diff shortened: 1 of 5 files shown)


yhirose:
@fidgetaryan445, thanks for the PR and for the benchmarks. The idea is sound, and I confirmed the speedup: matching a literal route through std::regex_match was costing far more than a plain comparison, and it shows up end to end once an application has enough routes.

I ended up implementing it a bit differently in #2538, which is now merged. `PathParamsMatcher` already does an exact literal comparison when it captures no parameter, so the same optimization works by reusing it instead of adding a `LiteralMatcher` class. The metacharacter check is the same as yours, so exactly the same routes take the fast path, and the throughput is the same in my measurements.

The reason I went that way is scope: it comes to 5 lines of code in httplib.h with no new class, and it does not add the CMake subproject that fetches Google Benchmark over the network, which does not fit how benchmark/ works today.

Closing this one, but thanks for spotting it. The improvement is in master now.

PR #2536, opened 2026-08-11
reject trailing bytes after IPv6 host literal in parse_url

metsw24-max:
**Malformed IPv6 authority accepted in parse_url**

After the closing bracket of an IPv6 literal the parser never checks that the next byte is a port, path, query or fragment delimiter, so an authority like `[::1]evil.com` is accepted with host `::1` and the trailing `evil.com` folded silently into the path. Following an untrusted `Location`, this connects to the bracketed address while the URL reads as another host. I made `]` require a delimiter (or end of input), which mirrors the existing rejection of an unclosed `[`, and added regression cases to `ParseUrlTest`.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -839,6 +839,15 @@ inline bool parse_url(const std::string &url, UrlComponents &uc) {
       }
 
       pos = close + 1;
+
+      // The IPv6 literal is the whole host, so ']' must be followed by a port,
+      // path, query or fragment delimiter (or the end of input). Otherwise the
+      // trailing bytes would be folded into the path while the connection
+      // still targets the bracketed address.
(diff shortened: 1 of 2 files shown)


yhirose:
@metsw24-max thanks for the PR! Just one suggestion. `url[pos]` is indexed four times here. So storing it in a local variable would make the delimiter check a bit easier to read.

```cpp
if (pos < url.size()) {
  auto c = url[pos];
  if (c != ':' && c != '/' && c != '?' && c != '#') { return false; }
}
```

P.S. I handled it at acd0640870d6fd26b0977b0d1b0f06dc838c2839

metsw24-max:
Agreed, the local variable reads better than indexing url[pos] four times. Thanks for tidying that up and merging.

PR #2537, opened 2026-08-13
Always use meson option non_blocking_getaddrinfo

jfsimoneau:


yhirose:
@Tachi107 could you please review this PR which modifies `meson.build`? Thanks!

meson.build
```diff
@@ -107,7 +107,7 @@ if host_machine.system() == 'windows'
 elif host_machine.system() == 'darwin'
   async_ns_dep = dependency('appleframeworks', modules: ['CFNetwork', 'CoreFoundation'], required: async_ns_opt)
 else
-  has_gai_a = cxx.has_function('getaddrinfo_a', args: '-D_GNU_SOURCE')
+  has_gai_a = not async_ns_opt.disabled() and cxx.has_function('getaddrinfo_a', args: '-D_GNU_SOURCE')
```
Tachi107: What are you trying to achieve here?

jfsimoneau: When a user specifies `-Dnon_blocking_getaddrinfo=disabled`, we should not check if `getaddrinfo_a()` is included in glibc because we want to fall back to `getaddrinfo()`. On the other hand, if the user specifies `-Dnon_blocking_getaddrinfo=enabled`, we only want to fail if `getaddrinfo_a()` is neither in glibc nor libanl.
At first, I wanted add `required: async_ns_opt` to `cxx.has_function()`, but that would fail if the user specifies `enabled` without checking for libanl and would also require meson 1.30.

jfsimoneau: Would you prefer I change the test at line 118? With the following example, meson would still find the `getaddrinfo_a()` function built in glibc, but it would skip the definition of `CPPHTTPLIB_USE_NON_BLOCKING_GETADDRINFO` when the feature is disabled. Not finding the `getaddrinfo_a()` function would fall back to the blocking `getaddrinfo()` too if the feature is set to `auto` and the failure would still happen when failing to find libanl if the feature is required.
```
if async_ns_opt.require(async_ns_dep.found(), "Unable to find getaddrinfo_a").allowed()
  deps += async_ns_dep
  args += '-DCPPHTTPLIB_USE_NON_BLOCKING_GETADDRINFO'
endif
```

Tachi107: Oh okay, now it makes sense, thanks for the explanation :)

Two comments then:

1. Instead of `not .disabled()`, maybe `.allowed()` reads better
2. I find the `.require` alternative a bit harder to follow

So, if you agree to change `not .disabled()` to `.allowed()`, than I'm okay with this patch!

shameershamus410-cmd (approved):

yhirose:
@jfsimoneau @Tachi107 thanks!

PR #2540, opened 2026-08-17
Make decode_uri the inverse of encode_uri

dngr2:
`decode_uri` is documented as the counterpart of `encode_uri` (the JS `encodeURI`/`decodeURI` pair), but its body is byte-for-byte identical to `decode_uri_component` — it decodes every `%XX`, including escapes of the reserved characters `encode_uri` deliberately leaves literal (`;/?:@&=+$,#`).

So `decode_uri` is not the inverse of `encode_uri`, and it promotes an escaped delimiter into a real one:

```cpp
decode_uri("http://h/a%2Fb");   // returned "http://h/a/b" — invents a path separator
decode_uri("%23");              // returned "#"
```

JS `decodeURI` leaves exactly this reserved set encoded for that reason. This keeps their escapes and decodes everything else, so `decode_uri(encode_uri(s)) == s` round-trips. `decode_uri_component` is unchanged (it still decodes reserved characters). Existing `DecodeUriTest` cases still pass — none of them exercised a reserved-character escape — and a test for the reserved-escape behavior is added.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -10603,7 +10603,19 @@ inline std::string decode_uri(const std::string &value) {
     if (value[i] == '%' && i + 2 < value.size()) {
       auto val = 0;
       if (detail::from_hex_to_i(value, i + 1, 2, val)) {
-        result += static_cast<char>(val);
+        auto c = static_cast<char>(val);
+        // Keep escapes of the reserved characters that encode_uri leaves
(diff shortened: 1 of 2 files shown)


yhirose:
@dngr2 thanks for your fine contribution!

PR #2541, opened 2026-08-18
Reduce syscall volume, and/or TCP segment generation

gsmecher:
Normally, httplib generates one TCP segment for headers and at least one more for body content. This is awful without TCP_NODELAY ("It's always TCP_NODELAY. Every damn time." [[1]](https://brooker.co.za/blog/2024/05/09/nagle.html)). Even with TCP_NODELAY, small responses can become unnecessarily segmented and throughput can suffer on deeply embedded hosts due to syscall overhead.

This patch series threads a vectorized write syscall (writev() on Linux/MacOS, something else under Win32) through httplib just enough to combine headers and body content into a single, zero-copy syscall under "happy path" conditions (chiefly, no ranged requests).

- New Response::set_content() allows callers to provide a copy-free "borrowed view" into C arrays. The existing std::move-based set_content() call isn't useful for non-std::string arguments, and the existing Response::set_content(char*, size_t) call required copying. This new call requires some pointer-lifetime effort but allows copy-free responses to come from non-strings (e.g. mmap'd files); both static-mount and set_file_content() responses now use it.

- Server::write_response_core uses writev() and "borrowed view" content to entirely combine most Responses into a single syscall. This avoids excess TCP segment generation (a good thing with or without NODELAY), and on resource-constrained systems, the avoided syscall overhead itself can be meaningful.

- A new Response::clear_content() drops all pending content representations at the error-reset sites and uncaught handler exceptions, and every content setter clears a previously set borrowed view, so stale borrowed bytes can never trail a response whose headers no longer describe them.

The abidiff test fails (because the ABI did change, and I haven't modified anything to match). I'm uncertain how you handle ABI changes but happy to follow your lead.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -172,6 +172,10 @@
 #define CPPHTTPLIB_RECV_BUFSIZ size_t(16384u)
 #endif
 
+#ifndef CPPHTTPLIB_WRITEV_MAX_BUFS
+#define CPPHTTPLIB_WRITEV_MAX_BUFS size_t(8u)
+#endif
+
 #ifndef CPPHTTPLIB_SEND_BUFSIZ
 #define CPPHTTPLIB_SEND_BUFSIZ size_t(16384u)
 #endif
@@ -314,6 +318,7 @@ using socklen_t = int;
 #include <pthread.h>
 #include <sys/mman.h>
 #include <sys/socket.h>
+#include <sys/uio.h>
 #include <sys/un.h>
 #include <unistd.h>
 
@@ -1749,6 +1754,31 @@ namespace detail {
 // `Response` stores one.
 enum class EncodingType { None = 0, Gzip, Brotli, Zstd };
 
+class ContentResource {
+public:
+  ContentResource() = default;
+  explicit ContentResource(ContentProviderResourceReleaser releaser)
+      : state_(releaser ? std::make_shared<State>(std::move(releaser))
+                        : nullptr) {}
+
+  void set_success(bool success) {
+    if (state_) { state_->success = success; }
+  }
+
+private:
+  struct State {
+    explicit State(ContentProviderResourceReleaser r)
+        : releaser(std::move(r)) {}
+    ~State() { releaser(success); }
+    State(const State &) = delete;
+    State &operator=(const State &) = delete;
+
+    ContentProviderResourceReleaser releaser;
+    bool success = false;
+  };
+  std::shared_ptr<State> state_;
+};
+
 } // namespace detail
 
 struct Response {
@@ -1781,6 +1811,11 @@ struct Response {
   void set_content(const std::string &s, const std::string &content_type);
   void set_content(std::string &&s, const std::string &content_type);
 
+  // Zero-copy path: s is not copied, and must remain until the response
+  // completes; resource_releaser is invoked exactly once.
+  void set_content(const char *s, size_t n, const std::string &content_type,
+                   ContentProviderResourceReleaser resource_releaser);
+
   void set_content_provider(
       size_t length, const std::string &content_type, ContentProvider provider,
       ContentProviderResourceReleaser resource_releaser = nullptr);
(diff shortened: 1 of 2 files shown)


yhirose:
Thanks for the PR! The benchmark could not see this change at all: its only endpoint returns a 12-byte `set_content()` body, the one case where the response line, the headers and the body already share a single `write()`. So I added static-file, large-body and TLS workloads to it first (on master now, so please rebase to pick them up). Three things came out.

**1. Large TLS responses lose a third to a half.**

```
# after rebasing
./benchmark/ab.sh --base master --head <this branch> --path /static/large.bin --tls
```

| workload | 1 MiB | 10 MiB |
|---|---|---|
| `/large` plain | 1.129x | 1.180x |
| `/static/large.bin` plain | 1.038x (ns) | 0.949x (ns) |
| `/large` TLS | **0.576x** | **0.640x** |
| `/static/large.bin` TLS | **0.563x** | **0.534x** |

Five alternating rounds each; every bold row separates at p = 0.008, and `/static/small.js` gains 1.260x plain and 1.358x TLS, so the harness is not simply reporting noise.

`SSLSocketStream` does not override `writev()`, so every TLS response takes the `Stream::writev()` fallback, which re-fragments the body into `CPPHTTPLIB_SEND_BUFSIZ` pieces: 64 `SSL_write()` calls for 1 MiB and 640 for 10 MiB, where master issued one. The comment reasons that "large ones chunk at the same granularity TLS fragments records anyway", but a 16 KiB maximum record does not make 16 KiB write calls free. master handed the whole buffer to OpenSSL and let it emit the records internally.

Handing a span larger than the staging buffer straight to `write()`, instead of copying it through, recovered this for me: 1 MiB dynamic went to 1709 req/s against master's 1589. Worth noting the fallback is what every non-socket `Stream` uses, user-defined ones included, so this is not only about TLS.

**2. A response can promise a body and send none.**

```cpp
TEST(BorrowedContentTest, ReplacingViewWithEmptyBodyClearsContentLength) {
  Server svr;

  const std::string data = "0123456789abcdefghij"; // 20 bytes
  svr.Get("/replaced", [&](const Request & /*req*/, Response &res) {
    res.set_content(data.data(), data.size(), "text/plain", nullptr);
    res.set_content("", "text/plain"); // drop it again
  });

  auto listen_thread = std::thread([&svr]() { svr.listen("localhost", PORT); });
  auto se = detail::scope_exit([&] {
    svr.stop();
    listen_thread.join();
  });
  svr.wait_until_ready();

  Client cli("localhost", PORT);
  cli.set_read_timeout(1, 0);

  auto res = cli.Get("/replaced");
  ASSERT_TRUE(res) << "Error: " << to_string(res.error());
  EXPECT_EQ("0", res->get_header_value("Content-Length"));
  EXPECT_EQ("", res->body);
}
```

On the branch this reports `Error: Failed to read connection`. `content_length_` doubles as the view's length, and the copying `set_content()` overloads clear `content_view_data_` without clearing it, so `apply_ranges()` believes the stale length while no content source is left to write the bytes. The response goes out with `Content-Length: 20` and no body on an open keep-alive connection, and the client reads the next response as this one's body. master is correct here, and this is reachable without ever calling the new API since the PR moves `handle_file_request()` onto it.

```cpp
inline void Response::set_content(const char *s, size_t n,
                                  const std::string &content_type) {
  body.assign(s, n);
  content_view_data_ = nullptr;
  content_length_ = 0;                   // add
  content_provider_ = nullptr;           // add
  is_chunked_content_provider_ = false;  // add
```

plus the same three lines in the `std::string &&` overload.

**3. The releaser is not always called.**

```cpp
TEST(BorrowedContentTest, ReplacingContentReleasesTheOldView) {
  Server svr;

  const std::string a = "AAAA";
  const std::string b = "BBBB";
  std::atomic<int> released_a{0};

  svr.Get("/twice", [&](const Request & /*req*/, Response &res) {
    res.set_content(a.data(), a.size(), "text/plain",
                    [&](bool) { released_a++; });
    res.set_content(b.data(), b.size(), "text/plain", nullptr);
  });

  auto listen_thread = std::thread([&svr]() { svr.listen("localhost", PORT); });
  auto se = detail::scope_exit([&] {
    svr.stop();
    listen_thread.join();
  });
  svr.wait_until_ready();

  Client cli("localhost", PORT);
  auto res = cli.Get("/twice");
  ASSERT_TRUE(res) << "Error: " << to_string(res.error());
  EXPECT_EQ("BBBB", res->body);
  EXPECT_EQ(1, released_a.load()); // the discarded view is still released
}
```

Setting content a second time overwrites the previous releaser instead of invoking it, so `released_a` stays 0. The library's own `[mm](bool) {}` releasers escape this only by luck: they hold the mapping in a captured `shared_ptr`, so destroying the discarded `std::function` unmaps it anyway. One that does its work in the body, like `[p](bool) { free(p); }`, has no such fallback.

```cpp
  void clear_content() {
    if (content_provider_resource_releaser_) {       // add
      content_provider_resource_releaser_(false);    // add
      content_provider_resource_releaser_ = nullptr; // add
    }
    body.clear();
```

Nothing to do about abidiff, by the way. It failing just makes the next release a minor bump instead of a patch.

gsmecher:
Wow - thanks, that's a more detailed review than I expected, and benchmarking feels like you've done some of my homework for me. Thank you. I will rebase, take a good look, and report back. 

Overall: do you think the juice will be worth the squeeze? I'm primarily optimizing for small embedded hosts, and if this PR feels like a distraction from your primary use case, please say so. 

Thanks, 
Graeme

gsmecher:
@yhirose - I have rebased to catch your changes to master, and pushed out a squashable commit to address your comments. If you prefer, I can squash this commit myself (leaving just the two logical commits you would want to merge.)

I've also reproduced your benchmarks and think the regressions are addressed now.

yuvarajayuvarajay276-wq:
```c++

```

yuvarajayuvarajay276-wq:
@yuvarajayuvarajay276-wq

yuvarajayuvarajay276-wq:
YUVARAJA

yuvarajayuvarajay276-wq:
[dockerignore.txt](https://github.com/user-attachments/files/31641092/dockerignore.txt)

yuvarajayuvarajay276-wq:
###

yuvarajayuvarajay276-wq:
### 

- [ ] [clang-format.txt](https://github.com/user-attachments/files/31641131/clang-format.txt)

gsmecher:
(rebased off master and simplified to match recent changes.)

PR #2544, opened 2026-08-18
validate bearer scheme in get_bearer_token_auth

metsw24-max:
**Unvalidated scheme in get_bearer_token_auth**

The helper strips a fixed `Bearer ` prefix with `substr(7)` but never checks the header value is that long or actually uses that scheme. An `Authorization` value shorter than seven bytes (for example `x`, or a bare `Bearer`) makes `substr` throw `std::out_of_range`, and a value from a different scheme like `Basic ...` comes back as a bearer token. It now matches the `Bearer ` scheme case-insensitively (RFC 7235) before returning the remainder, and returns an empty string otherwise. A well-formed `Bearer <token>` is unaffected.

Added a small regression test covering the short value, the wrong scheme, and the valid case.


yhirose:
@metsw24-max thanks!

Issue #847, opened 2021-01-27
Cusom HTTP methods


aldoshkind:
Hi!
I would like to use HTTP methods other than currently supported by cpp-httplib. Particularly SUBSCRIBE is demanded.
Is there any opportunity to create some interface allowing to add handlers for custom HTTP methods?

Something like:
```
Server svr;
svr.method("SUBSCRIBE", "/hi", [](const Request &, Response &) {
    ...
});
```

yhirose:
@aldoshkind, thanks for the feedback. I am not planning to add the routing handler support for custom HTTP methods at this point. But, you can do the same with [pre routing handle](https://github.com/yhirose/cpp-httplib#pre-routing-handler). Could you take a look at it to see if it satisfies your need?

yhirose:
@aldoshkind, it seems like the following code needs to modified to allow such custom methods...

https://github.com/yhirose/cpp-httplib/blob/78c474c744a2582ad18cb6d936ab9273fffc6350/httplib.h#L4383-L4386

aldoshkind:
@yhirose Thanks for rapid response! I'll check as soon as possible

PR #2552, opened 2026-08-25
Allow extending the set of accepted HTTP methods

GribanovIvan:
## Why the existing extension points don't help

In #847 you suggested `set_pre_routing_handler`, and that is the right shape for this — but it cannot be reached. Validation runs strictly earlier. In `process_request` the flow is:

1. `parse_request_line(...)` — rejects here, returns `400`;
2. …later `routing()` — and only *that* calls `pre_routing_handler_`.

So for an unknown method the handler is never invoked. In the same thread you then pointed at these exact lines as the ones that would need modifying, which is what this PR does.

Subclassing does not help either: `parse_request_line` is private and non-virtual.

## Use case

Serving WebDAV. `PROPFIND`, `PROPPATCH`, `MKCOL`, `COPY`, `MOVE`, `LOCK` and `UNLOCK` are ordinary HTTP methods defined by RFC 4918. Everything else a DAV endpoint needs — the `207 Multi-Status` XML, the `DAV:` headers — is perfectly implementable on cpp-httplib today; the method check is the single blocker.

We currently ship a one-line patch for exactly this, applied at build time and guarded so an upstream refresh cannot silently drop it. It works, but a supported knob would be better for everyone.

## The change

A configuration macro, matching the existing `CPPHTTPLIB_*` macros:

```cpp
#ifndef CPPHTTPLIB_ADDITIONAL_METHODS
#define CPPHTTPLIB_ADDITIONAL_METHODS
#endif

thread_local const std::set<std::string> methods{
    "GET",     "HEAD",    "POST",  "PUT",   "DELETE",
    "CONNECT", "OPTIONS", "TRACE", "PATCH", "PRI",
    CPPHTTPLIB_ADDITIONAL_METHODS};
```

Used as:

```cpp
#define CPPHTTPLIB_ADDITIONAL_METHODS "PROPFIND","PROPPATCH","MKCOL"
```

Empty by default, so behaviour is unchanged unless it is set. No new state, no runtime cost, header-only friendly. 14 lines added, 1 changed.

Note on the trailing comma: it is kept in the header rather than expected from the macro value. A trailing comma in a braced-init-list is well-formed, so the empty default compiles cleanly. The alternative — expecting the value to start with a comma — has a bad failure mode: omitting it makes the adjacent string literals concatenate, turning `"PRI" "PROPFIND"` into the single token `"PRIPROPFIND"`, which silently disables both methods and still compiles. With the comma in the header, the same mistake is a compile error.

## Verification

Same program, same handler — a server whose `set_pre_routing_handler` answers `PROPFIND` with `207`. Only the macro differs:

```
built without the macro:   GET /ping -> 200,   PROPFIND /dav/ -> 400
built with the macro:      GET /ping -> 200,   PROPFIND /dav/ -> 207
```

Without the macro that handler is never reached, which is the point above. An unpatched build still rejects `PROPFIND` with `400`, so existing behaviour is preserved.

Compile-checked in six configurations — default, `CPPHTTPLIB_OPENSSL_SUPPORT`, `CPPHTTPLIB_ZLIB_SUPPORT`, each with and without the macro — all clean.

`test/test.cc` contains no reference to `InvalidHTTPMethod`, so no existing test depends on the rejection behaviour.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -22,6 +22,18 @@
  * Configuration
  */
 
+/*
+ * Additional HTTP methods the server will accept, on top of the standard set.
+ * Define as a comma-separated list of string literals, e.g. for WebDAV:
+ *
+ *   #define CPPHTTPLIB_ADDITIONAL_METHODS "PROPFIND","PROPPATCH","MKCOL"
+ *
+ * Empty by default, so behaviour is unchanged unless it is set.
+ */
+#ifndef CPPHTTPLIB_ADDITIONAL_METHODS
+#define CPPHTTPLIB_ADDITIONAL_METHODS
+#endif
+
 #ifndef CPPHTTPLIB_KEEPALIVE_TIMEOUT_SECOND
 #define CPPHTTPLIB_KEEPALIVE_TIMEOUT_SECOND 5
 #endif
@@ -12735,7 +12747,8 @@ inline bool Server::parse_request_line(const char *s, Request &req) const {
 
   thread_local const std::set<std::string> methods{
       "GET",     "HEAD",    "POST",  "PUT",   "DELETE",
-      "CONNECT", "OPTIONS", "TRACE", "PATCH", "PRI"};
+      "CONNECT", "OPTIONS", "TRACE", "PATCH", "PRI",
+      CPPHTTPLIB_ADDITIONAL_METHODS};
 
   if (methods.find(req.method) == methods.end()) {
     output_error_log(Error::InvalidHTTPMethod, &req);


yhirose:
@GribanovIvan, thanks for the writeup, but I can't merge this.

Fist of all, it doesn't solve the problem you describe. Getting past the method check isn't enough for WebDAV, because a custom method still has no way to read its request body, so `PROPPATCH` and `LOCK` stay out of reach.

Beyond that, this isn't at a quality bar I can merge. There are no tests, and the macro silently does nothing in the split build (`httplib.h` + `httplib.cc`), which is how a lot of people consume this library.

Custom methods are a real gap and I do want to solve it, but it needs a design that works end to end and comes with tests. Anyway, thanks for your feedback.

GribanovIvan:
Huge thank you for implementing it the proper way

PR #2554, opened 2026-08-26
Fix accept() error handling on Windows

DTBob:
The accept loop in Server::listen_internal() inspects errno, but Winsock never sets the CRT errno - it reports failures through WSAGetLastError(). Both retry branches were therefore dead code on Windows, and every accept() failure fell through to the fatal path, which closes the listening socket and ends listen().

That is reachable in normal operation. WSAECONNRESET is documented for accept() when a peer resets a pending connection before it is accepted (a port scanner is enough to trigger it), and WSAEMFILE/WSAENOBUFS show up under load. One such event permanently stops the server accepting.

Add is_accept_resource_error() and is_accept_transient_error() beside the existing is_connection_error(), which already abstracts the same errno vs WSAGetLastError() difference, and use them in the accept loop.

POSIX behaviour is unchanged: EMFILE still sleeps and retries, EINTR and EAGAIN still retry immediately.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -6925,6 +6925,30 @@ inline bool is_connection_error() {
 #endif
 }
 
+// accept() failed because the process or the network stack is temporarily out
+// of resources. The listening socket is still good; retry after a short pause.
+inline bool is_accept_resource_error() {
+#ifdef _WIN32
+  auto err = WSAGetLastError();
+  return err == WSAEMFILE || err == WSAENOBUFS;
+#else
+  return errno == EMFILE;
+#endif
+}
+
+// accept() failed for a reason that says nothing about the listening socket:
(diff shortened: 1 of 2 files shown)


yhirose:
Superseded by #2561.

The report is right: Winsock never sets the CRT errno, so both retry branches were dead code on Windows. Thanks for finding it.

The patch itself does not meet the bar to merge. It breaks the split build, because `test.cc` calls `detail::is_accept_resource_error()` and `detail::is_accept_transient_error()`, which are defined below the border and so exist only in `httplib.cc`. CI is 15 failing jobs on that. The POSIX side is also left as it was, so `ECONNABORTED`, the POSIX spelling of the aborted pending connection the description is about, stays on the fatal path.

#2561 fixes the same bug with both addressed. Closing this one.

PR #2555, opened 2026-08-26
Clear svr_sock_ when the accept loop closes it after a fatal error

DTBob:
On the fatal path the accept loop closes the listening socket but leaves the descriptor in svr_sock_. Two consequences:

- A later stop() reads that stale value and calls shutdown()/close() on it. By then the OS may have reused the descriptor for an unrelated socket - a worker's keep-alive connection, or a socket opened by the application - and that connection is torn down instead.
- keep_alive() in the worker threads tests svr_sock_ to notice shutdown, so the workers keep waiting on a listening socket that no longer exists.

Take the descriptor with exchange(INVALID_SOCKET) before closing it, which is what stop() already does. That also settles the race with a concurrent stop(): whichever side takes the descriptor closes it exactly once, and the other sees INVALID_SOCKET and does nothing.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -13408,8 +13408,15 @@ inline bool Server::listen_internal() {
         } else if (errno == EINTR || errno == EAGAIN) {
           continue;
         }
-        if (svr_sock_ != INVALID_SOCKET) {
-          detail::close_socket(svr_sock_);
+        // Take svr_sock_ as we close it. Leaving the descriptor behind lets a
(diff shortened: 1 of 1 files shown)


yhirose:
Superseded by #2560.

This one is correct, and CI is clean apart from the `abi` job, which fails on master too. #2560 is the same change with a shorter comment. I took it in along with the rest of this series so that the five land together.

Closing this one. Thanks for the report.

PR #2556, opened 2026-08-26
Give DataSink's optional callbacks safe defaults

DTBob:
`DataSink` has four callbacks, but only `write` is assigned by every writer that hands a sink to a content provider:

  write_content_with_progress()   write, is_writable
  write_content_without_length()  write, is_writable, done
  write_content_chunked()         all four
  send_with_content_provider..()  write
  get_multipart_content_provider() write, done   (cur_sink)

A provider that calls one of the unassigned ones invokes an empty std::function, which throws std::bad_function_call. Nothing on that path catches it, so it unwinds out of the thread running the provider and terminates the process. The README's own idiom (`sink.is_writable()`, `sink.done()`) is enough to hit it: `sink.done()` is documented for the without-length overload, but a provider registered through set_content_provider() with a length gets a sink where `done` is empty.

Default the three optional callbacks instead. A sink is writable unless a writer says otherwise, and a sink that cannot carry trailers still has to finish, so done_with_trailer() falls back to done(). Capturing `this` for that is safe because DataSink is neither copyable nor movable.

A no-op `done` alone would only trade the crash for a hang on the two length-framed paths: they loop until `offset` reaches the promised length, so a provider that reports itself done without writing would be called again immediately, forever. Both now record that the provider finished and stop; the body is short of the Content-Length it promised, so the write fails rather than emitting a truncated message.

Adds DataSinkTest covering the defaults, the done_with_trailer fallback, and both the early-done and write-to-completion paths. The first two fail with "bad function call" without this change.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -1429,9 +1429,20 @@ public:
   DataSink &operator=(DataSink &&) = delete;
 
   std::function<bool(const char *data, size_t data_len)> write;
-  std::function<bool()> is_writable;
-  std::function<void()> done;
-  std::function<void(const Headers &trailer)> done_with_trailer;
+
+  // Every writer that hands a DataSink to a content provider assigns `write`,
+  // but only write_content_chunked() assigns all of the rest, so a provider
+  // that calls one of the others hits an empty std::function and throws
+  // std::bad_function_call. Nothing on the path catches it, so it reaches the
+  // thread that runs the provider and terminates the process. Default them to
+  // something harmless instead: a sink is writable unless a writer says
+  // otherwise, and a sink that cannot carry trailers still has to finish, so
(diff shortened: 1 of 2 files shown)


yhirose:
Superseded by #2562.

The report is right, and the process really does terminate on `sink.done()` from a length-framed provider. Thanks for finding it.

The patch itself does not meet the bar to merge. It breaks the split build, because `test.cc` calls `detail::write_content_with_progress()`, which exists only in `httplib.cc`. CI is 14 failing jobs on that.

#2562 fixes the same bug with that addressed, keeps `Error::Compression` for a failing compressor, and forwards `is_writable` to `cur_sink` in `get_multipart_content_provider()`. Closing this one.

PR #2559, opened 2026-08-26
Do not let a zero-length write end a chunked body

DTBob:
write_content_chunked()'s sink treated "the provider wrote nothing" as "the provider has finished":

    data_available = l > 0;

so sink.write(p, 0) ended the loop. Only done()/done_with_trailer() emit the terminating 0-length chunk, so the body was left unterminated - and the function still returned Success, because the post-loop check only reports the is_shutting_down() case. The peer waits for a last chunk that never arrives until it times out, and on a keep-alive connection anything written next is parsed as a chunk-size line.

A provider reaching a pass with nothing to hand over is ordinary: popping an empty buffer off a queue, or a compressor that has consumed its input without producing output yet. It is not the end of the message.

Ignore zero-length writes instead. A zero-length chunk is the terminator in chunked coding, so it must never be emitted mid-body either way, and data_available is now controlled only by done()/done_with_trailer(). This matches write_content_without_length(), where the sink's write never ends the body.

Adds WriteContentChunkedTest covering the zero-length write, done() with no data at all, and trailers after the terminating chunk. The first fails without this change: the stream comes out empty while the call still reports success.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -8405,8 +8405,15 @@ write_content_chunked(Stream &strm, const ContentProvider &content_provider,
   DataSink data_sink;
 
   data_sink.write = [&](const char *d, size_t l) -> bool {
+    // Writing nothing is not a way to end the body. A zero-length chunk is
+    // the terminator in chunked coding, so it must never be emitted in the
+    // middle of one, and only done()/done_with_trailer() may finish the
+    // message. Treating it as "no more data" (data_available = l > 0) ended
+    // the loop with the terminating chunk still unwritten, and this function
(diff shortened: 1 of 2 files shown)


yhirose:
Superseded by #2563.

The report is right: `sink.write(p, 0)` left the body unterminated while the call still returned Success. Thanks for finding it.

The patch itself does not meet the bar to merge. It breaks the split build, because `test.cc` calls `detail::write_content_chunked()`, which exists only in `httplib.cc`. CI is 14 failing jobs on that.

#2563 fixes the same bug with that addressed. Closing this one.

PR #2566, opened 2026-08-26
Fail make_file_body()'s provider when the file is short

DTBob:
make_file_body() measures the file once and returns that length, which on a server is already the response's Content-Length before a byte is read. The provider re-opens the file by path on each call, so if the file has been rewritten or truncated since, the read comes up empty:

    if (n == 0) { break; }
    ...
    return true;

Reporting success without writing leaves the caller's offset where it was, so write_content_with_progress() calls the provider again on the next pass, gets nothing again, and keeps going - a thread spinning on open/seek/read/close until the peer gives up. The file is opened shared, so nothing prevents a writer truncating it mid-transfer.

Return false instead, which is what every other failure in this provider already does: a bad path, a failed seek, a rejected write.

Adds two cases to MakeFileBodyTest, one truncating the file after make_file_body() has measured it and one sending a whole multi-read file unchanged. They drive the returned ContentProvider directly with a DataSink rather than going through write_content_with_progress(), which lives below the split border and so is not available to test.cc in the split build.

cd test && make: 867 tests, 863 passed, 4 skipped, 0 failed (baseline on master is 865/861/4/0). Reverting httplib.h with the tests in place fails the truncation case with "Actual: true".

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -1523,7 +1523,14 @@ make_file_body(const std::string &filepath) {
       auto to_read = (std::min)(sizeof(buf), length);
       f.read(buf, static_cast<std::streamsize>(to_read));
       auto n = static_cast<size_t>(f.gcount());
-      if (n == 0) { break; }
+      // Out of file with bytes still owed. make_file_body() measured the file
+      // once and that length is already the response's Content-Length, so the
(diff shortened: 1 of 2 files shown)


yhirose:
@DTBob thanks!

PR #2569, opened 2026-08-26
Parse quoted '=' and ';' in Content-Disposition parameters

### LINKED ISSUE #2568: parse_disposition_params: quoted = and ; corrupt multipart name/filename values (last-= wins, semicolon truncates)
## Description

`parse_disposition_params` mishandles both RFC 9110 §5.6.6 allowed characters inside quoted-string `Content-Disposition` values:

```cpp
split(b, e, ';', [&](const char *b, const char *e) {          // no quoted-";" awareness
  ...
  split(b, e, '=', [&](const char *b2, const char *e2) {      // unlimited '=' splits
    if (key.empty()) {
      key.assign(b2, e2);
    } else {
      val.assign(b2, e2);   // every later segment OVERWRITES val → text after LAST '=' wins
    }
  });
  params.emplace(trim_double_quotes_copy(key), trim_double_quotes_copy(val));
```

Two independent defects:

1. **`=` inside a quoted value**: inner split fires per `=`; after `key` is filled, each subsequent segment overwrites `val`, so the value becomes only the text after the last `=`.
2. **`;` inside a quoted value**: outer split truncates the pair at the semicolon; the remainder becomes a new bogus parameter fragment (stray quotes preserved because `trim_double_quotes_copy` strips only matched end quotes).

The library's own client-side sender (`escape_multipart_field`) escapes only `"`, CR, LF — so legitimate filenames like `report=v2.pdf` or `a;b.txt` reach this parser raw and come out corrupted.

**Empirically reproduced** (g++ 15.2.0, harness linking pristine upstream `httplib.h` v0.53.1 / master as of 2026-08-26):

```
name                 => [file1]        ok
filename             => [v2.pdf"]      BUG  expected: report=v2.pdf     (filename="report=v2.pdf")
name(x=y)            => [y"]           BUG  expected: x=y               (name="x=y")
filename("a;b.txt")  => ["a"           BUG  expected: a;b.txt           (filename="a;b.txt")
```

## Steps to Reproduce

Send/upload a multipart part whose header is:

```
Content-Disposition: form-data; name="file1"; filename="report=v2.pdf"
```

Server-side `req.files` / `file.filename` becomes `v2.pdf"` (and `has_file("report=v2.pdf")` fails).

## Expected Behavior

Per RFC 9110 §5.6.6, `=` and `;` are legal inside quoted-strings; the parser must respect quoting for both delimiters (scan-aware splitting or a small state machine), yielding `name=file1`, `filename=report=v2.pdf`.

## Actual Behavior

Value overwritten by last `=` segment; quoted `;` truncates pairs. Deterministic corruption of `file.name` / `file.filename`.

## Impact

Server-side breakage for real-world browser/client uploads containing these characters in filenames: wrong saved-file names, broken extension detection, failed `has_file()` lookups, round-trip infidelity with httplib's own client serializer. Correctness only — no memory-safety issue.

## Suggested Direction

Replace the naive double-split with a quote-aware scanner (track in-quotes state; on `;` commit pair, on first unquoted `=` transition key→value, appending subsequent raw segments—including further `=`—to value). Existing helper structure (`split`/`trim_double_quotes_copy`) stays; only the disposition-param path changes.

## Evidence

Repro output above from untouched upstream header. Dedup: `parse_disposition_params` search → 0 results; broader `content-disposition` (19) / `multipart filename` (27) sweeps show only sending-side escaping work and unrelated threads.

22elix3r:
Fixes #2568.

`parse_disposition_params` used `split()` on `;` and `=` with no quoted-string awareness. A filename like `report=v2.pdf` came out as `v2.pdf"`, and `a;b.txt` got cut at the semicolon.

This walks the parameter list while tracking quotes and quoted-pairs, so `=` and `;` inside the quoted value stay part of the value. Cases that already worked, like `filename = "a.html"`, are unchanged.

Tests:
- unit cases for quoted `=`, quoted `;`, quoted-pair, `filename*`, and spacing around `=`
- a client/server round-trip with `name="x=y"`, `filename="report=v2.pdf"`, and `filename="a;b.txt"`

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -8745,28 +8745,67 @@ inline bool parse_multipart_boundary(const std::string &content_type,
   return !boundary.empty() && boundary.size() <= 70;
 }
 
+// Offset of the first unquoted `d` in [b, e), or (e - b) if none. RFC 9110
+// quoted-string: ';' and '=' inside quotes are not delimiters. A quoted-pair
+// ('\\' + CHAR) does not end the quoted-string or start a delimiter.
+inline size_t find_unquoted(const char *b, const char *e, char d) {
+  auto in_quotes = false;
+  auto escaped = false;
+  for (auto p = b; p != e; ++p) {
+    const auto c = *p;
+    if (in_quotes) {
+      if (escaped) {
+        escaped = false;
+      } else if (c == '\\') {
+        escaped = true;
+      } else if (c == '"') {
+        in_quotes = false;
+      }
+    } else if (c == '"') {
+      in_quotes = true;
+    } else if (c == d) {
+      return static_cast<size_t>(p - b);
+    }
+  }
+  return static_cast<size_t>(e - b);
+}
+
 inline void parse_disposition_params(const std::string &s, Params &params) {
   std::set<std::string> cache;
-  split(s.data(), s.data() + s.size(), ';', [&](const char *b, const char *e) {
-    std::string kv(b, e);
-    if (cache.find(kv) != cache.end()) { return; }
-    cache.insert(kv);
+  auto cur = s.data();
+  const auto end = s.data() + s.size();
+  while (cur < end) {
+    const auto len = find_unquoted(cur, end, ';');
+    const auto r = trim(cur, cur + len, 0, len);
+    if (r.first < r.second) {
+      const auto tb = cur + r.first;
+      const auto te = cur + r.second;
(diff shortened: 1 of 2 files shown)


yhirose:
@22elix3r thanks for the pull request! I fixe this issue in a different way at #2573.

PR #2570, opened 2026-08-27
Parse empty Accept list elements per RFC 9110

### LINKED ISSUE #2567: parse_accept_header() rejects empty list elements, answering 400 for a legal Accept
## Behaviour

`parse_accept_header()` rejects any `Accept` value with a leading, trailing or doubled comma (httplib.h:8839):

```cpp
// Check for invalid patterns: leading/trailing commas or consecutive commas
if (s.front() == ',' || s.back() == ',' ||
    s.find(",,") != std::string::npos) {
  return false;
}
```

`Server::process_request()` validates `Accept` before routing, so a `false` here means the request is answered **400 and the connection closed**, whatever the target (httplib.h:13950):

```cpp
if (!detail::parse_accept_header(accept_header, req.accept_content_types)) {
  connection_closed = true;
```

So `Accept: text/html,` gets a 400 on every route.

## Why I think that is wrong

`Accept` is a `#rule` list — RFC 9110 §12.5.1: `Accept = #( media-range [ weight ] )` — and RFC 9110 §5.6.1.2 (Recipient Requirements) says:

> Empty elements do not contribute to the count of elements present. A recipient MUST parse and ignore a reasonable number of empty list elements: enough to handle common mistakes by senders that merge values, but not so much that they could be used as a denial-of-service mechanism.

`,application/json`, `text/html,` and `text/html,,*/*` are therefore all legal, and `,,,` is a legal empty list.

## The codebase already says so

`get_combined_header_value()` (httplib.h:7859) carries this comment:

```cpp
// RFC 9110 Section 5.6.1.2: a recipient has to parse and ignore empty list
// elements, so an empty field line must not contribute a bare comma to the
// combined value. parse_accept_header() rejects a leading comma outright,
// which would turn a legal request into 400 Bad Request.
if (it->second.empty()) { continue; }
```

That handles the case where the empty element comes from a *separate empty `Accept` field line*. It does not help when a single field value contains the empty element, which is the more common shape — a sender that merged values, or simply emitted a trailing comma.

## Note this looks deliberate

Two tests assert the current behaviour (test/test.cc:1072 and 1076):

```cpp
EXPECT_FALSE(detail::parse_accept_header(",application/json", result));
EXPECT_FALSE(detail::parse_accept_header(",,,", result));
```

which is why I am raising this as an issue rather than sending a patch — changing it means changing those, and that is your call.

## Possible directions

`split()` already skips empty elements, so simply deleting the check makes `text/html,` parse as `text/html` and `,,,` yield an empty list (equivalent to no `Accept`).

If the concern behind the check is the DoS angle the RFC itself raises, a middle option is to keep a bound — ignore empty elements up to some limit and reject beyond it — rather than reject any.

Happy to put a patch together with tests if you want it, in whichever direction you prefer.

22elix3r:
## Summary

`parse_accept_header()` rejected `Accept` values with a leading, trailing, or doubled comma, so a legal header like `Accept: text/html,` was answered 400 and the connection was closed.

RFC 9110 §5.6.1.2 requires recipients to parse and ignore empty list elements. Empty entries are now skipped; only genuinely invalid media types or quality values still fail the parse.

## Related Issue

Fixes #2567

## Changes Made

- Stop treating leading/trailing/consecutive commas as a parse failure in `parse_accept_header()`
- Ignore empty comma-separated list elements instead of setting `has_invalid_entry`
- Keep 400 behavior for actually invalid `q` values and media types
- Update unit tests and a client/server round-trip for `Accept: text/html,`
- Drop the empty-entry example from the invalid list in `example/accept_header.cc`

## Testing

Commands executed:

- `g++ -std=c++17 accept_driver.cc` (standalone driver calling `detail::parse_accept_header` from `httplib.h`)
- `./accept_driver`
- `cmake -S . -B build -DHTTPLIB_TEST=ON -DCMAKE_BUILD_TYPE=Debug`
- `cmake --build build --target httplib-test`
- `./build/test/httplib-test --gtest_filter='ParseAcceptHeaderTest.*'`

Results:

- standalone driver: all assertions passed
- 8/8 ParseAcceptHeaderTest tests passed, including `InvalidCases` and `ContentTypesPopulatedAndInvalidHeaderHandling`

## Notes

An all-comma value (`,,,`) is a legal empty list and now succeeds with an empty content-type vector, matching an omitted/empty `Accept` (no preference). Truly invalid entries such as `text/html;q=abc` still return 400.

### DIFF
--- a/example/accept_header.cc
+++ b/example/accept_header.cc
@@ -54,7 +54,6 @@ int main() {
         "text/html;q=-0.1,application/json", // q < 0.0
         "text/html;q=invalid,application/json", // invalid q value
         "invalidtype,application/json",      // invalid media type
-        ",application/json"                  // empty entry
     };
     
     for (const auto& invalid_accept : invalid_examples) {
--- a/httplib.h
+++ b/httplib.h
@@ -7858,8 +7858,7 @@ inline std::string get_combined_header_value(const Headers &headers,
   for (auto it = rng.first; it != rng.second; ++it) {
     // RFC 9110 Section 5.6.1.2: a recipient has to parse and ignore empty list
     // elements, so an empty field line must not contribute a bare comma to the
-    // combined value. parse_accept_header() rejects a leading comma outright,
-    // which would turn a legal request into 400 Bad Request.
+    // combined value.
     if (it->second.empty()) { continue; }
     if (!combined.empty()) { combined += ", "; }
     combined += it->second;
@@ -8836,12 +8835,6 @@ inline bool parse_accept_header(const std::string &s,
   // Empty string is considered valid (no preference)
   if (s.empty()) { return true; }
 
-  // Check for invalid patterns: leading/trailing commas or consecutive commas
-  if (s.front() == ',' || s.back() == ',' ||
-      s.find(",,") != std::string::npos) {
-    return false;
-  }
-
   struct AcceptEntry {
     std::string media_type;
     double quality;
@@ -8857,10 +8850,9 @@ inline bool parse_accept_header(const std::string &s,
     std::string entry(b, e);
     entry = trim_copy(entry);
 
-    if (entry.empty()) {
-      has_invalid_entry = true;
-      return;
-    }
+    // RFC 9110 §5.6.1.2: ignore empty list elements (leading, trailing,
+    // or consecutive commas). "text/html," and ",application/json" are legal.
+    if (entry.empty()) { return; }
 
     AcceptEntry accept_entry;
     accept_entry.order = order++;
--- a/test/test.cc
+++ b/test/test.cc
@@ -1067,13 +1067,27 @@ TEST(ParseAcceptHeaderTest, InvalidCases) {
   EXPECT_FALSE(
       detail::parse_accept_header("invalidtype,application/json", result));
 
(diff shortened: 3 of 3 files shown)


yhirose:
Thanks for the patch, and sorry for the timing. The fix landed on master in
e96a52e while this was open, so I am closing this as already fixed.

Your reading of RFC 9110 5.6.1.2 matches what went in, and so does the change
to parse_accept_header() itself. Two things differ on master:

* The `if (entry.empty())` guard inside the split() callback is gone as well.
  split() trims each element and skips the empty ones, so that branch was
  already unreachable.
* get_combined_header_value() got its own test. Its skip of empty field lines
  is no longer observable through a request now that a stray comma parses
  cleanly, and the existing test only exercised a leading empty field line,
  which the "combined is still empty" check already covers.

The example/accept_header.cc change was the one thing master had missed:
",application/json" was still listed as an invalid example, so the example
printed "Unexpectedly succeeded!" for it. That is fixed in f2b338a. Thanks
for catching it.

22elix3r:
Thanks for the explanation and for picking up the example. Glad the RFC 9110 empty-element reading matched what went in; no further action needed from my side.

PR #2571, opened 2026-08-27
Compress known-length content providers when the client accepts gzip

### LINKED ISSUE #2545: Static file responses are never compressed, even when the client accepts gzip
## Summary

A response whose body comes from a known-length content source is never compressed, even when the client sends `Accept-Encoding: gzip` and the content type is one we already classify as compressible. That covers `set_content_provider(length, ...)` and, through it, `set_mount_point()` and `Response::set_file_content()`.

The same bytes served through `set_content()` are compressed. So static file serving, which is where compression pays off most, is the one path that does not get it.

## Measurements

0.53.1, built with `CPPHTTPLIB_ZLIB_SUPPORT`, client sending `Accept-Encoding: gzip`.

| request | content source | Content-Encoding | bytes on wire |
|---|---|---|---|
| `/static/index.html` | `set_mount_point()` | none | 31,868 |
| `/static/style.css` | `set_mount_point()` | none | 25,690 |
| `/static/app.js` | `set_mount_point()` | none | 24,080 |
| `/static/data.json` | `set_mount_point()` | none | 16,981 |
| `/fc` (the same index.html) | `set_file_content()` | none | 31,868 |
| `/inline` (the same index.html) | `set_content()` | gzip | **2,581** |

The last two rows serve byte-identical content. Compression ratios for the rest (`gzip -9`): style.css 27.6x, app.js 14.5x, data.json 8.5x.

## Which sources compress today

| content source | compressed |
|---|---|
| `set_content()` (owned body) | yes |
| `set_chunked_content_provider()` | yes |
| `set_content_provider(length, ...)` | **no** |

The inconsistency is not obvious from the API, and the workaround is to give up mmap and read the file into a `std::string` so it lands in `res.body`.

## Cause

In `Server::apply_ranges()`, the compression block sits inside the branch taken when `res.body` is non-empty. A known-length provider has an empty `res.body` and a non-zero `content_length_`, so it takes the other branch, which only sets `Content-Length` and returns. `detail::encoding_type()` is computed before the split but is never consulted on that side.

## Notes for a fix

A few things to decide before changing anything:

1. **Range requests.** RFC 9110 applies Range to the representation after content coding, so a compressed 206 means compressing the whole file and then slicing it. Serving ranges uncompressed is the usual way out.
2. **Cost per request.** Compressing the same file on every hit is wasteful. Pre-compressed sidecar files (nginx `gzip_static`) or an in-memory cache are the usual answers, but both are a bigger feature than flipping the branch.
3. **Zero-copy.** Compressing an mmap'd file gives up the zero-copy path. `can_compress_content_type()` already returns false for images, video, and archives, so only text-ish files would pay it.
4. **Default or opt-in.** Turning this on changes long-standing behavior. Nine existing tests assert the uncompressed `Content-Length` for static responses: `GetFileContent`, `GetFileContentWithRange`, `GetFileContentWithContentType`, `HeadMethod200Static`, `StaticFileRange`, `StaticFileRangeHead`, `StaticFileRangeBigFile`, `StaticFileRangeBigFile2`, `StaticFileBigFile`.

I confirmed (4) by making the known-length branch compress: those nine fail, and nothing else does.

22elix3r:
## Summary

Static file responses and other known-length content providers were never compressed, even when the client sent `Accept-Encoding: gzip` and the content type is already treated as compressible. The same bytes served through `set_content()` were gzipped.

That covers `set_content_provider(length, ...)`, and through it `set_mount_point()` and `Response::set_file_content()`.

## Related Issue

Fixes #2545

## Changes Made

`Server::apply_ranges()` only ran the compressor when `res.body` was non-empty. A known-length provider has an empty body and a non-zero `content_length_`, so it took the other branch, which set `Content-Length` and returned. `detail::encoding_type()` was computed but never consulted on that path.

When the client accepts a supported encoding and the response is not a 206 range:

- set `Transfer-Encoding: chunked`, `Content-Encoding`, and `Vary: Accept-Encoding`
- write the provider through a compressor, using the same length-framing as `write_content()` (including a provider that calls `done()` short of the advertised length)

Compression is chunked rather than buffering the whole representation into `res.body`. Buffering would stall streaming known-length providers (timeouts, mid-stream abort) until they finished. 206 responses stay uncompressed so `Content-Range` still names identity bytes.

## Testing

Built with `cmake -DHTTPLIB_TEST=ON -DHTTPLIB_REQUIRE_ZLIB=ON` and ran `httplib-test`.

New tests (under `CPPHTTPLIB_ZLIB_SUPPORT`):

- `GzipStaticFile` / `GzipStaticFileDecompressed` — mount point (`/dir/1MB.txt`)
- `GzipFileContent` — `set_file_content()`
- `GzipKnownLengthContentProvider` — `set_content_provider(length)`
- `GzipStaticFileRangeUncompressed` — `Range` stays identity even with `Accept-Encoding: gzip`
- `GzipStaticFileHead` — HEAD gets the same encoding headers
- `NoGzipStaticOctetStream` — `application/octet-stream` is not compressed

Existing tests that pin uncompressed `Content-Length` for full known-length responses now send an empty `Accept-Encoding` so they keep testing the identity representation.

## Notes

Chunked encoding is used instead of a compressed `Content-Length` so mmap/static files and incremental `set_content_provider(length)` handlers share one path with `set_chunked_content_provider()`. Range requests are left uncompressed on purpose (RFC 9110 applies Range to the selected representation after content coding; serving identity ranges is the usual way out).

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -8590,6 +8590,68 @@ inline bool write_content_chunked(Stream &strm,
                                compressor, error);
 }
 
+// Compress a known-length provider as chunked. Length-framing (including a
+// provider that calls done() short of `length`) matches write_content(); the
+// payload is gzip/br/zstd-chunked like write_content_chunked(). Buffering the
+// whole representation would stall streaming providers until they finished.
+template <typename T, typename U>
+inline bool write_content_chunked_known_length(
+    Stream &strm, const ContentProvider &content_provider, size_t offset,
+    size_t length, const T &is_shutting_down, U &compressor) {
+  auto end_offset = offset + length;
+  auto ok = true;
+  auto finished = false;
+  DataSink data_sink;
+
+  auto emit = [&](const std::string &payload) -> bool {
+    if (payload.empty()) { return true; }
+    auto chunk = from_i_to_hex(payload.size()) + "\r\n" + payload + "\r\n";
+    return write_data(strm, chunk.data(), chunk.size());
+  };
+
+  data_sink.write = [&](const char *d, size_t l) -> bool {
+    if (!ok || l == 0) { return ok; }
+    offset += l;
+    std::string payload;
+    if (!compressor.compress(d, l, false,
+                             [&](const char *data, size_t data_len) {
+                               payload.append(data, data_len);
+                               return true;
+                             })) {
+      ok = false;
+      return false;
+    }
+    if (!emit(payload)) { ok = false; }
+    return ok;
+  };
+
+  data_sink.is_writable = [&]() -> bool { return strm.is_peer_alive(); };
+  data_sink.done = [&]() { finished = true; };
+
+  while (offset < end_offset && !finished && !is_shutting_down()) {
+    if (!strm.wait_writable() || !strm.is_peer_alive()) { return false; }
+    if (!content_provider(offset, end_offset - offset, data_sink)) {
+      return false;
+    }
+    if (!ok) { return false; }
+  }
+
+  if (offset < end_offset) { return false; }
+
+  std::string payload;
+  if (!compressor.compress(nullptr, 0, true,
+                           [&](const char *data, size_t data_len) {
+                             payload.append(data, data_len);
+                             return true;
+                           })) {
+    return false;
+  }
+  if (!emit(payload)) { return false; }
+
+  constexpr const char done_marker[] = "0\r\n\r\n";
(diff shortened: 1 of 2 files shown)


22elix3r:
Ready for review whenever you have a moment. I cannot request a reviewer from this account.

yhirose:
@22elix3r thanks again, but it's also fixed in a different way at #2572.

PR #2574, opened 2026-08-27
split header field parameters outside quoted-strings

metsw24-max:
The multipart parser reads each part's name and filename out of the Content-Disposition header by splitting the parameter list on ';' and each parameter on '=', and extract_media_type does the same for Content-Type parameters. Neither split is aware of quoted-strings, but RFC 9110 5.6.6 lets a parameter value be a quoted-string that itself contains those delimiters, so a part sent as filename="a;b=c.txt" is truncated to "a and picks up a stray b=c.txt parameter, while filename="x=y.txt" collapses to y.txt". The disposition value comes straight off an untrusted multipart body, so the filename an application ends up with can differ from the one the client sent and from the one a proxy or upload filter in front of the server inspected. I noticed it while checking how quoted filenames survive the parser and reproduced it against a part whose filename contains a semicolon. The fix moves both call sites onto a small shared helper that tracks the quoted-string state and only treats ';' and '=' as delimiters outside quotes, with a backslash escaping the next character; values that carry no quoted delimiter parse exactly as before. I kept it to that one helper so the disposition and media-type paths stay in step and there is a single place to reason about the quoting rules. Adds a unit test covering the quoted-delimiter cases.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -7290,6 +7290,64 @@ find_content_type(const std::string &path,
   }
 }
 
+// RFC 9110 §5.6.6: a header field parameter value may be a token or a
+// quoted-string, and a quoted-string may itself contain the ';' that
+// separates parameters and the '=' that separates a name from its value.
+// Splitting on those characters unconditionally therefore truncates a value
+// such as filename="a;b" or filename="a=b". Track the quoted-string state so
+// the delimiters are only honoured outside quotes; a backslash escapes the
+// next character inside quotes. `fn` receives the whitespace-trimmed name and
+// value with any surrounding quotes left in place for the caller to strip.
+inline void parse_header_field_parameters(
+    const std::string &s,
+    std::function<void(const std::string &, const std::string &)> fn) {
+  size_t i = 0;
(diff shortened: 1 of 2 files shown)


yhirose:
@metsw24-max, thanks for the detailed writeup and the test coverage. However, this is already fixed on master by #2573 (b4ec1bb), which added `split_unquoted()` / `divide_param_pair()` and routed both `parse_disposition_params()` and `extract_media_type()` through them. It covers the same quoted `;` and `=` cases this PR targets.

The one thing this PR adds beyond #2573 is backslash-escape handling inside quoted-strings. That was left out there on purpose. Neither browsers nor httplib's own sender produce a backslash-escaped quote, they percent-encode `"` instead, so there was nothing real to decode.

metsw24-max:
Ah, I'd missed that #2573 had already landed, it does cover the same cases. Fair point on the backslash escapes as well; if senders percent-encode the quote there's nothing real for the parser to unescape, so leaving that out keeps things simpler.

PR #2575, opened 2026-08-31
Don't compress a response whose handler already set Content-Encoding

jhen0409:
A handler serving pre-compressed content (e.g. build-time gzipped static assets) sets `Content-Encoding` itself. The server only consults `Accept-Encoding` and the content type, so it compresses the body a second time and appends a second `Content-Encoding: gzip` field line, clients that decode one coding per listed value (like Safari) render raw gzip bytes. The test suite currently works around it ("`image/jpeg` keeps the server from applying a content coding of its own"). Hit by llama.cpp's server while enabling zlib support (ggml-org/llama.cpp#28059).

Fix: `detail::encoding_type(req, res)` returns `None` when the response already carries `Content-Encoding`. Since the chunked path used to negotiate twice (headers in `apply_ranges()`, compressor at write time: where the
guard would now misfire on the header `apply_ranges()` itself set), the coding is decided once: `file_content_encoding_` (#2572) is generalized to `content_encoding_` and shared by the file-backed and chunked provider paths, which are mutually exclusive.

Adds `ContentEncodingTest.PreEncodedResponseIsNotCompressedAgain` (fails before this change: two field lines, double-encoded body). Full suite passes with ZLIB, Brotli, (887 tests).

NOTE: The fix is helped by Claude Code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1474,6 +1474,19 @@ The server can apply compression to the following MIME type contents:
 - application/protobuf
 - application/xhtml+xml
 
+A response that already carries `Content-Encoding` is sent as it is. A handler serving content it encoded itself, an asset compressed at build time for instance, keeps its own coding and its own bytes:
+
+```c++
+svr.Get("/app.js", [](const Request & /*req*/, Response &res) {
+  res.set_header("Content-Encoding", "gzip");
+  res.set_content(gzipped_asset, "application/javascript");
+});
+```
+
+This holds for every kind of response, including the file-backed ones below.
+
+`Vary: Accept-Encoding` is added only to responses the server encoded itself. A handler that chooses between an encoded and an identity representation by reading `Accept-Encoding` should set the field itself, so that shared caches keep the two apart.
+
 ### Static file compression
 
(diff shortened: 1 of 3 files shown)


yuvarajayuvarajay276-wq:
[](url)

yhirose:
Thanks again for the fix, I've merged it.

While reviewing I noticed static file responses had the same issue this PR fixes elsewhere: `Server::static_file_encoding()` goes through the content-type overload of `detail::encoding_type()`, so it never saw the new guard. A pre-compressed file served through `set_static_file_compression(true)` — a mount point of build-time compressed assets, or `set_file_content()` after the handler set `Content-Encoding` — would still get compressed a second time.

I pushed a small follow-up commit onto this branch to close that gap. It shares your guard rather than duplicating it: `encoding_type()` now takes the content type as an argument, since the only reason the file-backed path couldn't call it directly was that the response doesn't have a content type yet at that point. I also renamed `content_encoding_` to `content_coding_` (it holds the coding chosen for the body, not the header's value), added tests for the chunked-provider case in both directions, and added a note to the README about the behavior, including `Vary`, which is only added when the server picked the coding itself.

Hope that's a welcome addition rather than a surprise — happy to adjust anything if you'd rather it looked different.

PR #2577, opened 2026-09-02
let EXTRA_CXXFLAGS override -fsanitize=address in test/Makefile

avionicharshit-byte:
the test suite cannot start on arm64 linux kernels built with a 39 bit virtual address space, raspberry pi os included. asan maps its allocator at a fixed `0x500000000000`, the top user address there is `0x8000000000`, so it aborts before the first test:

```
AddressSanitizer: CHECK failed: sanitizer_allocator_primary64.h:131 "((kSpaceBeg)) == ((address_range.Init(TotalSpaceSize, PrimaryAllocatorName, kSpaceBeg)))" (0x500000000000, 0xfffffffffffffff4)
```

there is no way to opt out today. `EXTRA_CXXFLAGS` sits before `-fsanitize=address` in `CXXFLAGS`, so `-fno-sanitize=address` loses. this moves `$(EXTRA_CXXFLAGS)` to the end. existing ci uses are unaffected, `-m32` and `-fno-exceptions` are position independent and `-std=c++17` already relies on last wins.

verified on a raspberry pi 4, aarch64, gcc 14.2.0. `make CXX=g++ EXTRA_CXXFLAGS=-fno-sanitize=address test` builds with zero warnings under the existing warning set, and the suite runs:

```
846 tests from 162 test suites ran
[  PASSED  ] 842 tests
[  SKIPPED ] 4 tests
```

the skips are environmental, 3 want `CPPHTTPLIB_TEST_ISSUE_2431` and a sinkhole dns, 1 wants a comma decimal locale that is not installed.

this fixes nothing in `httplib.h`, arm64 and gcc both look clean.

### DIFF
--- a/test/Makefile
+++ b/test/Makefile
@@ -1,5 +1,5 @@
 CXX = clang++
-CXXFLAGS = -g -std=c++11 -I. -Wall -Wextra -Wtype-limits -Wconversion -Wshadow $(EXTRA_CXXFLAGS) -DCPPHTTPLIB_USE_NON_BLOCKING_GETADDRINFO -fsanitize=address # -fno-exceptions -DCPPHTTPLIB_NO_EXCEPTIONS
+CXXFLAGS = -g -std=c++11 -I. -Wall -Wextra -Wtype-limits -Wconversion -Wshadow -DCPPHTTPLIB_USE_NON_BLOCKING_GETADDRINFO -fsanitize=address $(EXTRA_CXXFLAGS) # -fno-exceptions -DCPPHTTPLIB_NO_EXCEPTIONS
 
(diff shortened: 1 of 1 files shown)


avionicharshit-byte:
the abi failure here isnt from this pr, its a one line reorder in test/Makefile with no library changes

abidiff builds against the last tag, which is v0.54.1, and 7d53a31 (#2575) added a Response param to
detail::encoding_type and Server::static_file_encoding after that tag was cut

master's own push run for that commit is red for the same two symbols
https://github.com/yhirose/cpp-httplib/actions/runs/33455962542

yhirose:
@avionicharshit-byte thanks!

PR #2578, opened 2026-09-07
Added a feature test to auto enable/disable CPPHTTPLIB_USE_NON_BLOCKI…

TobiasWallner:
## Summary

Add CMake feature detection for `CPPHTTPLIB_USE_NON_BLOCKING_GETADDRINFO` on Windows.

When `HTTPLIB_USE_NON_BLOCKING_GETADDRINFO` is enabled, CMake now checks whether `GetAddrInfoExCancel` is available before exporting `CPPHTTPLIB_USE_NON_BLOCKING_GETADDRINFO` through the `httplib::httplib` target. If the symbol is not available, the non-blocking getaddrinfo support is disabled automatically.

## Motivation

Some Windows machines/toolchains do not provide `GetAddrInfoExCancel`, which causes downstream projects using cpp-httplib via CMake to fail compilation when `CPPHTTPLIB_USE_NON_BLOCKING_GETADDRINFO` is exported unconditionally.

Previously, consumers had to work around this by manually removing the interface compile definition from `httplib::httplib`. With this change, cpp-httplib configures itself correctly out of the box.

## Testing

Verified on a Windows machine where `GetAddrInfoExCancel` is unavailable:

- CMake reports `GetAddrInfoExCancel - not found`
- `HTTPLIB_IS_USING_NON_BLOCKING_GETADDRINFO` is set to `FALSE`
- `CPPHTTPLIB_USE_NON_BLOCKING_GETADDRINFO` is not present on the consumer compile line
- compiled-library mode builds successfully

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,7 +15,7 @@
 	* HTTPLIB_REQUIRE_BROTLI (default off)
 	* HTTPLIB_REQUIRE_ZSTD (default off)
 	* HTTPLIB_DISABLE_MACOSX_AUTOMATIC_ROOT_CERTIFICATES (default off)
-	* HTTPLIB_USE_NON_BLOCKING_GETADDRINFO (default on)
+	* HTTPLIB_USE_NON_BLOCKING_GETADDRINFO (default on when supported)
 	* HTTPLIB_COMPILE (default off)
 	* HTTPLIB_INSTALL (default on)
 	* HTTPLIB_SHARED (default off) builds as a shared library (if HTTPLIB_COMPILE is ON)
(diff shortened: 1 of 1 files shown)


yhirose:
@sum01 @jimmy-park @Tachi107 this PR looks good to me. Could you please take a look at it?

yhirose:
@TobiasWallner could you take a look at the CI failures of `test / windows with SSL` and `test /windows without SSL`? (You don't need to worry about other failures.)

CMakeLists.txt
```diff
@@ -181,6 +181,20 @@ if(HTTPLIB_DISABLE_MACOSX_AUTOMATIC_ROOT_CERTIFICATES)
...
+
+	set(_httplib_cmake_required_definitions ${CMAKE_REQUIRED_DEFINITIONS})
+	list(APPEND CMAKE_REQUIRED_DEFINITIONS -D_WIN32_WINNT=0x0A00)
+	check_cxx_symbol_exists(GetAddrInfoExCancel "winsock2.h;ws2tcpip.h" HTTPLIB_HAVE_GETADDRINFOEXCANCEL)
+	set(CMAKE_REQUIRED_DEFINITIONS ${_httplib_cmake_required_definitions})
+	unset(_httplib_cmake_required_definitions)
+
+	if(NOT HTTPLIB_HAVE_GETADDRINFOEXCANCEL)
+		set(HTTPLIB_IS_USING_NON_BLOCKING_GETADDRINFO FALSE)
+		message(STATUS "GetAddrInfoExCancel is unavailable; disabling non-blocking getaddrinfo.")
```
sum01, line 195: Wouldn't it probably be better to use either `WARNING` or even `FATAL_ERROR` here?

I'm not sure what people would prefer personally, but having the build stop makes some sense to me, as it's an explicitly enabled option that will not work/be supported.

TobiasWallner: Good point, i would use a warning, because the user might still want to compile it and provide the function by itself or link something that cmake does not see at this point.

yhirose:
@TobiasWallner,  since the master branch has progressed quite a bit, please run `rebase master`.

TobiasWallner:
yes, sry i will, i just had not had the time and head for it lately

yhirose:
@TobiasWallner could you please remove things unrelated to CMake such as https://github.com/yhirose/cpp-httplib/pull/2578/changes/b29d191d28ca054eb34745dcce676c9f88c4fd0b?

TobiasWallner:
> @TobiasWallner could you please remove things unrelated to CMake such as [b29d191](https://github.com/yhirose/cpp-httplib/commit/b29d191d28ca054eb34745dcce676c9f88c4fd0b)?

Sry, sure done

CMakeLists.txt
TobiasWallner: > Wouldn't it probably be better to use either `WARNING` or even `FATAL_ERROR` here?
> 
> I'm not sure what people would prefer personally, but having the build stop makes some sense to me, as it's an explicitly enabled option that will not work/be supported.

Changed it to a warning

TobiasWallner:
> @TobiasWallner could you take a look at the CI failures of `test / windows with SSL` and `test /windows without SSL`? (You don't need to worry about other failures.)

https://github.com/yhirose/cpp-httplib/pull/2578/commits/522b60ac3cf87fb8fe28703cf800a6ec8014e977

Lets see if this correctly addresses that, I cannot reproduce the failure on my local machine, lets see what happens in CI.

PR #2579, opened 2026-09-08
send each credential only to its own hop in write_request

metsw24-max:
**Proxy credentials sent inside the CONNECT tunnel**
ClientImpl::write_request attaches Proxy-Authorization whenever a proxy is enabled for the host, so an SSLClient sends the proxy's Basic or Bearer credentials on every request inside the TLS tunnel, where only the origin reads them, and likewise puts the origin's Authorization on the CONNECT request that only the proxy reads. Both now go to the hop that actually reads the message, reusing the is_ssl() distinction the 407 handling already makes, and the tunnel fixture gains a test per scheme that checks what each hop received.


yhirose:
@metsw24-max thank you for the excellent contribution!

PR #2581, opened 2026-09-09
reject ambiguously framed responses in client read paths

metsw24-max:
the client reads a response body without rejecting ambiguous framing the way Server::process_request already does. clientimpl::process_request and clientimpl::open_stream hand the response straight to read_content/BodyReader, which prefer the chunked coding and drop content-length when both are present, and treat a transfer-encoding whose final coding is not chunked as bodyless. a hostile origin or an intermediary can then send a response carrying both a non-zero content-length and transfer-encoding: chunked, so this client ends the body by the chunked framing while a front-end that trusts content-length ends it elsewhere, desynchronising a pooled keep-alive connection so a later response is paired with the wrong request. this mirrors the server's rfc 9112 §6.3 guard onto both client read paths and refuses the same shapes, leaving head/204/304 alone since they legitimately carry framing headers with no body. note this is distinct from #2487, which only corrects whether a multi-coding transfer-encoding is recognised as chunked; a content-length paired with chunked passes that detection and still needs rejecting here. covered by ClientResponseSmugglingTest for both the buffered and open_stream paths.

### DIFF
--- a/httplib.h
+++ b/httplib.h
@@ -8365,6 +8365,17 @@ inline bool is_chunked_transfer_encoding(const Headers &headers) {
   return case_ignore::equal(last_coding, "chunked");
 }
 
+inline bool has_conflicting_content_length(const Headers &headers) {
+  // RFC 9112 §6.3: a message carrying both Transfer-Encoding and a non-zero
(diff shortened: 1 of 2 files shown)


yhirose:
@metsw24-max thank you!

Issue #2011, opened 2024-12-24
Port/Address re-use


Frank-Friemel:
https://github.com/yhirose/cpp-httplib/blob/d647f484a48ae59b429e7e3a86e2b8b3fc493167/httplib.h#L2044

Hi, I think there is a bug in the above line. This does not work. The return value is -1. I guess because you've previously enabled address re-use. Just applying SO_EXCLUSIVEADDRUSE (without SO_REUSEADDR) works.

I mean ... what is the purpose at all? In the Linux path you enable address/port re-usage. And in the Win32 path you try to disable it by acquiring the port exclusively? Currently it's enabled for both platforms, because SO_EXCLUSIVEADDRUSE  fails. I stumbled over this because for my project this behaviour was kind of unexpected. My expectation was that a port can only be bound by one application/process/thread.

yhirose:
@Frank-Friemel thank you for the report. I'll take a look at it when I am time.

yhirose:
@Frank-Friemel sorry for the delay. As you mentioned, there is inconsistency between the Windows implementation and others (Linux and macOS). I actually prefer 'address/port re-usage' over 'exclusive address usage'. As for Windows, it seems I misunderstood the WinSock spec. (https://github.com/yhirose/cpp-httplib/issues/506#issuecomment-637863292)

So in order to make all the platform behaviors consistent, I am actually removing `SO_EXCLUSIVEADDRUSE` from the Windows implementation.

If you would like to have 'exclusive address usage', you can set the socket option to SO_EXCLUSIVEADDRUSE with `Server::set_socket_options`.

Issue #506, opened 2020-05-31
Use of SO_REUSEADDR


vshu3000:
```
1908     // Make 'reuse address' option available
1909     int yes = 1;
1910     setsockopt(sock, SOL_SOCKET, SO_REUSEADDR, reinterpret_cast<char *>(&yes),
1911                sizeof(yes));
1912 
1913 #ifdef SO_REUSEPORT
1914     setsockopt(sock, SOL_SOCKET, SO_REUSEPORT, reinterpret_cast<char *>(&yes),
1915                sizeof(yes));
1916 #endif
```

So now it always reuses socket.  The line outside ifdef should be removed.

vshu3000:
The code actually always tries to reuse port.  That means in my case two servers can bind to the same port without causing EADDRINUSE.  This is unexpected behavior.

yhirose:
@vshu3000, are you saying that line 1910-1911 should be removed?

yhirose:
We need to handle this matter carefully according to the following information:
https://stackoverflow.com/questions/14388706/how-do-so-reuseaddr-and-so-reuseport-differ

PR #2584, opened 2026-09-14
Use SO_EXCLUSIVEADDRUSE instead of SO_REUSEADDR on Windows

dqsjqian:
## Problem

On Windows, `SO_REUSEADDR` has a very different meaning than on POSIX systems: it allows a socket to **steal** a bind — a second process can successfully bind to a port that another process is already bound to, and the two sockets then compete for incoming traffic. As a result, starting two `Server` instances on the same port silently succeeds on Windows, whereas on Linux/macOS the second bind fails with `EADDRINUSE` as expected.

This was reported in #506, #1144 and #2011. As noted in #506, on Windows "using SO_REUSEADDR and SO_EXCLUSIVEADDRUSE would be a correct answer" — but only the `SO_REUSEADDR` half of that statement is currently implemented, and it alone produces the stealing behavior.

## Fix

In `default_socket_options`, use `SO_EXCLUSIVEADDRUSE` on Windows instead of `SO_REUSEADDR`. `SO_EXCLUSIVEADDRUSE` is the documented Windows option for exclusive binding: binding to an occupied port fails with `WSAEADDRINUSE`, matching POSIX expectations while still allowing a server to rebind its own port after restart (the primary use case of `SO_REUSEADDR` that https://github.com/yhirose/cpp-httplib/issues/506 wants to keep — `SO_EXCLUSIVEADDRUSE` permits rebinding a port in TIME_WAIT left over by a previous instance of the *same* server, it only blocks binding over *another live socket*).

Non-Windows platforms are unchanged. Client sockets are unaffected: `SO_EXCLUSIVEADDRUSE` only affects bind semantics and cpp-httplib clients connect without binding.

## Notes

- One behavioral consideration: on Windows, a socket with `SO_EXCLUSIVEADDRUSE` must not later enable `SO_REUSEADDR` (`setsockopt` fails with `WSAEINVAL`). The only in-tree place that touches `SO_REUSEADDR` on Windows after socket creation is the AF_UNIX workaround in `create_socket`, which sets it to **0** (disables it), which is compatible.
- Verified against a downstream project's CI: with this change, a regression test that starts a server on a port already occupied by another process now sees the server exit with a bind error on Windows (previously it silently kept running), matching the Linux/macOS behavior.


yhirose:
@dqsjqian Thank you for the PR and the detailed write-up.

The current default is intentional. cpp-httplib enables address/port reuse on all platforms: `SO_REUSEPORT` on Linux and macOS, and `SO_REUSEADDR` on Windows. So two servers can bind to the same port on Linux and macOS as well, not only on Windows. We actually removed `SO_EXCLUSIVEADDRUSE` from the Windows path in #2011 to keep the behavior consistent across platforms. Switching Windows alone to exclusive binding would bring that inconsistency back, and changing the default on every platform would affect too many existing users.

If you need exclusive binding, you can replace the default options with `set_socket_options`:

```cpp
svr.set_socket_options([](socket_t sock) {
#ifdef _WIN32
  httplib::set_socket_opt(sock, SOL_SOCKET, SO_EXCLUSIVEADDRUSE, 1);
#else
  httplib::set_socket_opt(sock, SOL_SOCKET, SO_REUSEADDR, 1);
#endif
});

Thanks for your consideration.

PR #2585, opened 2026-09-20
reject control characters in chunk extensions in read_payload

metsw24-max:
**Chunk-size line accepts control characters after the size**
`ChunkedDecoder::read_payload` only looks at the first byte after the chunk-size, and `getline()` reads up to CRLF, so a bare LF (or CR, or any other control byte) after the size or inside the chunk-ext is taken as extension text, whereas an intermediary that ends the line on that LF would delimit the chunks differently (request smuggling, RFC 9112 §7.1.1). The rest of the line is now checked where it is parsed, which should cover the server and both client read paths since they share this decoder; the tests cover the rejected shapes and that ordinary extensions still pass.


yhirose:
Thanks!

